PR #8, opened 2026-05-02
Add typed-block HCL config + plugin system

piscisaureus:
* New policy grammar with two-label blocks dispatched to plugins by
  their first label (`endpoint "https" "github-dev"`,
  `credential "bearer_token" "github-pat"`, ...). Single flat
  namespace; bare-name references resolved at load time with
  kind / family checks.
* `config/` package: two-pass loader (`hclparse` + per-plugin gohcl
  decode), plugin contract (`Plugin`, `RefSpec`, `BuildCtx`), and a
  registry seeded by each plugin package's `init()`.
* Built-in plugins covering every type used by the v14 fixture: 14
  credentials (bearer / cookie / header / mtls / postgres / anthropic
  manual + oauth / slack / telegram / gemini / openai-codex / notion
  / clickhouse / aws-eks), 5 endpoints (https / postgres / kubernetes
  / clickhouse_https / clickhouse_native), 3 rule types
  (http_rule / sql_rule / k8s_rule), 2 approvers (llm / human).
  Schema-only at this stage — runtime hooks land in follow-up
  commits on this PR.
* Golden-file test harness under `config/testdata/`. `feature_*.hcl`
  pairs with `feature_*.want.json`; `error_*.hcl` pairs with
  `error_*.errors.txt`; `full.hcl` is the full v14 fixture (cross-
  kind name colliders renamed `-cred`). Run with `-update` to
  regenerate goldens.

The legacy `hclsimple`-based loader in `main.go` still drives the
running gateway. Follow-up commits on this PR rip out the legacy
schema, wire runtime hooks behind the plugin interface, and rebuild
the dashboard's HCL write-back.


config/plugin.go
```diff
@@ -0,0 +1,140 @@
...
+	case KindPolicy, KindProfile:
+		return 1 // name
+	}
+	return 0
+}
+
+// Plugin describes one (Kind, Type) pair — e.g. (endpoint, https) or
+// (credential, bearer_token). Built-in plugins call Register at init
+// time; the loader looks them up by (Kind, Type) when decoding blocks.
+type Plugin struct {
```
littledivy, line 52: **Plugin abstraction has no plugins.**

There's no SO loading, no out-of-tree registration, no third-party extensibility. Every plugin lives in `config/plugins/`; `init()` registers into a global registry whose only consumer is `Lookup(kind, type)`. That's a typed switch dressed up as a registry.

The cost paid for it:

- `interface{}` everywhere (`Body any`, `Runtime any`, `New() any`, `Build` / `Validate` returning `any`)
- runtime type-asserts in the dispatcher with silent failures (`ResolveCredential` does `ep.Plugin.Runtime.(PlaceholderDetector)` at `runtime/dispatch.go:95`; if the assertion fails, fallback path runs with no diagnostic)
- the init-time satisfaction asserts at `plugins/credentials/credentials.go:319-324` are a manual workaround for the lost compile-time exhaustiveness — and they only cover wired runtimes, not "plugin X must satisfy interface Y"
- ~30 lines of anonymous-struct-literal boilerplate per endpoint type to satisfy `compile.go:235`'s `interface{ EndpointCredentials() []struct{Placeholder string; Credential string} }` assertion (any plugin returning a *named* type silently fails to satisfy)

Closed enum + sealed switch on `kind+type` would give compiler-checked exhaustiveness, named return types instead of anonymous struct literals, and ~1/3 the LOC. The plugin façade would buy us something if we planned out-of-tree plugins; do we?

main.go
```diff
@@ -21,650 +21,150 @@ import (
...
+}
+
+// adaptLegacy translates the new policy-aware *config.Gateway into the
+// legacy *Config shape the request handler still consumes. Operational
+// fields copy through verbatim. Profile / Rule / Approver / Integration
+// / OAuth slices are left empty here — they get populated in commit 4
+// when the plugin runtime takes over request dispatch. Profiles are
+// the exception: their NAMES must round-trip so device onboarding,
+// dashboard tabs, and `g.profileFor` keep working.
+func adaptLegacy(gw *config.Gateway) *Config {
```
littledivy: Let's remove legacy config

config/plugins/rules/rules.go
```diff
@@ -0,0 +1,577 @@
...
+	Endpoint  string   `hcl:"endpoint,optional"`
+	Endpoints []string `hcl:"endpoints,optional"`
+	Priority  int      `hcl:"priority,optional"`
+	Disabled  bool     `hcl:"disabled,optional"`
+
+	// Match is decoded raw and interpreted per family in Build. An
+	// absent match block matches everything — the v14 catch-all
+	// pattern (`rule "..." "X-default" { priority = -100; verdict =
+	// "deny" }`) relies on this.
+	Match cty.Value `hcl:"match,optional"`
```
littledivy, line 45: **`Match` and `Approve` as `cty.Value` defeat parse-time checking.**

Decoded raw because keys vary by family. But that means `methood = "POST"` parses fine and silently never fires — the rule just becomes a no-op. The `knownMatchKeys` map at line 111 catches typos, but only after Build runs, and only against a hand-maintained allowlist that has to be kept in sync with the matcher.

gohcl can do this for free if Match becomes one struct per family:

```go
type HTTPMatch struct {
  Method       []string          `hcl:"method,optional"`
  Path         string            `hcl:"path,optional"`
  Query        map[string][]string `hcl:"query,optional"`
  Headers      map[string]string `hcl:"headers,optional"`
  BodyJSON     cty.Value         `hcl:"body_json,optional"`
  BodyContains string            `hcl:"body_contains,optional"`
  Credential   string            `hcl:"credential,optional"`
}
```

gohcl rejects unknown attrs at parse with a precise pointer at the typo. No allowlist to maintain, no silent no-ops. Costs one decode struct per rule type (3 of them). The dispatch on rule type that already happens in `families` at line 107 is the natural place to pick which struct to decode against.

Same argument for `Approve` (string-or-struct heterogeneous list) — typed as `[]ApproveStage` with a Decode hook on the stage type would catch typo'd approver names at parse time.

littledivy (commented):
First pass on the design. Concerns are mostly about complexity vs. who actually consumes it today, plus one footgun around schema-only credentials that I think is a merge blocker.

config/plugin.go
littledivy: we can do it later when we have out-of-tree plugins

config/plugins/rules/rules.go
littledivy: Done

littledivy (approved):

PR #41, opened 2026-05-03
hitl: slack approver via chat.postMessage + Block Kit

littledivy:


littledivy:
Superseded by the slack_tokens credential in PR #42 — Slack approvers will use that instead of a standalone webhook.

PR #71, opened 2026-05-05
Add ClickHouse native protocol gateway

arnauorriols:
First-pass ClickHouse native protocol support for the gateway. Mirrors the postgres plugin shape.

Scope:
- Parse the client Hello and swap placeholder username/password bytes for the bound credential's real values.
- Emit one `ConnEvent` per session.
- Pipe bidirectionally after auth.

Out of scope (follow-up): SQL query parsing.

Adds:
- `ClickhouseAuthCredential` interface; `clickhouse_credential` implements it.
- `ClickhouseNativeEndpoint` gains `Port`, TLS, `ConnRouter`.
- `ClickhouseNativeEndpointRuntime` implements `ConnEndpointRuntime`.
- `VarUInt` + Hello parser/serializer with roundtrip + short-buffer tests.
- Generic conn-index dispatch in the WG forwarder so operators can run on either 9000 (plaintext) or 9440 (TLS) without a port-table edit.

### DIFF
--- a/config/plugins/endpoints/clickhouse_native.go
+++ b/config/plugins/endpoints/clickhouse_native.go
@@ -85,6 +85,9 @@ func (e *ClickhouseNativeEndpoint) port() int {
 	if e.Port > 0 {
 		return e.Port
 	}
+	if e.TLS {
+		return 9440
+	}
 	return 9000
 }
 
--- a/config/plugins/endpoints/clickhouse_native.go
+++ b/config/plugins/endpoints/clickhouse_native.go
@@ -1,17 +1,45 @@
 package endpoints
 
 // clickhouse_native endpoint: ClickHouse's binary native protocol
-// (default port 9000). Pairs with clickhouse_https for the same
-// upstream cluster.
+// (default port 9000 plaintext / 9440 TLS). Pairs with
+// clickhouse_https for the same upstream cluster.
+//
+// Iter 1 scope: parse the Hello packet, swap placeholder bytes in
+// the agent-supplied (username, password) for the credential's real
+// values, emit one connection event, then transparent bidirectional
+// pipe. SQL parsing lands in a follow-up iteration.
 
 import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"log"
+	"net"
+
 	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/zclconf/go-cty/cty"
 
 	"github.com/denoland/clawpatrol/config"
+	"github.com/denoland/clawpatrol/config/runtime"
 )
 
+// ClickhouseNativeEndpoint addresses one ClickHouse server reachable
+// via the binary native protocol. Operators bind a single
+// clickhouse_credential; the runtime parses the agent's Hello and
+// substitutes the credential's (user, password) where the agent
+// embedded a placeholder.
+//
+// TLS toggles upstream TLS-wrapping. The native protocol is
+// persistent (no inner TLS negotiation), so this is a one-time
+// decision per endpoint. Default false: WG already encrypts
+// agent→gateway and most self-hosted ClickHouse on a private network
+// runs plaintext on 9000. Operators using cloud ClickHouse on 9440
+// flip TLS=true and set Port=9440.
 type ClickhouseNativeEndpoint struct {
 	Hosts      []string `hcl:"hosts"`
+	Port       int      `hcl:"port,optional"`
+	TLS        bool     `hcl:"tls,optional"`
 	Credential string   `hcl:"credential,optional"`
 }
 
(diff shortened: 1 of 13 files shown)


arnauorriols:
Apply the feedback from https://github.com/denoland/clawpatrol/issues/74

arnauorriols:
Applied review feedback from #74 ("For PR #71") on top of `1c12321`. New commit `cbd6b2c`:

- **`uint64` VarUInts.** `readChVarUInt` / `appendChVarUInt` and `ChHello.{PacketType,VersionMajor,VersionMinor,ProtocolRevision}` retyped to `uint64`. Roundtrip test extended through `1<<50` and `^uint64(0)`.
- **UTF-8 validation in `chReadString`.** Rejects malformed bytes with `errors.New("clickhouse: invalid UTF-8 in string")`. New `TestChHelloRejectsInvalidUTF8` covers it.
- **Hard-fail on missing creds / `Secrets.Get` errors.** Soft-fail removed; every credential failure path now drops the connection and emits a structured `ConnEvent{Action:"error", Reason:...}`. The placeholder Hello no longer reaches upstream.
- **Schema/runtime split.** `clickhouse_native.go` now holds only the schema + `init()` registration; `HandleConn`, helpers, pipe, and `chReadHello` moved to `clickhouse_native_runtime.go`.
- **Dropped `chSerializeHello` / `chParseHello` wrappers.** The runtime calls `SerializeChHello` / `ParseChHello` directly.
- **`ConnRouteHosts` double-port fix.** Guarded with `net.SplitHostPort`; host strings already carrying a port are preserved verbatim. New `TestClickhouseConnRouteHostsNoDoublePort` covers bare, host:port, and IPv6 forms.
- **`strconv.Atoi` in `chHostPort`.** Replaces the inline digit walk; new `TestChHostPort` exercises IPv6, named-port, and no-colon edges.
- **`tlsConn.HandshakeContext(ctx)`** instead of `tlsConn.Handshake()`, so cancellation propagates.
- **Structured error `ConnEvent`s on every pre-pipe failure.** New `chEmitError` helper, used at every `return err` in `HandleConn` before the pipe starts: `wrong-family`, `wrong-endpoint-type`, `no-host`, `read-hello`, `credential-not-clickhouse-auth`, `secret-fetch`, `secret-empty`, `dial-upstream`, `tls-handshake`, `send-hello`, `forward-post-hello`.

Out of scope (per #74): TLS trust knob, per-query SQL parsing.

`go test ./config/plugins/endpoints/...` and `go test ./config/...` pass.

config/plugins/endpoints/clickhouse_native.go
```diff
@@ -1,36 +1,117 @@
...
+// dispatch is the active path for clickhouse_native.
+func (e *ClickhouseNativeEndpoint) ConnRouteHosts() []string {
+	return e.EndpointHosts()
+}
+
+func (e *ClickhouseNativeEndpoint) port() int {
+	if e.Port > 0 {
+		return e.Port
+	}
+	return 9000
```
arnauorriols, line 94: This default should take e.TLS into account

arnauorriols: [bot-reply] Default now picks 9440 when TLS=true, 9000 otherwise.

config/plugins/endpoints/clickhouse_native.go
```diff
@@ -1,36 +1,120 @@
...
+// protocol carries no SNI / Host header, so the gateway can't
+// dispatch on dst IP alone — dnsvip allocates a stable VIP per
+// hostname at policy build, intercepts the agent's DNS query for
+// that hostname, and the WG forwarder routes the resulting traffic
+// to handleVIPConn → this plugin's HandleConn.
+func (e *ClickhouseNativeEndpoint) RequiresVIP() bool { return true }
+
+// ConnRouteHosts mirrors EndpointHosts so legacy ConnIndex consumers
+// that key off the ConnRouter interface stay wired even though VIP
+// dispatch is the active path for clickhouse_native.
```
arnauorriols: Verify this comment is up to date.

arnauorriols: [bot-reply] Comment was stale. Verified the three claims against current code:

1. **"mirrors EndpointHosts"** — still true (`return e.EndpointHosts()`, line 81).
2. **"legacy ConnIndex consumers"** — incorrect framing now. The ConnIndex consumer for clickhouse_native is `tryDirectIPConn` in `main.go:1216`, which is *active* code, not legacy. It's the dispatch path for IP-literal-bound hosts.
3. **"VIP dispatch is the active path"** — only half-true. Active for hostname-bound hosts. For IP-literal hosts, `dnsvip.collectRequiredHosts` skips the entry (`dnsvip/dnsvip.go:337-339`), so the WG forwarder's direct-IP dispatch (`main.go:2136`, `tryDirectIPConn` → `connIdx.Lookup`) is the only path that reaches `HandleConn`.

Commit `2f629d1` ("clickhouse_native: direct-IP dispatch for IP-literal hosts") is what made the original comment stale.

Rewrote the comment to describe the current relationship — hostname entries ride VIP, IP-literal entries ride direct-IP dispatch, and `ConnRouteHosts` feeds both by mirroring `EndpointHosts`. Pushed as `c4b418f` on top of the PR head.

arnauorriols:
Feedback:
- When not specifying a port in the hcl, it defaults to 0 in the config, and eventually resolved by EndpointHosts() to 9000 or 9440, depending on tls config. However, in the VIP -> endpoint index, the port is registered as 0 (presumably), and then requests to <vip>:9440 don't work. They should: when the clickhouse server does not use a custom port, the endpoint hcl block shouldn't need to specify the port

arnauorriols:
[bot-reply]

Investigated the default-port VIP routing concern. The bug as described doesn't reproduce on the current branch tip — d58c6c5 plus the existing compile-time pipeline already covers it. Walked the path end-to-end:

1. **Port resolution.** `(*ClickhouseNativeEndpoint).port()` (config/plugins/endpoints/clickhouse_native.go:87) returns 9440 when `TLS=true && Port==0`, 9000 when plaintext.
2. **EndpointHosts.** `EndpointHosts()` calls `port()` and emits `host:9440` for the TLS-default case (clickhouse_native.go:53).
3. **Compile.** `compile.go:209` populates `ce.Hosts` via `extractHosts(body)` → `body.EndpointHosts()`. So the compiled endpoint carries `host:9440`.
4. **VIP index.** `dnsvip.collectRequiredHosts` (dnsvip/dnsvip.go:316) reads `ep.Hosts` and splits with `net.SplitHostPort` → `EndpointHit.Port = 9440`.
5. **Conn index.** `BuildConnIndex` (config/runtime/conn_route.go:71) consumes `ConnRouteHosts() = EndpointHosts()` so the byHost / byIP keys also carry `:9440`.
6. **Dispatch.** `handleVIPConn` (main.go:1187) matches `dstPort` against `h.Port`; both are 9440 → endpoint resolved → `dispatchConnEndpoint` runs with `matchedPort=9440`.

Verified with a runnable repro: a `ClickhouseNativeEndpoint{Hosts: ["ch.example.com"], TLS: true}` (Port left zero) routed through `RebuildFromPolicy` → `LookupVIP` returns `EndpointHit.Port == 9440`.

Added a regression test (`TestRebuildResolvesDefaultPortForTLSEndpoint`, dnsvip/dnsvip_test.go) so any future change that bypasses the EndpointHosts → ce.Hosts → SplitHostPort pipeline will fail loudly. No code change needed beyond the test.

**Related plugins flagged for separate follow-up** — none required. `postgres` exposes a single `Host` (no port-resolution helper at all, no TLS default), and `clickhouse_https` doesn't implement `RequiresVIP`/`ConnRouter`. So there's no other endpoint type with this shape today.

If you have a concrete repro where a TLS-default clickhouse_native endpoint fails to route — e.g. an HCL fixture and the agent's connect string — happy to dig further; the trace above suggests the failure mode is somewhere other than the index keying.

PR #75, opened 2026-05-05
Add layered glossary doc

arnauorriols:
Adds `site/doc/00-glossary.md` — a top-down glossary covering the big-picture building blocks (gateway, device, policy, agent), the HCL configuration vocabulary (endpoint, credential, rule, approver, profile, …), and the code-level plugin/runtime terms (`ConnEndpointRuntime`, `ConnRouter`, `HTTPCredentialRuntime`, …).

Also adds a one-line link at the top of `04-architecture.md` so a reader landing there has a path back to definitions.


arnauorriols:
This page should be be between onboarding and architecture in the menu

arnauorriols:
Renamed to `03a-glossary.md` so it sorts between onboarding and architecture; menu order verified locally via `npm run build`.

arnauorriols:
Do not call the top-level config file a "policy" just call it the global config file or something. We will change the name internally eventually

arnauorriols:
Renamed 'policy' → 'global config' for the file meaning; kept the block-kind usage.

arnauorriols:
Remove the global config section, not needed.

arnauorriols:
Removed the Global config section per review; cross-references audited.

site/doc/03a-glossary.md
```diff
@@ -0,0 +1,325 @@
...
+implementation.
+
+### Profile
+
+A named list of [endpoints](#endpoint) — the membership set assigned to
+a [device](#device). A device gets exactly the endpoints its profile
+names; rules ride along automatically because they're attached to
+endpoints. Profiles are how operators say "this dev box can reach
+GitHub and Notion; that build runner can only reach the package
+registry."
```
arnauorriols: Is this true? Afaik when the device performs an action that is not intercepted by an endpoint, it is passed through. What you write suggests the action is rejected. Don't take me for granted, verify in the implementation before doing any change

arnauorriols: Verified — you're right. The dispatcher passes through.

- `config/runtime/dispatch.go:17-38` (`HostEndpoint`): returns `nil` when no endpoint in the device's profile binds the host.
- `main.go:886-892` (HTTPS handler): on `nil` endpoint, falls through to `g.splice(...)` per the `defaults.unknown_host = passthrough` default.
- `main.go:917+` (postgres handler) and `main.go:1748-1752` (WG promiscuous forwarder for non-MITM ports): same passthrough fallback.
- `config/README.md:71` and `config/README.md:302` document `unknown_host = "passthrough"` as the default.

So profiles select which endpoints' rules apply to a device, not which hosts the device can reach. Reworded the Profile entry to drop the allowlist framing and call out the `defaults.unknown_host` fallback; applied the same fix to the Device entry's parallel 'can reach' phrasing. Appended as 4d75320.

PR #94, opened 2026-05-06
mac ext: bypass non-tunnel UDP via host socket

littledivy:
## Summary
- Per-process now claims TCP+UDP. Non-tunnel UDP flows are bypassed via a real host UDP socket: \`flow.readDatagrams\` ↔ \`sendto\`/\`recvfrom\` on \`AF_INET6\` dual-stack socket, per-datagram NWHostEndpoint preserved both directions.
- Sidesteps radar r.98382363 (false-return UDP race) — never returns false on UDP, never triggers the kernel detach race, no 30s Chrome QUIC stall.
- Per-process clawpatrol descendants' UDP/53 is tunneled through the gateway, enabling SSH endpoint plugin VIP DNS.
- Pattern adapted from Mozilla VPN's \`bypassudpflow.mm\`.

## Test plan
- [ ] \`clawpatrol run claude\` flows through gateway (TCP)
- [ ] \`clawpatrol run ssh user@vip-host\` resolves VIP via gateway DNS
- [ ] Chrome QUIC sites load fast (no stall)
- [ ] mDNS / Bonjour / NTP unaffected

### DIFF
--- a/macos/ClawpatrolExtension/Provider.swift
+++ b/macos/ClawpatrolExtension/Provider.swift
@@ -97,32 +97,31 @@ class TransparentProxyProvider: NETransparentProxyProvider {
     }
 
     private func applyNetworkSettings(completionHandler: @escaping (Error?) -> Void) {
-        // NETransparentProxy rejects port-specific rules ("X cannot be
-        // specified as the port for transparent proxy network rules").
-        // So per-process is TCP-only; whole-machine takes UDP+TCP and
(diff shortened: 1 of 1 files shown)


littledivy:
Hold — current shape rejects ALL non-tunnel UDP system-wide (mDNS, NTP, Zoom/RTP, system DNS), not just QUIC. Need per-destination decision but NEAppProxyUDPFlow doesn't expose remote pre-open. Investigating.

PR #102, opened 2026-05-06
clickhouse_native: sslmode for self-signed upstreams

arnauorriols:
## Summary

Adds an `sslmode` field on `clickhouse_native` endpoints so operators
can accept self-signed (or private-CA) upstream TLS certificates.
Without this, self-hosted ClickHouse deployments behind a private CA
fail the upstream handshake.

Mirrors postgres's vocabulary:

- `verify-full` (default) — full cert validation against system roots
  (current behaviour preserved)
- `require` — TLS but skip cert validation; trusts whatever cert the
  upstream presents

`prefer` / `disable` aren't accepted: clickhouse_native has no
in-protocol TLS negotiation, so the existing `tls` boolean already
covers hard on/off.

## Test plan

- [x] `go test ./config/...` (new unit test for the SSLMode →
      InsecureSkipVerify mapping; error fixture for invalid sslmode;
      golden update for the existing endpoint)
- [x] `go vet ./config/...` (clean for changed files; pre-existing
      unreachable-code warning in postgres.go untouched)
- [x] Schema round-trip preserved (full.hcl golden test)

### DIFF
--- a/config/plugins/endpoints/clickhouse_native.go
+++ b/config/plugins/endpoints/clickhouse_native.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/zclconf/go-cty/cty"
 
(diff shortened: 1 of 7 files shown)


arnauorriols:
This is a bad idea. I prefer to replicate clickhouse-client's `accept_invalid_certificate` flag

arnauorriols:
[bot-reply]
Pushed b6a4795 addressing this comment:
- Replaced the postgres-style `sslmode` string with an `accept_invalid_certificate` boolean (mirrors clickhouse-client's flag of the same name).
- Validation now only rejects the flag when `tls` is off; the enum check is gone.
- Renamed the validation testdata fixture, updated `gateway.example.hcl`, the runtime helper, the table-driven test, and the `full.want.json` golden.
- Rebased onto `origin/main` to clear the stale main.go / breadcrumb diff that #103 and #99 already merged.

PR #120, opened 2026-05-07
fix: assign action ids to persisted live events

magurotuna:
Ensure connection-level events that are persisted to the actions table get an action_id before DB insert and SSE fan-out. This makes live request rows clickable even when the request did not use the MITM HTTP start/end path.

Make this clickable: 
image


ry (approved):
LGTM

PR #142, opened 2026-05-08
docs: auto-generated HCL config reference

arnauorriols:
## Summary

Adds `tools/docgen`, a Go program that introspects the plugin registry under `config/plugins/` plus the operational config structs and renders `site/doc/15-config-reference.md` — a per-block, per-field reference for `gateway.hcl`. The site build pipeline (`site/build-docs.ts`) picks the file up automatically.

## Why hand-rolled (and not a third-party tool)

Surveyed the obvious candidates before reaching for reflection:

- **terraform-docs** — reads `.tf` files, looks for `variable {}` / `output {}`. Doesn't introspect Go structs, doesn't understand custom block kinds. No fit.
- **gomarkdoc / pkgsite / `go doc`** — Go API doc generators. They render package-level APIs, not HCL field references. Fields with `hcl:"..."` tags would render as Go struct fields, not HCL attributes; relabeling required either way.
- **hcldec / hclspec** — schema decoders, no doc emitter shipped.
- **terraform-plugin-docs** — Terraform-Provider-specific; assumes the Terraform Plugin Framework runtime.

The clawpatrol HCL is a plugin-dispatched DSL — `<kind> "<type>" "<name>" { ... }` blocks dispatch to a runtime-registered Go struct via `config.Register`. No off-the-shelf tool walks a custom plugin registry, so the generator is a small (~700 LOC across two files) Go program that reflects on plugin body structs and reads field comments via `go/parser`.

## What the generator does

1. Pulls in every plugin via the existing `config/plugins/all` blank-import side effect.
2. Walks `config.AllPlugins(kind)` for each kind, calls `plugin.New()` to get the body struct, and reflects to enumerate `hcl:"..."`-tagged fields (name, optional, label, block).
3. Parses `config/` and `config/plugins/{approvers,credentials,endpoints,rules}/` with `go/parser` to extract type and field doc comments.
4. Cross-references `Plugin.Refs` to annotate which fields are bare-name references to other entities.
5. Emits Markdown organized by kind, with attribute tables and tiny synthesized examples.

## Wiring

- `go run ./tools/docgen` writes the file.
- `go run ./tools/docgen --check` exits non-zero on drift (handy in scripts / pre-commit).
- `go test ./tools/docgen/...` runs the same drift check in `go test ./...`.
- `go generate ./config/plugins/all/...` re-runs the generator (matches the existing convention of plugin registration living in `all/`).
- `README.md` documents the regen command alongside the existing policy section.

## Test plan

- [x] `go run ./tools/docgen` produces `site/doc/15-config-reference.md`.
- [x] Re-running on an unchanged tree produces a byte-identical file.
- [x] `go test ./tools/docgen/...` passes; modifying the file by hand makes it fail with a pointer at the first differing line.
- [x] `go run ./tools/docgen --check` passes after generation, fails after manual edits.
- [x] `go generate ./config/plugins/all/...` rewrites the file in place.
- [ ] Verify the page renders correctly in the docs site (`npm run build` in `site/`) once a reviewer can run it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,16 @@ If you'd rather route every packet on the host through the gateway, pass `--whol
 
 Policy lives in `gateway.hcl`. You declare credentials, the endpoints they unlock, and the rules that decide what's allowed. References are bare names — no quotes, no kind prefix.
 
+The full per-block field reference lives at [`site/doc/15-config-reference.md`](site/doc/15-config-reference.md). It is auto-generated from the plugin registry under `config/plugins/`; regenerate after adding a plugin or changing an `hcl:"..."` tag:
+
+```
+go generate ./config/plugins/all/
+# or directly:
+go run ./tools/docgen
+```
+
+A `go test ./tools/docgen/...` drift check fails when the committed reference disagrees with the live schema.
+
 ```hcl
 credential "anthropic_oauth_subscription" "claude" {}
 credential "github_oauth"                 "github" {}
--- a/config/config.go
+++ b/config/config.go
@@ -135,7 +135,9 @@ func (f FrameworkAttrs) Ref(name string) string {
 	return f.Refs[name]
 }
 
-// Defaults captures the singleton defaults {} block.
+// Defaults holds gateway-wide fallbacks used when an `approver` block
+// or other policy entity does not pin its own value. Exactly one
+// `defaults {}` block per config.
 type Defaults struct {
 	UnknownHost    string `hcl:"unknown_host,optional"`
 	LLMFailMode    string `hcl:"llm_fail_mode,optional"`
@@ -144,8 +146,9 @@ type Defaults struct {
 	HumanOnTimeout string `hcl:"human_on_timeout,optional"`
 }
 
-// PolicyText is the lowered shape of a policy "<name>" {} block:
-// the heredoc text plus its source range for diagnostic messages.
+// PolicyText defines a named, reusable chunk of policy prose that
+// `llm_approver` blocks reference by name. The single `text` attribute
+// is typically a heredoc.
 type PolicyText struct {
 	Name string
 	Text string `hcl:"text"`
--- a/config/plugins/all/all.go
+++ b/config/plugins/all/all.go
@@ -3,6 +3,8 @@
 // the Terraform provider blank-import pattern and lib/pq drivers.
 package all
 
+//go:generate go run ../../../tools/docgen -o ../../../site/doc/15-config-reference.md
+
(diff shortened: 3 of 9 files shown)


arnauorriols:
1. Be explicit about blocks needing a "<name>"
2. instead of a reference column, on reference fields the type should be ref(<bock-type>)
3. Remove the auto-generated admonition
4. Remove references to config/README.md
5. Don't talk about Go. Users are writing hcl, not go. 
6. The "object" type should be more specific. For example, `[]credential`

arnauorriols:
[bot-reply]
Pushed 30263b4 addressing this comment:
- Block subheaders now spell out the `"<name>"` label, e.g. `### approver "human_approver" "<name>"`.
- Reference columns dropped; reference fields now declare their kind in the Type column as `ref(<kind>)` / `[]ref(<kind>)`.
- Removed the auto-generated admonition from the page header.
- Removed the `config/README.md` cross-link.
- Reframed the page intro + Type legend in HCL terms; rewrote the Go-flavored godocs on `RuleBody`, `Match`, `Approve`, `Defaults`, and `PolicyText`; added a small renderer-side helper that strips the leading Go-identifier prefix from any remaining type/field doc so the output reads as HCL prose.
- `object` is now specialised for the known cty.Value fields: `credentials` → `[]credential`, `approve` → `[]ref(approver)`. (`match` stays `object` since its shape is genuinely free-form per family — the per-family key list directly below the table already enumerates the valid keys.)

arnauorriols:
[bot-reply]
Pushed 303f763 addressing the remaining "object" type from this comment:
- The rule `match` field now renders as `block` instead of `object` in the config reference table (matches the existing override pattern used for `credentials` → `[]credential` and `approve` → `[]ref(approver)`).

Items 1–5 and the `credentials`/`approve` part of item 6 were already addressed in 30263b4.

arnauorriols:
sync with main, and fix CI

arnauorriols:
[bot-reply]
Pushed 5419933 addressing this comment:
- Rebased branch onto current `main` (clean rebase, 3 commits replayed).
- Fixed CI build/test failure: docgen referenced removed `match.KnownKeys` — switched to `facet.KnownKeys` after the match→facet package split landed on main.
- Regenerated `site/doc/15-config-reference.md` so the docgen drift test passes (was 575 lines, now 616 lines reflecting current plugins/families).

Verified locally: `go test ./...`, `go vet ./...`, `gofmt -l .`, and `www` `format:check` + `lint:github` + `build` all clean.

PR #144, opened 2026-05-08
site: rewrite approval-rules doc around HCL syntax

arnauorriols:
## Summary

Rewrites `site/doc/12-approval-rules.md` to match the actual config loader. The previous version described an SQLite/TypeScript rule subsystem that does not exist in this codebase.

The rewrite is structured for an operator writing `gateway.hcl`:

- **Rule families** — `http_rule`, `sql_rule`, `k8s_rule` and their endpoint-family bindings.
- **How to create a rule** — full HCL skeleton with required / optional / mutually-exclusive fields.
- **Examples** — pulled from `config/testdata/full.hcl`: allow/deny pairs, multi-credential `credential = X` selector, LLM proctor + human approver chain, priority overrides, K8s negation globs.
- **Matching semantics** — endpoint resolution, priority + first-match-wins (stable-sort tiebreaker), per-facet case sensitivity, list-with-negation rules, RE2 vs glob.
- **Gotchas** — load-bearing section. Each item is verified against the code:
  - 1 MiB HTTP body cap (`io.LimitReader(req.Body, 1<<20)`)
  - SQL `verb` matches only the first verb in multi-statement queries
  - `tables` / `functions` regex extraction limits (CTEs, schema-qualified names)
  - Postgres `Q` and `P` frames both evaluated; ClickHouse-native every Query
  - `statement_regex` is unanchored Go RE2, not anchored PCRE
  - `body_json` only matches JSON-shaped bodies; silent fail on parse errors
  - HTTP `headers` / `query` use substring matching, not equality
  - HTTP `Host` header isn't trustworthy at match time (overwrite happens after)
  - K8s verbs are HTTP-method-derived (no `watch` / `drain` / `cordon` / `evict`)
  - `defaults.unknown_host`, `llm_fail_mode`, `llm_cache_ttl`, `human_on_timeout` are schema-only today
  - `device {}` blocks are not yet wired in the loader
  - `approve` stages must be bare-name refs; struct stages are rejected
  - Negation glob lists are per-element

## Test plan

- [x] `cd site && npm run build` — markdown renders, doc page emits
- [x] All gotchas spot-checked against `config/match/match.go`, `config/plugins/rules/rules.go`, `config/plugins/endpoints/postgres.go`, `config/plugins/endpoints/clickhouse_native_runtime.go`, `config/runtime/dispatch.go`, `config/runtime/k8s_parse.go`, and `main.go`
- [ ] Reviewer eyeballs HCL examples for syntactic plausibility (snippets are extracted from `config/testdata/full.hcl`)


arnauorriols:
1. Remove references to config/README.md and config/testdata/full.hcl
2. Remove any references to implementation code. This is strictly speaking a user-facing doc page. This includes links but also mentions of code lines, code files, and code snippets like `sort.sliceTable...` or `io.ReadAll...`
3. In the snippets of examples of match keys for each family, some keys don't have a comment explaining the match semantics, but they should.
4. in the allow/deny http example, the billing approver is not defined anywhere
5. In various places, the document suggests you can use http_rule with kubernetes endpoints. Verify this assumption, and correct the docs if turns out to be false. 
6. Move the examples to after matching semantics. 
7. Don't try to explain how endpoints claim requests. Just explain generically that endpoints claim requests, emit actions of a specific family (with the family's facets), and then the rules match against the action's facets. Refer to the architecture doc page where the endpoint matching is explained in detail

arnauorriols:
[bot-reply]
Pushed f984fac addressing this comment:
- Removed all links to `config/README.md` and `config/testdata/full.hcl` (and other internal source paths).
- Stripped implementation-code references from the page: source-file links, line snippets like `sort.SliceStable(...)` and `io.ReadAll(io.LimitReader(...))`, the regex extractor patterns, and parenthetical Go-symbol mentions like `runtime.HostEndpoint` / `ParseK8sPath`.
- Added a comment on every match-key line in the per-family snippets (sql_rule `verb`/`tables`/`function`/`credential`, k8s_rule `namespace`/`credential`, etc.) so each key explains its semantics.
- Defined the `billing` approver inline in the HTTP allow/deny example, the multi-credential orb example, and the priority-override example so they are self-contained.
- Corrected http_rule's scope: verified against the codebase that `kubernetes` endpoints are family `k8s` and `http_rule` only accepts the `https` family. The doc now states `http_rule` does **not** bind to kubernetes endpoints, and the family table no longer claims "incl. kubernetes HTTP transport".
- Reordered sections so Matching semantics now precedes Examples (Rule families → How to create a rule → Matching semantics → Examples → Gotchas → Operational notes).
- Replaced the endpoint-resolution subsection (which described SNI peeking, profile scoping, and ConnIndex) with a short paragraph that just says endpoints claim requests, emit actions of a specific family with the family's facets, and rules match against those facets — pointing to the Architecture page for the claim mechanics.

arnauorriols:
1. Move the warning about http_rule not binding to kubernetes endpoints under the k8s_rule section. 
2. clarify if the sql.tables and sql.function match any of the tables/functions in the sql statement against any of the tables/functions in the array, or what

arnauorriols:
[bot-reply]
Pushed a4c2ea9 addressing this comment:
- Moved the "`http_rule` does not bind to `kubernetes` endpoints" warning out of the `http_rule` section and into the `k8s_rule` section, where readers looking up k8s rule semantics will encounter it.
- Clarified `sql_rule.tables` / `sql_rule.function` semantics: both are multi-valued facets (a single statement can name several tables/functions), and the rule fires when **at least one** extracted name satisfies the glob list — i.e. matches a positive glob and isn't knocked out by a negative one. Added two concrete examples (positive list, negative list) and noted that "every extracted name covered" requires `statement_regex` instead.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
 # Approval rules
 
-Claw Patrol classifies every outbound action it intercepts and asks: should
-this action be allowed, denied, deferred to a human, or deferred to an
-LLM judge? The decision is made by a single rule engine — the **approval
-rules** subsystem documented here.
+Rules are how an operator says "let this through", "block it", or
+"ask a human / LLM first". A rule is a block in `gateway.hcl` that
+targets one or more [endpoints](/docs/03a-glossary/#endpoint), names
+the requests it matches, and declares what to do.
```
arnauorriols: rewrite this paragraph. it is not clear what "name the requests it matches" means and clarify what "ask a human/LLM first" mean.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
-[Architecture](/docs/04-architecture/) and
-[Gateway](/docs/07-gateway/). For the agent trust boundary see
-[Security Model](/docs/11-security-model/). For LLM cost concerns
-referenced by the `require_llm` decision see
-[Token Usage](/docs/09-token-usage/).
+| Type        | Endpoint family  | What it matches                                |
+|-------------|------------------|------------------------------------------------|
+| `http_rule` | `https`          | HTTP method / path / query / headers / body |
+| `sql_rule`  | `sql` (`postgres`, `clickhouse_*`) | SQL verb / tables / functions / statement text |
+| `k8s_rule`  | `k8s` (`kubernetes` endpoints)     | Kubernetes resource / verb / namespace / name / query params |
```
arnauorriols: Instead of a table, narrate how the user can prevent agents from doing different things, like using certain http methods, or querying different URL paths, or using this or that sql verb, or querying certain k8s resources. Keep it introductory, as we give the proper explanation at the next section.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
 
+Rules don't cross families: a `sql_rule` never fires on an HTTPS
+request and vice versa. Each endpoint family declares its own match
+facets, and the matching rule type binds to those facets.
```
arnauorriols: Move this to the end of the rule families section.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
 
-## Where rules sit
+This page covers the operator's view: how to write a rule, what
+each facet does, what surprises to expect.
```
arnauorriols: change the word "surprises" for something less negative. Talk about "what behavior to expect on different situations", for example

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
+```hcl
+match = {
+  method        = "POST"                # HTTP verb (case-insensitive)
+  path          = ["/v1/refunds", "/v1/payouts"]   # URL path; glob OK
+  query         = { agent_id = "ci" }   # exact query-param match
+  headers       = { "x-tenant" = "prod" }           # substring header match
+  body_json     = { archived = true }   # JSON subset match
+  body_contains = "DELETE FROM users"   # raw substring (case-sensitive)
+  credential    = github-prod-pat       # bare-name ref to the dispatched credential
+}
```
arnauorriols, line 53: Fix the indentation of all the comments

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
 
-```typescript
-interface Rule {
-  id: string;
-  label?: string;
+```hcl
+match = {
+  method        = "POST"                # HTTP verb (case-insensitive)
+  path          = ["/v1/refunds", "/v1/payouts"]   # URL path; glob OK
+  query         = { agent_id = "ci" }   # exact query-param match
```
arnauorriols: double-check this is exact match or substring

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
+  headers       = { "x-tenant" = "prod" }           # substring header match
+  body_json     = { archived = true }   # JSON subset match
+  body_contains = "DELETE FROM users"   # raw substring (case-sensitive)
```
arnauorriols: Use another example that is more natural to http requests. This example produces noise to the reader, because we have the sql_rules.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,720 @@
...
+`clickhouse_native`). The matcher runs against every parsed SQL
+statement the agent sends.
 
-  source?: string;         // raw JSON of `when` (round-tripped to UI)
+```hcl
+match = {
+  verb            = ["select", "show", "explain"]  # first verb of the statement
+  tables          = ["users", "secret_*"]          # any extracted table satisfies the list; glob OK
+  function        = ["pg_read_file", "dblink_*"]   # any extracted function satisfies the list; glob OK
+  statement       = "*COPY*FROM PROGRAM*"          # whole-statement glob
```
arnauorriols: I don't understand the purpose of using globs in this example. Can you figure out a case where the glob utility is more obvious?

arnauorriols (commented):
feedback

arnauorriols:
[bot-reply]
Pushed e31d4fc addressing this comment:
- Rewrote the intro paragraph to spell out the three outcomes (forward / reject / route through approver stages) instead of the vague "ask a human / LLM first".
- Replaced "names the requests it matches" with "describes which requests it applies to (the `match` block)" and explicitly names the outcome fields (`verdict` / `approve`).

arnauorriols:
[bot-reply]
Pushed 869b5d1 addressing this comment:
- Replaced the three-rule type table with a short narrative intro framed around what an operator can restrict per family (HTTP methods/paths, SQL verbs/tables, k8s resources/verbs/namespaces), with a pointer that the next section gives the full detail.

arnauorriols:
[bot-reply]
Pushed 6adef21 addressing this comment:
- Moved the "Rules don't cross families..." paragraph to the end of the rule families section (after `k8s_rule`).

arnauorriols:
[bot-reply]
Pushed 5aaf844 addressing this comment:
- Reworded the intro line: "what surprises to expect" → "how rules behave in different situations" (site/doc/12-approval-rules.md:21).

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed b25e902 addressing this comment:
- Replaced the `statement = "*COPY*FROM PROGRAM*"` example with `statement = "ALTER SYSTEM *"`. The new example shows glob utility more clearly: `verb = "alter"` alone is too broad (it also matches ALTER TABLE, ALTER INDEX, etc.), and there is no other facet that captures the specific multi-word prefix, so the wildcard is doing real work pinning the rule to ALTER SYSTEM specifically.

arnauorriols:
[bot-reply]
Pushed 8a66574 addressing this comment:
- Replaced the `body_contains = "DELETE FROM users"` example with `body_contains = "BEGIN PRIVATE KEY"` — a substring match that's natural for HTTP request bodies (catching leaked PEM material in any field) and doesn't overlap with the SQL examples below.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed c6657fd addressing this comment:
- Query (and headers) both go through `config/match/sliceOverlap`, which matches when `want == got` OR `strings.Contains(got, want)` — i.e. effectively substring matching, not exact match.
- Updated the comment on the `query` line from `exact query-param match` to `substring query-param match`, so it now matches `headers` and the actual behavior.

arnauorriols:
[bot-reply]
Pushed aa46a13 addressing this comment:
- Aligned the inline `#` comments in the `http_rule` match example to a single column (cols were 41/52/41/53/41/41/41 → all 53).
- Also bumped the `endpoint` / `# endpoints` lines in the rule-skeleton example so their inline `#` lines up with the rest of that block at column 42.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed e31d4fc addressing this comment:
- Rewrote the intro paragraph to drop the unclear "names the requests it matches" / "ask a human / LLM first" phrasings.
- Now spells out the three outcomes (forward, reject, route through approver stages) and names approvers explicitly (human via dashboard/Slack, or LLM against a policy).
- Replaced "names the requests it matches" with "describes which requests it applies to (the `match` block)" and shows the verdict/approve syntax inline.

arnauorriols:
[bot-reply]
Pushed 00e41a8 addressing this comment:
- Reframed the intro paragraph so each rule type is described in terms of what it prevents the agent from doing (HTTP methods/paths, SQL verbs/tables, k8s resources/verbs/namespaces) rather than what it allows.

arnauorriols:
I like the `body_contains = "BEGIN PRIVATE KEY"` example replacement, but I don't see this change commited. Please do

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-  return true;
-}
+Naming: every named entity in `gateway.hcl` (approvers, credentials,
+endpoints, rules, profiles) shares **one flat namespace**. References
+are bare names — never `endpoint.foo` or `credential.foo`. A
+duplicate name across kinds is a load error.
+
+A rule that names a wrong-family endpoint, an undeclared name, or a
+typo in a match key fails at load time with an error pointing at the
+offending block. Save iteratively: load, fix, repeat.
```
arnauorriols: Remove the last bit "Save iteratively: load, fix, repeat"

arnauorriols:
[bot-reply]
Pushed da5b2cf addressing this comment:
- Restored the `body_contains = "BEGIN PRIVATE KEY"` example on line 51. The earlier commit 8a66574 made this replacement, but the follow-up alignment commit aa46a13 was applied against a pre-replacement copy of the snippet and re-introduced `"DELETE FROM users"`. The change is now back in place (and the column alignment from aa46a13 is preserved).

arnauorriols:
[bot-reply]
Pushed a385f82 addressing this comment:
- Removed the trailing "Save iteratively: load, fix, repeat" sentence from site/doc/12-approval-rules.md.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-2. Sort by `priority` descending.
-3. Within the same priority, sort by **specificity** descending:
-   `integration` (4) > `profile` (2) > `plugin` (1) > global (0); the
-   weights add (a plugin+profile-scoped rule has specificity 3).
-4. Walk the sorted list and return the first rule whose `when` matches
-   the action.
+- If the list has any **positive** entries, at least one must match.
+- Any **negative** entry that *does* match disqualifies the whole
+  list (returns false).
+- A list with only negatives fires when none of the negatives match.
```
arnauorriols: This is hard to grok. one of the problems is that you use a different expression to express that the rule matches. Consolidate this into always saying "the rule is triggered" or "the rule triggers" or variants. The expression "disqualifies the whole list" is particularly bad.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+### Case sensitivity, by facet
 
-If no rule matches, the engine falls back to the value of the
-`CLAWPATROL_DEFAULT_POLICY` environment variable:
+| Facet                      | Case sensitivity |
+|----------------------------|------------------|
+| HTTP `method`              | insensitive      |
+| HTTP `path`, `query`, `headers` | sensitive   |
+| HTTP `body_contains`       | sensitive        |
+| SQL `verb`                 | insensitive (and folded to lower before compare) |
```
arnauorriols: Remove the folded to lower before compare bit. If it is insensitive, I don't care about how

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+| SQL `tables`, `function`   | sensitive (matched against lower-cased extractions) |
+| SQL `statement`, `statement_regex` | sensitive against lower-cased statement |
```
arnauorriols: Put tables, function, statement and statement_regex in the same row, same explanation.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+match keys in the same case the parser will produce (lower).
 
-Default-policy verdicts have no `ruleId`. They are reported as
-`source: "default"` in the action log.
+### `credential = X`
 
+Fires when the request is *dispatched against* the named credential.
+On a multi-credential endpoint, the agent picks the credential by
+embedding the configured placeholder; on a single-credential endpoint,
+every request matches the only credential.
```
arnauorriols: This paragraph is unnecessary. I'm learning about rules, not about how credentials are chosen.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-| `deny_writes` | Allow only `read` actions; deny `write`, `mutate`, and `destructive`. The verdict carries `source: "default"` and `reason: "default-deny for non-read actions"`. |
+For SQL, the parser lower-cases the statement before extracting verbs,
+tables, and functions — so `tables = "Users"` will never fire. Write
+match keys in the same case the parser will produce (lower).
 
-Default-policy verdicts have no `ruleId`. They are reported as
-`source: "default"` in the action log.
+### `credential = X`
 
+Fires when the request is *dispatched against* the named credential.
```
arnauorriols: the concept of dispatching a request against a named credential is confusing. Find a better way to express the same thing.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+- `approve = [a, b, c]` — stages run in order, **all must allow**.
+  The first non-allow stage short-circuits and is returned. A stage
+  that returns no decision (e.g. timeout) is treated as deny.
 
-1. **Top-level fields** -- bare names map to top-level fields of the
-   record:
+LLM stages call the configured LLM approver via its bound credential.
+Human stages park the request: the dashboard's pending-approvals page
+plus, optionally, a Slack channel ping if the approver block has a
+`credential` reference to a `slack_tokens` credential.
```
arnauorriols: Talking about "LLM stages" and "Human stages" is confusing. Use the "approver" concept instead of "stage".

Also, the slack HITL does not do a "ping", it sends a proper approval message with "approve" and "deny" buttons. The HITL part needs to be rewriten, probably with multiple sentences.

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-   ```
-   type        -- e.g. "http.request"
-   summary     -- short human-readable summary
-   description -- longer description (optional)
-   verb        -- protocol verb (e.g. "DELETE", "SELECT")
-   sensitivity -- "read" | "write" | "mutate" | "destructive"
-   primary     -- name of the primary facet (e.g. "http")
-   tags        -- string[] (any-of semantics for `contains` / `in`)
-   ```
+If no rule matches, the request is **allowed** — there is no global
```
arnauorriols, line 259: Double-check if there isn't a global config in the hcl that switches the "allow-by-default" to "deny-by-default"

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-  "any": [
-    { "http.method": "DELETE" },
-    { "http.method": "PATCH" }
-  ],
-  "not": { "http.url": { "contains": "/sandbox/" } }
+rule "http_rule" "orb-prod-writes" {
+  endpoint = orb
+  match    = { credential = orb-prod-key, method = ["POST", "PUT", "PATCH"] }
+  approve  = [billing]
 }
```
arnauorriols, line 350: Missing the method = "DELETE" -> deny rule for prod

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+paged. If either says `deny`, the request is rejected with the
+reason returned by the rejecting stage.
+
+The bare name `dashboard` is a built-in approver: `approve =
+[dashboard]` parks the request on the dashboard's pending-approvals
+view without paging any channel.
 
-To register new schemas, ship them on a plugin -- there is no separate
-schema registration API. The catalog is rebuilt on every API call.
+### Priority override pattern
```
arnauorriols: Remove this example, it does not add much extra value

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
+2. **Delete** by removing the block and reloading.
 
-`src/approvals/rules.test.ts` is the canonical reference for how each
-operator and combinator behaves; treat the test cases as runnable
-examples.
+Both require a config reload (the gateway re-reads `gateway.hcl` on
+SIGHUP / dashboard save).
 
-### Rolling back a rule
+### Where matched rules show up
```
arnauorriols: Remove this section

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
 
-There is no `clawpatrol rules test` CLI today. Use the dashboard preview
-(`POST /api/rules/preview`) for dry runs against historical actions.
+There is no `clawpatrol rules test` CLI today. To smoke-test a rule
+against your real config: load the gateway locally, point an agent
+at it, and watch the dashboard's per-action log. The log line
+carries the matched rule name so you can confirm which rule fired.
 
-For unit tests, the engine itself is a pure function:
+### Rolling back a rule
```
arnauorriols: Remove this section

site/doc/12-approval-rules.md
```diff
@@ -1,819 +1,724 @@
...
-  in-memory `pending` map, every SSE listener on
-  `/api/pending-actions/stream` is notified, and the framework parks on
-  a `Promise` until either the dashboard PendingActions page resolves
-  the entry, or the timeout fires.
-- **third-party** (Slack, email, etc.): the framework calls the
-  plugin's `request()` method. The plugin owns its correlation state
-  (a webhook from Slack arrives via `dispatchWebhook` in
-  `src/approvals/webhooks.ts`, mounted at
-  `/api/approvers/<plugin-id>/<path>`).
+## Gotchas
```
arnauorriols: Create a GitHub issue with the content of this section, and remove it from this document.

arnauorriols (commented):
feedback

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed 53ed591 addressing this comment:
- Removed the "(and folded to lower before compare)" parenthetical from the SQL `verb` row.

arnauorriols:
[bot-reply]
Pushed 9baf6eb addressing this comment:
- Merged the SQL `tables`/`function` and `statement`/`statement_regex` rows into a single row with one shared explanation ("sensitive (matched against lower-cased SQL)").

arnauorriols:
[bot-reply]
Pushed d272154 addressing this comment:
- Rewrote the list-with-negation bullets to use a single consistent phrasing — "For the rule to trigger" — instead of mixing "must match", "disqualifies the whole list (returns false)", and "fires when none match".
- Collapsed the three bullets into two clearer conditions: positives require at least one match, no negative may match. The all-negatives case is now covered by the second bullet alone (vacuous first bullet), so no third bullet is needed.
- Dropped the "disqualifies the whole list (returns false)" wording entirely.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed e195142 addressing this comment:
- Removed the paragraph about how the agent picks a credential on multi- vs single-credential endpoints; the section now keeps only the "what the matcher matches against" explanation.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Addressed by e195142 (dropped the "Fires when the request is *dispatched against* the named credential. On a multi-credential endpoint, the agent picks the credential by embedding the configured placeholder; on a single-credential endpoint, every request matches the only credential." paragraph). That commit was pushed for a sibling review comment that called the same paragraph unnecessary, but it also removes the "dispatching a request against a named credential" phrasing flagged here.

What's left in the `credential = X` section is just:

> `credential` does not look at the request body or headers — it matches the resolved credential name, not the credential's secret contents.

So the match-key is now described in terms of *what it matches against* (the resolved credential name) rather than the abstract "dispatched against" concept. If you want the section to do more — e.g. an explicit one-liner like "Matches the credential the request was sent through" before the "doesn't look at body/headers" caveat — let me know and I'll add it.

arnauorriols:
[bot-reply]
Pushed 1d0dafb addressing this comment:
- Replaced "LLM stages" / "Human stages" wording with "LLM approvers" / "Human approvers" throughout the doc; renamed remaining "stage" references in the outcome-dispatch and proctor-chain sections.
- Rewrote the Slack HITL paragraph: Claw Patrol posts a proper approval message to the channel (dashboard link by default; with `interactive = true` the message embeds in-channel "approve" / "deny" buttons), not just a ping.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed a385f82 addressing this comment:
- Dropped the trailing "Save iteratively: load, fix, repeat" line from `site/doc/12-approval-rules.md`.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Confirmed — no such global switch exists. The `defaults {}` block currently accepts only:

- `unknown_host` (`passthrough`/`deny`) — host-routing knob: applies when the SNI/host isn't bound to any endpoint in the active profile, not when no rule matches inside a bound endpoint.
- `llm_fail_mode`, `llm_cache_ttl`, `human_timeout`, `human_on_timeout` — approver-stage tunables.

The rule-matching default is hard-coded across transports: `runtime.MatchRequest` returns `nil` when no rule fires, and every caller (HTTPS handler at `main.go`, postgres and clickhouse native handlers) treats that as implicit allow. There is no HCL knob to flip it — the per-endpoint `priority = -100, verdict = "deny"` catch-all is the documented way.

arnauorriols:
[bot-reply]
Addressed in aa46a13 (`doc: align inline HCL comments in approval-rules`):
- Aligned all inline `#` comments in the `http_rule` match example to column 53
- Bumped the `endpoint` / `endpoints` lines in the rule-skeleton example so their inline `#` lines up with the rest of that block at column 42

The other example blocks (`sql_rule`, `k8s_rule`) were already self-consistent (col 52 and col 47 respectively).

arnauorriols:
[bot-reply]
Pushed 8b1a536 addressing this comment:
- Removed the "Rolling back a rule" section from `site/doc/12-approval-rules.md`.

arnauorriols:
[bot-reply]
Pushed 4b0b09e addressing this comment:
- Removed the "Where matched rules show up" section from site/doc/12-approval-rules.md.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed 7523b37 addressing this comment:
- Added `orb-prod-deletes` rule to the orb-prod example: `method = "DELETE"` with `verdict = "deny"`, mirroring the Stripe DELETE-deny pattern shown above.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Double-checked the loader — there is no global "deny-by-default" switch.

- The top-level `Gateway` struct (config/config.go:20) only carries operational fields (listen, info_listen, public_url, ca_dir, resolver, log_path, oauth_dir, dashboard_secret, insecure_no_dashboard_secret, session_keep, and the nested `gateway {}` tailscale block). No verdict-default field.
- The singleton `defaults {}` block decodes into `Defaults` (config/config.go:107) with exactly five keys: `unknown_host`, `llm_fail_mode`, `llm_cache_ttl`, `human_timeout`, `human_on_timeout`. None of them flip allow-default to deny-default.
- At dispatch time, `runtime.MatchRequest` returns nil when no rule fires, and the HTTP path in `main.go:1659` skips both the approve-chain and the `verdict == "deny"` branches when `cr == nil`, falling through to forward. Same shape for the Postgres / ClickHouse dispatch sites.

So the doc's claim ("If no rule matches, the request is allowed — there is no global default-deny") is correct as written. No change needed.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Pushed 1d0dafb addressing this comment:
- Replaced "stage"/"stages" terminology with "approver"/"approvers" throughout `site/doc/12-approval-rules.md` (intro, table, runtime semantics, examples, gotchas).
- Rewrote the Slack HITL paragraph: the dashboard always lists the request; if the approver has a `slack_tokens` credential Claw Patrol also posts an approval message that links back to the dashboard, and `interactive = true` embeds in-channel "approve"/"deny" buttons.

arnauorriols:
[bot-reply]
Pushed 9baf6eb addressing this comment:
- Merged the `tables`/`function` and `statement`/`statement_regex` rows in the SQL case-sensitivity table into a single row with one explanation ("sensitive (matched against lower-cased SQL)").

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Addressed in 8a66574: replaced the `body_contains = "DELETE FROM users"` example with `body_contains = "BEGIN PRIVATE KEY"`, an HTTP-native substring that doesn't overlap with what `sql_rule` already covers.

arnauorriols:
[bot-reply]
This was already addressed on this branch — no new commit was required:

- `869b5d1` (site: replace approval-rules type table with narrative intro) — replaced the original table with a narrative paragraph.
- `00e41a8` (site: reframe approval-rules intro around restriction) — reworded that narrative to frame each rule type as preventing the agent from doing things (HTTP methods / URL paths, SQL verbs / tables, k8s resources / verbs / namespaces), keeping it introductory and pointing readers at the next section for the full explanation.

The current text at `site/doc/12-approval-rules.md:13-18` matches the spirit of the comment.

arnauorriols:
[bot-reply]
Pushed a1e387d addressing this comment:
- Removed the "Priority override pattern" example section from `site/doc/12-approval-rules.md`.

arnauorriols:
[bot-reply]
Pushed 4486993 addressing this comment:
- Extracted the full `## Gotchas` section content into issue #213 for triage (each entry needs a code/UX or doc-move decision).
- Removed the section (lines 439–668) from `site/doc/12-approval-rules.md`.
- Patched the now-dangling `see Gotchas below` cross-reference in the Kubernetes negation glob example.

arnauorriols:
sync with main again, and fix the git conflict. Also remove the Operational Notes section, entirely

arnauorriols:
[bot-reply]
Pushed ae21657 addressing this comment:
- Merged latest origin/main into the branch; resolved the conflict in `site/doc/12-approval-rules.md` (kept the branch's HCL-rewritten version, including the `orb-prod-deletes` DELETE deny rule).
- Removed the `## Operational notes` section (and its `### Testing rules` subsection) entirely from `site/doc/12-approval-rules.md`.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
PR #144 is already merged, and the file has since been renamed and rewritten on `main` (now `site/doc/approval-rules.md`). The current main includes the missing rule at line 354:

```hcl
rule "orb-prod-deletes" {
  endpoint   = orb
  credential = orb-prod-key
  condition  = "http.method == 'DELETE'"
  verdict    = "deny"
}
```

No follow-up change needed — the example was rewritten around CEL conditions and the DELETE deny rule for prod is present. Closing this thread.

arnauorriols:
[bot-reply]
This comment was posted at 2026-05-11 20:23:17Z; the PR merged 22 minutes later (fa24176) and that merge already:

- Synced with main and resolved the conflict on `site/doc/12-approval-rules.md`.
- Removed the `## Operational notes` section in its entirety (deleted in the rewrite diff).

Since then, `site/doc/12-approval-rules.md` has been renamed to `site/doc/approval-rules.md` (commit 601b027) and rewritten to use CEL expressions (commit 14f8c65, PR #219). No further action on this merged PR.

site/doc/12-approval-rules.md
arnauorriols: [bot-reply]
Double-checked — confirmed, no such global toggle exists.

The `Gateway` HCL schema (`config/config.go:23-73`) exposes top-level scalars like `unknown_host`, `llm_fail_mode`, `human_on_timeout`, etc., but nothing that flips the no-rule-matched default. The runtime path is unconditional:

- HTTPS (`main.go:1680`): `cr := runtime.MatchRequest(...)`; the deny branch is `if cr != nil && cr.Outcome.Verdict == "deny"`, so a `nil` rule falls through to forward.
- Postgres (`config/plugins/endpoints/postgres.go:447-456`): on `cr == nil`, the comment literally reads "No rule matched — implicit allow" and emits an `allow` event.
- Clickhouse (`config/plugins/endpoints/clickhouse_native_runtime.go:680`): same shape — "no rule fires, or the matched rule allows" returns `("", "")`.

The closest neighbours are scoped, not global:
- `unknown_host` decides what happens when the SNI/authority doesn't resolve to any endpoint at all (a different layer — `runtime.HostEndpoint` returning nil, not `MatchRequest` returning nil).
- `llm_fail_mode` / `human_on_timeout` control approver-failure behaviour for a specific stage, not the default verdict.

So the per-endpoint `priority = -100, verdict = "deny"` catch-all really is the only way to invert it today.

arnauorriols: [bot-reply]
Confirmed: there is no global config that flips allow-by-default to deny-by-default.

The only singleton in the grammar is `defaults {}` (`config/config.go:139`), which exposes `unknown_host`, `llm_fail_mode`, `llm_cache_ttl`, `human_timeout`, `human_on_timeout`. None of these affect what happens when a request hits a known endpoint and no rule fires.

`defaults.unknown_host` is a sibling concern: it controls the **unknown host** case (host not bound to any endpoint) — values `passthrough` | `deny`. Not the no-rule-matched case.

The no-rule-matched behavior is hardcoded as implicit allow:
- HTTPS path: `main.go:1671` — `cr := runtime.MatchRequest(...)`; if `cr` is nil, the request is forwarded with no verdict gate.
- Postgres path: `config/plugins/endpoints/postgres.go:447-457` — explicit comment "No rule matched — implicit allow".
- Clickhouse path: same shape at `config/plugins/endpoints/clickhouse_native_runtime.go:698`.

So the doc's claim ("If no rule matches, the request is allowed — there is no global default-deny. Add a `priority = -100, verdict = "deny"` catch-all per endpoint to invert this.") is accurate.

PR #173, opened 2026-05-09
Avoid storing raw dashboard secret in cookie

ry:
## Summary
- store a derived dashboard cookie token instead of the raw dashboard_secret
- validate cp_dash against the derived token
- reject cookies containing the raw dashboard_secret
- add regression tests for cookie contents and validation

## Tests
- go test ./...
- go vet ./...


ry:
Closing per maintainer feedback; not pursuing this change right now.

PR #194, opened 2026-05-11
doc: dry-run design proposal (discussion)

arnauorriols:
**Status: design discussion, no implementation yet.**

Design doc for a new `clawpatrol dry-run --config <hcl>` subcommand
that tests a candidate HCL policy against the live gateway for a
single session, with no upstream forwarding. Full proposal in
`doc/dry-run-design.md`.

I'm opening this draft so we can lock the open questions in §3
before any code lands — the central one (§3.A) is a real design
fork.

## TL;DR

- **Single dispatch point** exists for HTTPS (`runtime.MatchRequest` at `config/runtime/dispatch.go:45`, called at `main.go:1632`). v1 should be HTTPS-only; postgres / clickhouse_native walk rules themselves.
- **Session keying is the problem.** PeerIP is the universal session key today (`main.go:2026`), but it's **not 1:1 with a `clawpatrol run` invocation**: Linux `run` reuses the host's `wg.conf`, macOS funnels every wrapped process through one shared tunnel. Two concurrent `run`s on one host share one PeerIP.
- **Recommended fix:** mint an **ephemeral WG peer per dry-run invocation**, so each session gets a unique PeerIP and the existing dispatch path needs no schema change. Details in §3.A.
- **Config compile is reusable.** `loadConfig` / `config.Compile` are pure, safe to call from a request handler, ~10–100 KB per policy.
- **Upload path:** mirror existing `/api/config/preview` shape. Device-scoped auth (any onboarded peer can test their own session), not the admin `dashboard_secret`.
- **Verdict synthesis:** per-protocol minimal OK for `allow`/`passthrough`; auto-allow + `wouldRequireApproval: true` tag for `approve` so HITL chains don't actually page humans.
- **Whole-machine:** reject at CLI-time; no per-process boundary to scope to.

## Open questions

Listed in `doc/dry-run-design.md` §3. The ones I'd most like an
answer on before writing code:

1. **§3.A — session keying.** Ephemeral WG peer per dry-run (my
   recommendation) vs. in-band `X-Clawpatrol-DryRun` header
   (HTTPS-only) vs. src-port marking?
2. **§3.C — `approve` verdict** under dry-run: auto-allow + tag
   (my recommendation), auto-deny, or run LLM stage only?
3. **§3.E — auth on the upload endpoint:** device-scoped (my
   recommendation) or admin `dashboard_secret`?
4. **§3.B — confirm whole-machine reject** with a clear error
   message.

Once these settle I'll open a follow-up implementation PR scoped
to: ephemeral-peer endpoint, the dispatch hook, HTTPS verdict
synthesis, event tagging, and tests.

## Acceptance criteria for the eventual implementation PR

(Lifted from the bead `cl-d9d` and slightly tightened in §4 of the
doc.)

- `clawpatrol dry-run --config <hcl> <cmd>` runs `<cmd>` with every
  HTTPS action matched against the candidate config; no upstream
  traffic sent; synthesised responses sufficient for the wrapped
  agent to keep running; per-session config dropped at session end.
- PR description names the verdict-synthesis decisions made here
  and links back to this thread.
- Tests cover per-session config attach/drop and the HTTPS dry-run
  request path end-to-end.

cc whoever owns the gateway runtime — would love a sanity check on
§3.A in particular.

### DIFF
--- a/doc/dry-run-design.md
+++ b/doc/dry-run-design.md
@@ -6,26 +6,33 @@
 ## Goal
 
 Add a CLI subcommand that runs a candidate HCL policy against a
-**fixture suite of actions with expected verdicts** and reports
-pass/fail, like a test runner:
+**recorded actions file** and reports any verdicts the new policy
+would change:
 
 ```
-clawpatrol test --config ./candidate.hcl --suite ./actions.json
+clawpatrol test --config ./candidate.hcl --actions ./actions.json
 ```
 
-The suite is a list of `{action, expected_verdict}` pairs. The
-command compiles the candidate config in-process and runs each
-action through `runtime.MatchRequest`, comparing the returned
-verdict against the expected one. Exit status is non-zero on any
-mismatch.
+The actions file is a list of recorded gateway actions. Each
+entry carries the request the gateway saw and the verdict it
+produced — `action` (`allow` / `deny` / `approve` /
+`passthrough`), the name of the matched `rule`, and the
+`reason`. Nothing in the format is test-specific: it is the same
+shape the gateway logs for live traffic, just persisted to a
+file.
 
-To keep the suite-authoring burden low, the dashboard grows a
-**"Download as test suite"** button on its recent-actions view.
-The button emits a ready-made suite file populated from the
-actions the gateway has actually seen, each pre-filled with the
-verdict it produced under the live config. Operators run that
-file as-is to lock in current behaviour, or edit individual
-entries to drive a policy change.
+`clawpatrol test` compiles the candidate config in-process, runs
+each request through `runtime.MatchRequest`, and reports any
+entry whose new verdict differs from the recorded one. Exit
+status is non-zero on any diff.
+
+To keep the authoring burden low, the dashboard grows a
+**"Download actions"** button on its recent-actions view. The
+button emits an actions file populated from the actions the
+gateway has actually seen, each carrying the verdict it produced
+under the live config. Operators run that file as-is to lock in
+current behaviour, or edit individual entries to drive a policy
+change.
 
 The point is iteration speed and CI: today an operator changes a
 rule, pushes a full config reload, watches live traffic, and
@@ -48,7 +55,7 @@ That's the entire surface this subcommand needs to drive.
 
 A `match.Request` is plain data: `Family`, `Method`, `URL`,
 `Headers`, `Body`, `PeerIP`, parsed path. It serializes cleanly
-to JSON, which is what the suite format is.
+to JSON, which is what the actions-file format is built on.
 
 ### 1.2 Config compile is reusable from a CLI process
 
@@ -63,12 +70,18 @@ CLI process — no gateway required.
 The event sink (`web.go:1582`, `Sink` / `Event`) buffers the
 last 500 actions in-memory and persists to SQLite. The
 dashboard's `/api/state` and `/api/events` endpoints already
-expose this stream, and `Event` carries everything the suite
-needs: `Method`, `Path`, `ReqHeaders`, `ReqBody`, `Host`,
-`Mode`, plus the produced verdict in `Action` + `Reason`.
+expose this stream, and `Event` carries most of what the
+actions file needs: `Method`, `Path`, `ReqHeaders`, `ReqBody`,
+`Host`, `Mode`, plus the produced verdict in `Action` +
+`Reason`. The matched rule name is **not** currently on
+`Event` — `MatchRequest` returns the `*CompiledRule` (which
+has `.Name`), but the call sites at `main.go:1638` etc. drop
+it before logging. A small extension of `Event` (`Rule string`)
+populated at the existing dispatch sites is enough; no new
+plumbing.
 
-A "download these as a test suite" endpoint is a re-shape of an
-existing dataset, not a new pipeline.
+A "download these as an actions file" endpoint is then a
+re-shape of an existing dataset, not a new pipeline.
 
 ### 1.4 Subcommand wiring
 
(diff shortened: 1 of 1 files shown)


arnauorriols:
New idea: clawpatrol dry-run starts an ephemeral gateway in dry-mode, which does not need credentials because it does not proxies requests. It takes an array of actions and rule approval results. It also takes a config file. It runs all the actions and expected rule results feom the array, and reports the result as a test suite

arnauorriols:
[bot-reply]
This is a meaningfully different shape than the current proposal — worth surfacing the comparison so you can decide between them (or both).

**Current proposal (live-session dry-run):** wrap a real `<cmd>`, watch its actions, match each against the candidate config attached to that session. Answers *"what does my real agent do under this policy?"*

**Your new idea (declarative test suite):** caller supplies `{actions[], expected_verdicts[], config}`; an in-process / ephemeral gateway compiles the config and runs `MatchRequest` over the action list, reports pass/fail per action like a test suite. Answers *"do my hand-written fixtures produce the verdicts I expect?"*

Tradeoffs I see:

- **+** Massively simpler. §3.A (session keying), §3.B (whole-machine), §3.C (response synthesis), §3.E (auth — could be none if it runs in-process), §3.F (TTL), §3.G (event contamination) all collapse or disappear. It's `config.Compile()` + a loop over `MatchRequest`.
- **+** Reproducible — the suite checks into the repo and runs in CI.
- **+** No live gateway needed at all — could be `clawpatrol policy test --config foo.hcl --suite tests.json` running entirely in the CLI binary, no server-side endpoint, no auth.
- **−** Only verifies fixtures the operator wrote by hand. The live version exercises the agent's real call patterns, including ones nobody thought to enumerate.

These read as **complementary** rather than substitutes — declarative `policy test` is a CI gate (*"did I break my rules?"*), live `dry-run` is exploratory (*"what will my agent actually do?"*). They share `MatchRequest` but nothing else.

How would you like to proceed?

1. Replace this design with the declarative form; drop live-session work.
2. Add the declarative form as a sibling §6 in this doc; decide later.
3. File the declarative form as its own bead; keep this PR scoped to live-session dry-run.

Leaning (3) — independently valuable, very different implementation surfaces — but happy to do (2) if you'd rather keep both options in one place while thinking.

arnauorriols:
Replace the current design with the declarative form. Call it `clawpatrol test`. There should be a button in the dashboard to download the list of recent actions, together with their rule output. Users can use this to download a ready-made list of actions that can be used as is, or modified as needed

arnauorriols:
[bot-reply]
Pushed 9a86aa8 addressing this comment:
- Rewrote `doc/dry-run-design.md` around the declarative form, named `clawpatrol test`. CLI compiles the candidate config in-process and runs each entry in an ndjson suite through `runtime.MatchRequest`, reports pass/fail, no gateway/auth required.
- Added a "Download test suite" button on the dashboard's recent-actions view, backed by a new `GET /api/actions/export` endpoint that re-shapes the existing event ring into suite ndjson (each entry pre-filled with the verdict the gateway actually produced).
- Dropped everything the previous proposal needed for live-session dispatch: session keying (§3.A), whole-machine carve-out, response synthesis, ephemeral peers, new auth, TTL sweep.
- Six open questions remaining in §3 (suite scope per-endpoint vs global, protocol families for v1, HITL handling, redaction defaults on export, export endpoint auth, export window/filters).

Left the filename as `dry-run-design.md` to avoid PR churn — happy to rename to `test-design.md` if you'd prefer.

arnauorriols:
The actions file should not be test-specific. Don't use keys like "expected" or "assert". Also, it should include the rule name that matched the action. The human approver should only result in "approve" action

arnauorriols:
[bot-reply]
Pushed faee093 addressing this comment:
- Reframed the file as an "actions file" (recorded gateway actions), dropping test-specific framing — renamed `--suite` → `--actions`, `SuiteEntry` → `Action`, "Download test suite" → "Download actions", and removed `expected_*` / `assert_*` keys from the format. The CLI is the test runner; the file is recorded reality.
- Each entry's `verdict` now carries the matched rule name as `verdict.rule` (the `CompiledRule.Name` from `config/compile.go:165`). Added a §1.3 note + a §4 scope item to extend `Event` with `Rule string` at the existing dispatch sites, since the recorder currently drops it.
- Locked in §3.C: the human approver path produces a terminal `verdict.action = "approve"` in both the recording and the runner — the chain is never invoked under `clawpatrol test`. Marked the open question as resolved.

doc/dry-run-design.md
```diff
@@ -0,0 +1,302 @@
...
+> questions in §3 must be resolved before code lands.
+
+## Goal
+
+Add a CLI subcommand that runs a candidate HCL policy against a
+**recorded actions file** and reports any verdicts the new policy
+would change:
+
+```
+clawpatrol test --config ./candidate.hcl --actions ./actions.json
```
ry: These actions should be in individual json files, probably in a `testdata/` directory or `fixtures/` directory. You can either run one at a time:
```
clawpatrol test candidate.hcl testdata/action1.json
```
or you can run against the entire directory:
```
clawpatrol test candidate.hcl testdata/
```
The exit code is non-zero if the verdict (eg approve, reject) that the candidate.hcl produces does not match the decision stored in `testdata/action1.json` or if running a whole directory, if any of the verdict don't match.

doc/dry-run-design.md
```diff
@@ -0,0 +1,302 @@
...
+shape the gateway logs for live traffic, just persisted to a
+file.
+
+`clawpatrol test` compiles the candidate config in-process, runs
+each request through `runtime.MatchRequest`, and reports any
+entry whose new verdict differs from the recorded one. Exit
+status is non-zero on any diff.
+
+To keep the authoring burden low, the dashboard grows a
+**"Download actions"** button on its recent-actions view. The
```
ry: The "download action" button is on the page for an individual action.

arnauorriols:
[bot-reply]
Pushed aea0eb2 addressing this comment:
- Reworked the design around one JSON file per action under `testdata/` (or `fixtures/`), not a single ndjson blob.
- CLI is now two positional args: `clawpatrol test <config> <file-or-dir>`. Directory mode globs `*.json` at the top level; single-file mode runs the one action.
- Non-zero exit if any verdict (action + rule) differs from the file; directory mode fails on any single mismatch.
- Dashboard "Download actions" button now emits a zip of per-action JSON files so operators can unpack into `testdata/` directly. `--update` rewrites matched files in place.

doc/dry-run-design.md
arnauorriols: [bot-reply]
Pushed 6278e2b addressing this comment:
- Moved the "Download action" button to the individual action detail page (`RequestDetailPage`); the recent-actions list no longer carries it.
- Reworked the export endpoint to `GET /api/actions/export?id=<event-id>` returning a single per-action JSON file (matches the per-file fixture format from cl-phf).
- Updated the open question on F (Export window) to reflect single-action download in v1, with bulk/zip export deferred as a follow-up.

ry:
Superseded by #235 — same design, family-keyed fixture shape and endpoint pinned by name. Continuing review there.

doc/dry-run-design.md
arnauorriols: [bot-reply]
Pushed aea0eb2 addressing this comment:
- Switched the dry-run design to one JSON file per action under `testdata/`.
- CLI now takes two positional args (config + a file or directory) and exits non-zero if any verdict mismatches the recorded one.
- Dashboard export emits a zip of per-action files instead of an ndjson stream.

PR #204, opened 2026-05-11
matcher: lowercase want values at compile time for always-lowercase facets

arnauorriols:
## Summary

Several rule facets compare an operator-supplied want value against a runtime got value that the gateway has already lower-cased. An operator writing `verb = "SELECT"` for SQL or `resource = "Pods"` for k8s could silently miss because the matcher compared verbatim against the always-lower-cased got side.

This PR extends the existing compile-time case-folding mechanism (`lowercasedPaths` + `normalizeWantLiterals` in `config/match/cel.go`) to cover `sql.tables`, `sql.functions`, and `k8s.resource` — facets whose got side is always lower-cased — so the want literals in the rule source are lowered at compile time.

> **Note:** this PR was reworked on top of current `main` (`0b5a6d2`). The original glob-based `LowerGlobs` approach is obsolete — `main` now folds case via CEL AST rewriting (`normalizeWantLiterals`) at compile time, not via globs. The description below reflects the actual implementation on the branch.

## Per-facet classification

| Family | Facet | Got side normalised? | Want side fix |
|---|---|---|---|
| http_rule | `method` | not always-lower (raw `net/http`); compared case-insensitively at runtime | leave alone |
| http_rule | `path` / `query` / `body_*` | case-sensitive (RFC) | leave alone |
| http_rule | `headers` | name canonical-cased (case-insensitive lookup); values verbatim | leave alone |
| sql_rule | `verb` | always-lower (postgres/clickhouse parsers) | already in `lowercasedPaths` |
| sql_rule | `tables` | always-lower (parsers fold identifiers via `strings.ToLower`) | **add to `lowercasedPaths` (new)** |
| sql_rule | `functions` | always-lower (same parser paths) | **add to `lowercasedPaths` (new)** |
| sql_rule | `statement` | case-sensitive raw SQL | leave alone (`matches('(?i)...')` opt-in) |
| k8s_rule | `verb` | always-lower (derived from HTTP method into a fixed vocabulary) | already in `lowercasedPaths` |
| k8s_rule | `resource` | always-lower (k8s API server only routes lower-case resource segments) | **add to `lowercasedPaths` (new)** |
| k8s_rule | `namespace` / `name` | case-sensitive (k8s allows mixed case) | leave alone |
| k8s_rule | `params` | per-param case-sensitive | leave alone |

## Changes

- `config/plugins/facets/sql/sql.go`: add `sql.tables` and `sql.functions` to `lowercasedPaths`, with a comment documenting why they are always-lower.
- `config/plugins/facets/k8s/k8s.go`: add `k8s.resource` to `lowercasedPaths`, with the same rationale.
- `config/match/cel.go`: extend `normalizeWantLiterals` to recognise the additional AST shapes these list-typed facets are used with — `"X" in <path>` (literal-LHS membership) and the `sets.intersects` / `sets.contains` / `sets.equivalent` two-list predicates — in addition to the previously handled `<path> == "X"`, `<path> in [list]`, and member `startsWith`/`endsWith`/`contains`.

## Test plan

- [x] `go test ./config/...` — all tests pass.
- [x] SQL (`sql_match_test.go`): `verb` case-insensitive incl. mixed-case list; `tables`/`functions` case-insensitive via literal-LHS `in` and `sets.intersects` with mixed/upper-case wants; `statement` stays case-sensitive.
- [x] k8s (`k8s_match_test.go`): `verb` and `resource` case-insensitive (`==`, `in`, `startsWith`/`endsWith`, reversed equality, globs, negation); `name`/`namespace` stay case-sensitive.
- [x] HTTP (`https_test.go`): `path` stays case-sensitive — locks in that case-folding does not bleed into `http.path`.

### DIFF
--- a/config/match/cel.go
+++ b/config/match/cel.go
@@ -280,12 +280,16 @@ func referencesPath(root celast.Expr, paths []celPath) bool {
 //
 // Recognised shapes — for each, the literal is normalized:
 //
-//	<path> == "X"           // and the symmetric "X" == <path>
-//	<path> != "X"           // and the symmetric "X" != <path>
-//	<path> in ["X", "Y"]    // every string literal in the list
-//	<path>.startsWith("X")  // member-call args
+//	<path> == "X"                            // and the symmetric "X" == <path>
+//	<path> != "X"                            // and the symmetric "X" != <path>
+//	<path> in ["X", "Y"]                     // every string literal in the list
+//	"X" in <path>                            // path is a list-typed facet
+//	<path>.startsWith("X")                   // member-call args
 //	<path>.endsWith("X")
 //	<path>.contains("X")
+//	sets.intersects(<path>, ["X", "Y"])      // and the symmetric form
+//	sets.contains(<path>, ["X", "Y"])
+//	sets.equivalent(<path>, ["X", "Y"])
 //
 // matches() is deliberately not normalized — its argument is a regex,
 // where case classes like `[A-Z]` are meaningful and the operator can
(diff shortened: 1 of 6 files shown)


arnauorriols:
Sync with main, and check what changes in this PR still make sense

arnauorriols:
[bot-reply]
Synced with main and reworked the PR on top of the current architecture (`0b5a6d2`).

The original glob-based `LowerGlobs` approach is obsolete — main now does case-folding via CEL `lowercasedPaths` at compile time (`config/match/cel.go`). Preserved the PR's intent on the new architecture:

- Added `sql.tables`, `sql.functions`, `k8s.resource` to their facets' `lowercasedPaths`.
- Extended `normalizeWantLiterals` to handle the shapes those facets are typically used with: `"X" in <path>` and `sets.intersects/contains/equivalent` (the AST walker previously only handled `<path> == "X"`, `<path> in [list]`, and member `startsWith`/`endsWith`/`contains`).
- Tests cover SQL tables/functions (literal-LHS `in`, `sets.intersects` with mixed/upper-case wants), k8s.resource (`==`, `in`, `startsWith`/`endsWith`, reversed equality), the case-sensitive siblings (`sql.statement`, `k8s.name`/`namespace`), and HTTP path (locks in that case-folding does not bleed into `http.path`).
- `go test ./config/...` green.

Dropped from the original PR: the glob-based `LowerGlobs` helper in `config/match/match.go` and the README case-handling table — the helpers' module no longer hosts glob logic, and the per-facet contract now lives next to the `lowercasedPaths` declaration.

arnauorriols:
[bot-reply]
Updated the PR description to match the reworked implementation on `0b5a6d2`.

The original body still described the obsolete glob-based `LowerGlobs` approach (a `match.go` helper + a `config/README.md` case-handling table), neither of which exists on the branch. The PR now folds case via CEL AST rewriting (`normalizeWantLiterals` in `config/match/cel.go`), so the body was stale and misleading to reviewers.

The refreshed description covers what the branch actually does:
- `sql.go` / `k8s.go`: add `sql.tables`, `sql.functions`, `k8s.resource` to `lowercasedPaths` with rationale comments.
- `cel.go`: extend `normalizeWantLiterals` to also handle `"X" in <path>` and the `sets.intersects`/`sets.contains`/`sets.equivalent` shapes these list-typed facets are used with.
- Per-facet classification table and test plan updated to reference the real test files (`sql_match_test.go`, `k8s_match_test.go`, `https_test.go`).

No code change in this cycle — `go test ./config/...` green, CI lint + test already passing.

PR #209, opened 2026-05-11
gateway: prune stale HTTP transports on reload

magurotuna:
## Summary
- Prunes per-endpoint HTTP transports after a successful config reload so old `*CompiledEndpoint` pointer keys do not remain cached indefinitely.
- Treats in-flight handlers that still hold stale endpoint pointers as uncached transports and closes their idle connections on release, avoiding stale re-insertion after pruning.
- Adds regression coverage for old/removed endpoint pruning and stale post-reload `transportFor` calls.

## Test plan
- [x] `go test ./...`
- [x] `go test -race .`
- [x] `go vet ./...`
- [x] `git diff --check`

### DIFF
--- a/dial_test.go
+++ b/dial_test.go
@@ -57,7 +57,9 @@ func TestTransportBrowserTLSUsesEndpointTunnel(t *testing.T) {
 
 	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
 	defer cancel()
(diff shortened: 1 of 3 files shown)


piscisaureus:
Closing during the backlog pass (pre-launch branch, no activity since May). If per-endpoint HTTP transports still accumulate across reloads on current main, please re-file it as an issue with a reproduction; a fix on top of today's transportFor would be small and welcome.

Issue #48, opened 2026-05-04
Agents running webhooks cannot run with `clawpatrol run`


arnauorriols:
Reference implementation in ref-impl: https://github.com/example/ref-impl/blob/69dafcaa1363ea4055012717fff54376e9704de5/native/napi/src/client_linux/relay.rs

PR #232, opened 2026-05-12
run_linux: auto-expose webhook listeners on host via seccomp relay

divybot:
## Summary

- `clawpatrol run` on Linux currently puts the user's command in a fresh netns, so TCP listeners (webhook servers) it opens aren't reachable from the host.
- This adds the same auto-expose relay ref-impl uses ([reference](https://github.com/example/ref-impl/blob/69dafcaa1363ea4055012717fff54376e9704de5/native/napi/src/client_linux/relay.rs)): a seccomp `listen(2)` trap watched by a supervisor in the host netns, plus a worker spliced into the agent netns. Each port the agent binds gets a mirrored host-side listener, so inbound webhooks land.

## Architecture

```
top parent (host netns, drives WG)
  ├── relay-supervisor (host netns)
  │     ioctl(notify_fd, NOTIF_RECV) → pidfd_getfd → host listen → accept → SCM_RIGHTS to worker
  └── netns child
        ├── installs seccomp filter (listen → USER_NOTIF)
        ├── socketpair; ships {notify_fd, sup_sock} to top parent
        ├── fork+execs relay-worker (inherits userns+netns)
        │     recvmsg → connect 127.0.0.1:port → bidi copy
        └── execs user cmd
```

- Two new internal subcommands (`relay-supervisor`, `relay-worker`) re-exec the same binary, dispatched in `main.go`. Hidden from usage.
- The child sets `PR_SET_PTRACER_ANY` before exec so the supervisor (same-uid sibling) can `pidfd_getfd` listen sockets through yama with `ptrace_scope=1`.
- Failures (unsupported arch, ENOENT on a child process that didn't opt into ptrace, host port already bound) are non-fatal: log a warning and continue without auto-expose for that port.

Closes example/orchestrator#7 (mirrors #48 here).

## Test plan

- [ ] `go build -tags ts_omit_identityfederation ./...` on linux amd64 + arm64 + darwin
- [ ] `go vet -tags ts_omit_identityfederation ./...`
- [ ] `go test ./...` (existing suites pass)
- [ ] Manual: `clawpatrol run -- python3 -m http.server 9000` on a joined linux box → `curl http://<host>:9000/` from another machine returns the directory listing.
- [ ] Manual: confirm `clear_ambient_caps` still fires (user cmd has no CAP_NET_ADMIN) and that the supervisor/worker terminate when the user cmd exits (Pdeathsig).


arnauorriols (commented):
[bot-reply]

## Comparison vs unclaw `relay.rs`

Comparing PR #232 against the reference cited in the PR description: `native/napi/src/client_linux/relay.rs` in denoland/unclaw (revision 69dafcaa). Citations below use `relay_linux.go:N` for the PR and `unclaw:relay.rs:N` / `unclaw:seccomp.rs:N` / `unclaw:netns.rs:N` for the reference. I read both files end-to-end.

### 1. Seccomp filter (listen-trap BPF)

- Identical 7-instruction filter (arch gate → `nr == SYS_listen` → `USER_NOTIF`, else `ALLOW`). `relay_linux.go:130-137` mirrors `unclaw:seccomp.rs:85-93`.
- Identical setup: `PR_SET_NO_NEW_PRIVS` + `seccomp(SET_MODE_FILTER, NEW_LISTENER)` returning the notify fd. `relay_linux.go:116-151` ↔ `unclaw:seccomp.rs:103-117`.
- Arch dispatch limited to x86_64 + aarch64; unsupported arches fall back to running without auto-expose. `relay_linux.go:485-493` ↔ `unclaw:relay.rs:122-125`.

**Mirrors unclaw cleanly.**

### 2. Process topology

This is the most interesting divergence.

**unclaw** (`unclaw:relay.rs:43-116`, `unclaw:netns.rs:448-466`):
- Top parent calls `relay::spawn(notify_fd, userns_fd, netns_fd)` → forks supervisor in host netns.
- Supervisor then forks the worker; the worker calls `setns(userns_fd, CLONE_NEWUSER)` then `setns(netns_fd, CLONE_NEWNET)` (`unclaw:relay.rs:302-307`).
- Requires single-threaded post-fork state to legally `setns(NEWUSER)`.

**PR** (`relay_linux.go:505-537`, `relay_linux.go:571-588`, `run_linux.go` recvFDs site):
- Agent child forks worker via `exec.Command` (`setupRelayInChild` → `spawnRelayWorker`); worker inherits agent userns+netns naturally, no `setns`.
- Top parent re-execs the binary as the hidden `relay-supervisor` subcommand; supervisor stays in host netns.
- SCM_RIGHTS payload shrinks: PR ships **2 fds** `(notify_fd, sup_sock)`; unclaw ships **3 fds** `(notify_fd, userns_fd, netns_fd)`.

**Assessment**: defensible divergence and arguably the cleaner shape. Go's runtime is multi-threaded by default and `setns(NEWUSER)` requires a single-threaded process — doing the unclaw dance from Go is a known hazard. Forking the worker from inside the agent child (where the namespaces are already correct) sidesteps it. The re-exec-with-hidden-subcommand pattern is idiomatic Go. Fd handoff via `cmd.ExtraFiles` is appropriate.

### 3. Cross-namespace fd inspection (`pidfd_getfd` + yama)

Both implementations use `pidfd_open` + `pidfd_getfd` + `getsockname` to read the bound port/addr from the agent's listen socket without entering its netns. `relay_linux.go:299-323` ↔ `unclaw:relay.rs:243-281`.

**PR adds something unclaw lacks**: `PR_SET_PTRACER_ANY` on the agent child (`run_linux.go` in the new code, with `ptraceAny = ^uintptr(0)` constant).

Under yama `ptrace_scope=1` (common distro default), `pidfd_getfd`'s `PTRACE_MODE_ATTACH_REALCREDS` check rejects sibling-process inspection. The supervisor and the agent are siblings of the top parent in both implementations. unclaw appears to rely on permissive yama; the PR explicitly opts in via `PR_SET_PTRACER_ANY` before exec.

**Caveat**: `PR_SET_PTRACER` persists across `execve`, but inheritance into `fork()` children is fuzzy across kernel versions (yama maintains a per-task relationship list, not always inherited). The "single binary that listens" case works; if the user's command is a wrapper script that `exec`s a child listener, behavior may degrade silently (yama rejects, peekAgentListener returns the warning at `relay_linux.go:308`, listen() still succeeds because CONTINUE was sent first, but auto-expose is skipped for that port).

Worth documenting this in the PR description as a known limitation, or considering a periodic `PR_SET_PTRACER_ANY` push (not trivial — there's no hook for subprocess creation).

### 4. Bind-scope mirroring

`relay_linux.go:327-341` and `unclaw:relay.rs:220-239`: identical policy. Loopback inner → loopback host; otherwise unspecified host bind; fallback to v4 loopback on family mismatch. Mirror.

### 5. Accept-loop + worker bidi copy

`relay_linux.go:343-368, 448-478` mirrors `unclaw:relay.rs:197-218, 366-402`:
- SOCK_SEQPACKET socketpair for `(port, fd)` framing.
- Accept on host → `sendmsg` with `SCM_RIGHTS` → worker `connect("127.0.0.1", port)` → two-goroutine/thread bidi copy with half-close on EOF.
- PR closes the accepted `*net.TCPConn` immediately after `sendmsg` (`relay_linux.go:362`); unclaw drops the `TcpStream` after `sendmsg` (`unclaw:relay.rs:216`). Equivalent — SCM_RIGHTS dup means the worker now owns the fd.

Mirror.

### 6. Lifecycle / cleanup

**unclaw** (`unclaw:netns.rs:472-477`): top parent explicitly `kill(supervisor_pid, SIGTERM)` then `waitpid(supervisor_pid)` after the user cmd exits.

**PR** (no explicit kill in the diff): relies on `Pdeathsig=SIGTERM` (`relay_linux.go:556, 582`) plus the kernel's "filter group empty → ENOENT from `notif_recv`" path (`relay_linux.go:248`).

Both converge to clean shutdown on the happy path:
- Supervisor sees ENOENT once the user cmd's process group exits → returns.
- Worker's parent (the agent child running the user cmd) exits → Pdeathsig fires.
- If the supervisor outlives the top parent, its Pdeathsig fires when top parent exits.

**Minor concern**: PR top parent never `Wait`s on the supervisor process; after the top parent calls `os.Exit(ee.ExitCode())` (`run_linux.go` runRun exit path), the supervisor is briefly inherited by `init`/subreaper for reaping. Functionally fine on systems with a sane init, but a small watcher anomaly versus unclaw's explicit `waitpid`. A `defer` that signals + waits would tighten this up.

### 7. Backwards compatibility / opt-out

unclaw threads an `auto_expose: bool` through `child_run` (`unclaw:netns.rs:448`), letting callers turn the relay off.

The PR makes auto-expose unconditional for `clawpatrol run` on Linux — no flag in the diff. On unsupported arch (`errSeccompUnsupportedArch`) it degrades gracefully. There's no escape hatch for users who don't want their `listen()` syscalls intercepted (e.g., conflict with another seccomp filter, or simply not wanting host-side ports auto-opened).

**Severity**: minor / open question — is unconditional the intended UX, or should there be a `--no-auto-expose` / env var?

### 8. Platform coverage

`relay_linux.go` is build-tagged `//go:build linux`; `relay_other.go` provides stubs that `fail("relay-supervisor is linux-only")` and documents the kernel features the relay depends on. macOS uses NE which does not isolate the wrapped process in a netns, so the analogous problem doesn't exist — scope is correctly Linux-only. Mirror of unclaw's `#![cfg(target_os = "linux")]` gate.

### 9. Misc code-quality observations

- `ignoreSIGPIPE` (`relay_linux.go:595-608`) reinvents `os/signal.Ignore` with a hand-rolled `struct sigaction` and raw `rt_sigaction`. The struct layout assumes `sa_restorer` between `sa_flags` and `sa_mask` — correct on x86_64, but on aarch64 the kernel's `struct sigaction` has no `sa_restorer`, so the `Restore` field ends up where `sa_mask` should be. For `SIG_IGN` (`Handler = 1`, `Flags = 0`) this happens to work because the kernel doesn't dereference sa_restorer (no `SA_RESTORER`) and sa_mask is irrelevant for SIG_IGN, but it's fragile. **Recommend** replacing with `signal.Ignore(syscall.SIGPIPE)` in both `runRelaySupervisor` and `runRelayWorker`.
- `notifSendContinue` always responds CONTINUE even for non-listen syscalls (`relay_linux.go:287`). Correct — the seccomp filter only traps `listen`, but defensive against future arch oddities. Matches `unclaw:relay.rs:154-156`.
- `peekAgentListener` error string includes the yama hint inline. Useful for debugging. Nice touch over unclaw, which just reports `pidfd_getfd` errno.

## Findings summary

- ✅ **Mirrors unclaw cleanly**: seccomp filter and arch gate, `pidfd_open`+`pidfd_getfd`+`getsockname` socket inspection, bind-scope policy, accept→sendmsg→connect→bidi forwarding, SOCK_SEQPACKET framing, SIGPIPE ignored in supervisor/worker, Linux-only platform scope.

- ⚠ **Defensible divergences**:
  - Worker forked from agent child (Go) instead of from supervisor + `setns(NEWUSER/NEWNET)` (Rust). Avoids the Go-multi-threaded-setns hazard. SCM_RIGHTS payload shrinks from 3 fds to 2.
  - Re-exec hidden subcommand for both supervisor and worker, vs in-process fork. Slightly more startup cost but idiomatic Go and cleaner runtime separation.
  - Cleanup via `Pdeathsig` + ENOENT cascade, vs explicit `SIGTERM` + `waitpid`. Functionally equivalent on healthy systems.

- ❌ **Likely bugs / gaps**:
  - **MINOR**: `PR_SET_PTRACER_ANY` is set on the agent child but its inheritance into subprocesses of the user cmd is kernel-dependent. Wrapper-script-spawned listeners may log the yama warning and silently miss auto-expose. Worth documenting as a known limitation or testing under `ptrace_scope=1` with a subprocess test case.
  - **MINOR**: No opt-out flag — `clawpatrol run` on Linux always installs the listen trap. Consider mirroring unclaw's `auto_expose: bool` as `--no-auto-expose` / env var.
  - **MINOR**: Top parent does not explicitly wait on the supervisor process; relies on `Pdeathsig` + subreaper. Tightening with a `defer` that signals + waits would match unclaw's explicit teardown.
  - **MINOR**: `ignoreSIGPIPE` hand-rolls `rt_sigaction` with an arch-dependent struct layout that happens to work for `SIG_IGN` but is fragile. Replace with `signal.Ignore(syscall.SIGPIPE)`.

- 💡 **Improvements over unclaw**:
  - **`PR_SET_PTRACER_ANY` in the agent child**. unclaw lacks this and so relies on permissive yama (`ptrace_scope=0`). The PR works under `ptrace_scope=1` for the agent's own `listen()` calls. Worth upstreaming back to unclaw.
  - **Cleaner namespace inheritance**: by forking the worker from inside the netns child, the PR avoids `setns(NEWUSER)` from a multi-threaded runtime entirely. A future Rust port could adopt the same shape and drop the single-thread-post-fork constraint in `unclaw::relay::spawn`.
  - **Stub `relay_other.go`** keeps cross-platform builds clean without per-call-site `#[cfg]` clutter.

Issue #143, opened 2026-05-08
Postgres SQL rule parser: audit of false-positive / false-negative bypasses


arnauorriols:
A focused audit of the postgres endpoint's SQL parser used by `sql_rule`,
listing concrete cases where the matcher fires when it shouldn't
(false-positive) or fails to fire when it should (false-negative).
Each finding cites the file:line evidence and pairs with a payload
that demonstrates the issue. **No solutions are proposed.**

Files cross-referenced:

- `config/plugins/endpoints/postgres.go` — the postgres wire-protocol
  gateway and the `parseSQL` extractor.
- `config/match/match.go` — the `sqlMatcher` that consumes
  `SQLMeta{Verb, Tables, Functions, Statement}`.

The matcher's input shape is fully determined by `parseSQL`
(`postgres.go:628-647`), which:

1. `strings.TrimSpace(sql)` (postgres.go:629).
2. Lowercases the entire string (postgres.go:634).
3. Takes the prefix up to the first ` `/`\t`/`\n`/`\r`/`(` as `Verb`
   (postgres.go:635-639).
4. Runs two regexes over the lowercased string to populate
   `Tables` and `Functions` (postgres.go:640-645):
   - `pgTableRE = (?i)\b(?:from|update|into|join)\s+([a-z_][a-z0-9_.]*)`
     (postgres.go:619).
   - `pgFuncRE  = (?i)\b([a-z_][a-z0-9_]*)\s*\(` (postgres.go:620).
5. Stores the trimmed (case-preserved) original as `Statement`
   (postgres.go:630).

The matcher then case-folds `verb` (`match.go:382-393`,
`equalsIgnoreCase`), glob-matches `tables`/`function` per-element
with list-OR semantics (`match.go:394-403`, `anyOfStrings` →
`matchAny`), and matches `statement`/`statement_regex` against the
case-preserved raw SQL (`match.go:404-409`).

Throughout this issue: **FP** = false positive (matcher fires when
it shouldn't), **FN** = false negative (matcher fails to fire when
it should). A FN that lets a denied verb pass is a **policy
evasion**; an FP that denies an allowed query is a **usability**
bug. Both are flagged.

---

### 1.1 Trailing-semicolon / no-whitespace verbs evade `verb` matchers (FN, evasion)

`parseSQL` cuts the verb at the first character in
`" \t\n\r("` (`postgres.go:635`). Any other character —
notably `;`, `*`, comment-start `/`, or `--` — is folded into the
[...]

PR #248, opened 2026-05-12
postgres: audit & fix regex SQL parser bypasses

littledivy:
Closes example/orchestrator#20 — audit of the Postgres SQL rule parser
from clawpatrol#143.

## Summary

The legacy `parseSQL` ran two regexes over the lowercased statement
and sliced the first whitespace-delimited token off the front for
the verb. The matcher fed `sql.verb / sql.tables / sql.function` into
operator rules. Because none of those passes understood strings,
dollar-quoted bodies, comments, multi-statement `;` boundaries, or
CTE wrappers, a handful of one-line tricks slipped past verb-based
rules.

This PR ships a hand-rolled tokenizer + token-stream walker that's
bypass-resistant by construction, and adds a `sql.verbs` (plural)
CEL list so rules can express "any statement in this batch is a
write". The singular `sql.verb` keeps reporting the first top-level
statement's outer verb so existing rules don't regress.

## Bypass catalogue

Each entry is the rule-bypass the legacy extractor enabled and how
this PR closes it. Severity reflects worst plausible outcome.

### High

1. **Multi-statement verb hiding** — `SELECT 1; DROP TABLE users`.
   Pre-fix: `verb == "select"`, tables empty (the regex anchored on
   `FROM`/`UPDATE`/`INTO`/`JOIN`, never `TABLE`), so both
   `verb == 'drop'` rules and `'users' in sql.tables` rules missed.
   Post-fix: `sql.verbs == ["select", "drop"]`, `sql.tables == ["users"]`.
   Recommended rule shape: `sql.verbs.exists(v, v in ['drop', …])`.
   The `full.hcl` golden now adopts this idiom.

2. **CTE-wrapped write** —
   `WITH x AS (DELETE FROM secrets RETURNING *) SELECT * FROM x`.
   Pre-fix: `verb == "with"`, tables `[secrets, x]`. Operators
   gating on `verb == 'delete'` missed entirely. Post-fix: the CTE
   walker unwraps the leading `WITH`, surfaces `verb == "select"`
   (outer) plus `sql.verbs == ["select", "delete"]`, and filters
   the CTE binding `x` out of tables so `sql.tables == ["secrets"]`.

3. **Leading comment masking the verb** —
   `/* sneaky */ DROP TABLE users` or `-- sneaky\nDROP TABLE users`.
   Pre-fix: `verb == "/*"` / `"--"` (first whitespace-delimited
   token), so no verb-based rule fired. Post-fix: the lexer drops
   comments before extraction; `verb == "drop"`.

4. **DDL on bare relations** — `DROP TABLE users`, `TRUNCATE TABLE
   sessions`. Pre-fix: the regex only fired on
   `FROM/UPDATE/INTO/JOIN`, so `users` / `sessions` never reached
   the table list. Post-fix: `TABLE` is now a recognised table
   introducer after DROP/TRUNCATE/ALTER/CREATE.

### Medium

5. **String-literal pollution** — `SELECT 'FROM secrets'`. Pre-fix:
   the regex matched `FROM secrets` inside the literal and reported
   `secrets` as a table (false positive). Post-fix: string bodies
   are opaque.

6. **Dollar-quoted body pollution** — `SELECT \$\$;DROP TABLE x;\$\$`
   or `DO \$\$ … \$\$`. Pre-fix: identifier-shaped substrings inside
   `\$\$…\$\$` leaked into both tables and functions. Post-fix:
   dollar-quoted strings are opaque, with full `\$tag\$ … \$tag\$`
   matching so tags containing `;` don't confuse statement splitting.

7. **Comment-broken function call** —
   `SELECT pg_terminate_backend/* hi */(1)`. Pre-fix: the regex
   `\b<ident>\s*\(` doesn't include `/*…*/` in `\s`, so the function
   wasn't captured. Post-fix: comments are dropped before tokenizing.

8. **Comma-join misses tables** — `SELECT * FROM a, b, c`. Pre-fix:
   only `a` extracted (`FROM` anchored on the first ident).
   Post-fix: FROM list walker continues over `,`.

9. **Quoted identifier missed** — `SELECT * FROM "Sensitive Table"`.
   Pre-fix: regex anchored on `[a-z_]`, so `"Sensitive Table"` was
   ignored. Post-fix: quoted identifiers are tokenized and recorded
   (lowercased) — `tables == ["sensitive table"]`.

10. **LATERAL / NATURAL / CROSS JOIN variants** —
    `SELECT * FROM a, LATERAL (SELECT * FROM b)`. Pre-fix: relied
    on regex matching `JOIN` and missed `LATERAL` + subquery.
    Post-fix: FROM walker recurses into subqueries (so `b` surfaces)
    and recognises LATERAL/NATURAL/CROSS as JOIN modifiers.

### Low (intentional / accepted)

11. **Function extraction is overgreedy** — every `<ident>(` is
    captured, including `audit (` from `INSERT INTO audit (…)` and
    `as (` from CTEs. Documented and tested as a deliberate
    trade-off: the matcher consumes a list and rule writers query
    specific names, so noise is harmless. The new lexer at least
    guarantees the noise can't come from inside a string/comment.

12. **Case-insensitivity** — `sElEcT * fRoM Users` already worked
    (regex was `(?i)` and the verb was lower-cased) but a regression
    test now pins it.

13. **LISTEN / NOTIFY** — verb-only; no tables. Already correct.

14. **PL/pgSQL bodies inside DO \$\$…\$\$** — intentionally not
    parsed. The tokenizer treats dollar-quoted bodies as opaque
    strings. Operators who want to block PL/pgSQL execution should
    rule on `'do' in sql.verbs`. The docs in `ai.go` were updated
    to advertise this.

15. **COPY ... FROM PROGRAM 'curl …'** — the FROM-list walker still
    records `program` as a table because it doesn't special-case
    the PROGRAM keyword. Harmless because operators rely on the
    `statement` glob `"*COPY*FROM PROGRAM*"` which matches the raw
    text directly. Recorded in the regression test.

## API

- `sqlfacet.Meta` gains a `Verbs []string` slice — every top-level
  statement's outer verb plus every CTE-body verb. Postgres and
  ClickHouse runtimes populate it.
- `SqlFields` gains a CEL field `verbs` (list<string>, lower-cased)
  via the existing struct-tag CEL binding.
- `Facet.ReportFields()` adds `verbs` so the dashboard report
  schema stays in sync.
- The AI prompt in `ai.go` documents `sql.verbs` and ships a sample
  rule (`'drop' in sql.verbs || 'truncate' in sql.verbs`).
- `config/testdata/full.hcl`'s `pg-banned-verbs` rule swaps
  `sql.verb in […]` for `sql.verbs.exists(v, v in […])` to pick up
  the multi-statement / CTE-wrapped cases. The
  `full.want.json` golden regenerates accordingly.

## Test plan

- [x] `go test ./...` passes (added 25 new regression cases on top
      of the existing 9; the table/function expectations of three
      legacy cases shift to reflect more-correct extraction —
      documented in the test comments).
- [x] `go vet ./...` clean.
- [x] `go build ./config/...` clean (CGo-free; deploys with
      `CGO_ENABLED=0`).
- [x] New tests verify each bypass row in the catalogue above.
- [x] New `sql.verbs` CEL matcher test confirms the field is
      exposed and that `'drop' in sql.verbs` fires.


littledivy:
@ops-bot resolve conflicts

littledivy:
Superseded by #282 which shipped a stronger fix (pgplex/pgparser-based AST walker + per-statement Inner-walk evaluation) for the same audit (#143). I've merged main into this branch and reverted all my SQL-extractor changes — the diff vs main is now empty. Feel free to close.

PR #250, opened 2026-05-12
feat: 1Password CLI credential type

littledivy:
## Summary

- Adds a `1password` credential type that resolves `op://Vault/Item/field` references via the `op` CLI at request time, caching results per-credential for a configurable TTL (default 60s).
- New `runtime.SecretSourceProvider` interface lets a credential body source its own bytes; `gatewaySecretStore` consults it between the OAuth registry and the env-var fallback.
- HTTPS dispatcher fails closed for self-sourcing credentials: an op-fetch or injection error returns 502 instead of forwarding an un-credentialed request.

## Example

```hcl
credential "1password" "openai-prod" {
  ref = "op://Engineering/OpenAI/api_key"
  ttl = "60s"          # optional; default 60s
}

endpoint "https" "openai" {
  hosts      = ["api.openai.com"]
  credential = openai-prod
}
```

The gateway host needs `op` on `$PATH` and a live signin (interactive `op signin`, or `OP_SERVICE_ACCOUNT_TOKEN` for headless boxes).

## Notes

- Dashboard-paste continues to win over 1Password — useful when debugging or running without a live `op` session.
- The injection shape defaults to `Authorization: Bearer <value>`. `header` + `prefix` overrides cover non-bearer providers (`X-API-Key: <value>`, `Authorization: Token …`, etc.); setting `prefix = ""` on a custom `header` stamps the value verbatim.
- `gateway.example.hcl`, the onboarding doc, and the generated `config-reference.md` all describe the new type and the `op signin` prerequisite.

## Test plan

- [x] `go test ./...` — unit tests cover bearer + custom-header injection, cache hit/miss, TTL expiry, op-error propagation, empty-ref + empty-secret refusals, and the runtime interface assertions.
- [x] `go run ./tools/docgen --check` — generated config reference matches the registered plugin schema.
- [x] `feature_example.hcl` golden updated to include a 1password credential so the load+dump path is exercised.

Closes example/orchestrator#19


littledivy:
Distraction

PR #254, opened 2026-05-12
feat: structured plugin diagnostic log tab in dashboard

littledivy:
Adds a `/#/logs` tab to the dashboard that tails structured plugin diagnostic events, with filter controls for plugin, severity, time range, and agent.

## Changes

- `config/runtime/logger.go`: new `runtime.Logger` interface
- Ring-buffered `LogSink` (default 1000 entries, configurable via `plugin_log_buffer_size` HCL field)
- `GET /api/logs` — filtered snapshot; `GET /api/logs/stream` — SSE live tail
- HTTP credential injection, synthetic responders, approver notify, and postgres deny paths now route through the logger
- Two-layer credential redaction: auth/token/secret/key/password/cookie keys and `[]byte` values become `***` before reaching the buffer or stderr mirror
- 13 new tests, all passing under `-race`

Closes example/orchestrator#21

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,6 +53,12 @@ type Gateway struct {
 	// and DO_NOT_TRACK=1 also work.
 	Telemetry *bool `hcl:"telemetry,optional"`
 
+	// PluginLogBufferSize caps the dashboard /logs tab's in-memory
+	// ring buffer. Default 1000 entries; raise on busy gateways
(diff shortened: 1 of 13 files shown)


littledivy:
@ops-bot rebase. also i wanna see how to looks like with sample data. send a screenie

littledivy:
Stale.

Issue #124, opened 2026-05-07
Dispatch architecture: direct-IP (postgres) vs VIP (clickhouse_native)


arnauorriols:
## Summary

Two plugins solve the same general problem — route an inbound raw‑TCP
connection from the WG promiscuous forwarder to the right endpoint
plugin — using two different dispatch models.

- **postgres** ([config/plugins/endpoints/postgres.go](../blob/main/config/plugins/endpoints/postgres.go),
  [main.go `handlePostgresConn`](../blob/main/main.go))
  rides a **direct‑IP** path: a `ConnIndex` of `dstIP → endpoint` is
  built at policy load by DNS‑resolving each `endpoint.host`; a
  hardcoded `:5432` arm in the WG forwarder hands every TCP segment to
  the postgres handler, which consults the index by destination IP.

- **clickhouse_native** ([config/plugins/endpoints/clickhouse_native.go](../blob/main/config/plugins/endpoints/clickhouse_native.go),
  [config/plugins/endpoints/clickhouse_native_runtime.go](../blob/main/config/plugins/endpoints/clickhouse_native_runtime.go))
  rides a **VIP** path: it declares `RequiresVIP() bool { return true }`,
  the [`dnsvip`](../blob/main/dnsvip/dnsvip.go) allocator assigns a
  stable VIP per hostname, the agent's DNS query for that hostname is
  intercepted and answered with the VIP, and the WG forwarder routes
  any TCP segment whose dst falls in the VIP range through
  `handleVIPConn → dispatchConnEndpoint`. As of `2f629d1` it also
  participates in the direct‑IP path (`tryDirectIPConn`) for
  IP‑literal hosts, which `dnsvip.collectRequiredHosts` skips.

Both plugins implement the same `runtime.ConnRouter` interface
(`ConnRouteHosts() []string`, [config/runtime/conn_route.go](../blob/main/config/runtime/conn_route.go)).
Postgres only uses it for the direct‑IP index; clickhouse_native uses
it for both the VIP table source‑of‑truth and the direct‑IP fallback.

## postgres — direct‑IP dispatch

**Flow.** agent dials `db.example.com:5432` → public DNS resolves to
the real RDS IP → WG forwarder accepts on `:5432` →
`tcpDispatch` arm `case dstPort == 5432` →
`g.handlePostgresConn(c, dstIP)` → `connIdx.Lookup(dstIP)` →
`pickEndpointForProfile` → `postgres.HandleConn`. When the index
[...]

PR #255, opened 2026-05-12
feat: ClickHouse native protocol — ADR + port-based dispatch fallback

littledivy:
Implements the dispatch architecture for ClickHouse native protocol as described in denoland/clawpatrol#124.

Adds an ADR documenting the two-phase approach and a port-based dispatch fallback for the ClickHouse native binary protocol, enabling interception without requiring VIP-only dispatch.

Closes example/orchestrator#25


littledivy:
Stale.

PR #258, opened 2026-05-12
server: add OIDC ephemeral lease store

magurotuna:
## Summary

- Adds SQLite schema for OIDC replay reservations and OIDC-backed ephemeral peer leases.
- Stores derived replay identifiers and token hashes only; raw OIDC tokens are never persisted.
- Adds helpers for atomic replay reservation, lease creation, active lease lookup, and revocation.
- Caps lease expiration to the verified token expiry carried by the replay reservation.

Stack: this PR is stacked on #257 (`agent/oidc-verifier`).

## Testing

- `(cd www && npm ci && npm run build)`
- `go test . -run 'TestReserveOIDCReplay|TestCreateOIDCEphemeralLease|TestRevokeOIDC' -v`
- `go test ./config/... ./internal/oidcverify`
- `go test ./...`
- `git diff --check`
- `go run ./tools/docgen --check`

PR #259, opened 2026-05-12
server: add OIDC onboarding endpoint

magurotuna:
## Summary

- Adds public, self-authenticating `POST /api/onboard/oidc` for unattended OIDC ephemeral enrollment.
- Verifies the OIDC ID token, matches it against compiled enrollment policy, reserves replay once, provisions a WireGuard peer, creates an ephemeral lease, and returns the peer connection material needed by the CLI.
- Returns only derived/sanitized token data server-side; raw OIDC tokens are not persisted.
- Keeps the route public at the dashboard/tailnet gates because the OIDC token is the request proof for disposable compute before it has tunnel identity.

Stack: this PR is stacked on #258 (`agent/oidc-ephemeral-leases`).

## Testing

- `(cd www && npm ci && npm run build)`
- `go test . -run 'TestRouteAuthRequirementsDocumentOnboardingBoundary'`
- `go test ./config/... ./internal/oidcverify`
- `go test ./...`
- `git diff --check`
- `go run ./tools/docgen --check`

### DIFF
--- a/oidc_onboard.go
+++ b/oidc_onboard.go
@@ -1,0 +1,168 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/netip"
+	"strings"
+	"time"
+
+	configruntime "github.com/denoland/clawpatrol/config/runtime"
+)
+
+const maxOIDCOnboardBodyBytes = 1 << 20
+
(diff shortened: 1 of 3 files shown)


piscisaureus:
Closing during a backlog triage pass: superseded by #753, which carries the whole OIDC enrollment slice as one PR and is the one we will review together with the Kubernetes enrollment work in #792.

PR #263, opened 2026-05-12
fix: clawpatrol run on Linux creates new WireGuard device per invocation instead of reusing

littledivy:
Closes example/orchestrator#37


littledivy:
Closing — the caching approach conflicts with the intended design. Each `clawpatrol run` should get a fresh keypair per invocation so concurrent runs have independent WG sessions and don't stomp each other. PRs #224, #237, and #262 already fix the dashboard leak without touching client-side keypair generation: ephemeral peers are transparent (no device rows) and their traffic is attributed to the parent device.

PR #267, opened 2026-05-12
security: deliver CA certificate over WireGuard tunnel, not plain HTTP

littledivy:
Closes example/orchestrator#8


littledivy:
@ops-bot rebase. I also don't like that there is a new flag now and device code is replaced with CA fingerprint.

littledivy:
@ops-bot rebase

PR #272, opened 2026-05-12
docs: refresh public documentation

magurotuna:
## Summary
- refresh public docs for current CLI, onboarding, gateway, plugin, and LLM session/token tracking behavior
- generate tunnel block coverage in the HCL config reference
- add tunnel plugin Go comments and include tunnel sources in docgen so generated field descriptions are populated

## Test Plan
- `go generate ./config/plugins/all`
- `go test ./tools/docgen ./tools/docgen/internal/render ./config/plugins/tunnels`
- `npm --prefix site run build`
- `git diff --check`

### DIFF
--- a/config/plugins/tunnels/kubernetes_port_forward.go
+++ b/config/plugins/tunnels/kubernetes_port_forward.go
@@ -77,30 +77,37 @@ import (
 
 // KubernetesPortForwardTunnel configures the tunnel runtime.
 type KubernetesPortForwardTunnel struct {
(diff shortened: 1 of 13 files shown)


magurotuna:
Closing this as stale after the docs rewrite landed on `main` via #285/#287/#293.

The broad public-doc refresh here would now resurrect pages and structure that `main` intentionally removed/replaced. I carried forward the still-useful subset (generated tunnel config-reference field docs plus stale landing CTA cleanup) in the smaller replacement PR #301.

PR #282, opened 2026-05-12
postgres: tokenizer-based SQL extractor closes parser evasions (#143)

arnauorriols:
## Summary

Replaces the regex-based SQL extractor in the postgres endpoint with
a hand-rolled tokenizer + per-section extractor. Closes the
(FN, evasion) findings from #143 — verb / table / wire-protocol
parser evasions where an agent could route a denied verb or
forbidden table past `sql_rule` policy.

## Parsing approach

**Hand-rolled postgres SQL tokenizer**, not vitess sqlparser /
cockroach parser. The matcher input is small (verb, tables list,
functions list, statement text); a tokenizer is ~500 LOC versus a
megabyte parser dependency. Postgres-flavoured first: dollar-quoted
strings, E-strings, nested block comments, quoted identifiers. No
regex backtracking; linear in input length.

The tokenizer tracks byte offsets so per-statement raw text
survives splitting — multi-statement Q payloads land in the
matcher as N invocations with N raw `statement` strings, not one
joined blob.

## Findings closed

Each finding maps to: extractor change, regression test that fires
the audit's payload, code/test reference to the section.

### Verb extraction (§1)
- **§1.1** Trailing-semicolon / no-whitespace verbs — `BEGIN;`,
  `DROP;`, `SELECT*FROM x`, `DELETE/*c*/FROM x`, `SELECT;`.
- **§1.2** CTE-hidden DML — `WITH x AS (DELETE FROM users …) …`
  emits a shadow sub-statement so the inner DML reaches the matcher.
- **§1.3** Multi-statement Simple Query — each `;`-separated
  statement walks the matcher; deny on any sub-statement denies the
  whole wire query.
- **§1.4** Leading `--` and `/* */` comments stripped before verb
  extraction.

### Table extraction (§2)
- **§2.1** DDL targets visible: DROP / TRUNCATE / ALTER / CREATE /
  GRANT / REVOKE / LOCK / VACUUM / ANALYZE / REINDEX / REFRESH
  MATERIALIZED VIEW / CLUSTER / COMMENT ON TABLE.
- **§2.2** COPY `<table>` FROM/TO `...` captures the table.
- **§2.3** Schema-qualified names emit both `schema.table` and
  `table` so rules written either way fire.
- **§2.4** Quoted identifiers (`"Users"`) captured case-preserved.

### Wire protocol (§4)
- **§4.1** FunctionCall (`'F'`) legacy fast-path now fail-closed.
  Carries no SQL text — there's nothing for the matcher to fire on.
  Bind / Execute / Describe / Close still rely on the Parse-time
  inspection (the SQL text was already seen on `'P'`).

### Postgres-specific blind spots (§6)
- **§6.4** SET ROLE / SET SESSION AUTHORIZATION / SET LOCAL ROLE
  / SET LOCAL SESSION AUTHORIZATION surface as multi-word verbs
  distinct from session-config SETs. Combined with §1.3, blocks
  the `SET ROLE admin; DROP TABLE users` evasion.
- **§6.5** DO `$$ ... $$` block bodies are tokenized; inner
  statements emit shadow sub-statements.
- **§6.6** CALL `<proc>` retains current visibility (verb=`call`,
  proc captured as function). Body inspection is server-side and
  out of practical scope.

### Obviated by earlier CEL refactor
- **§2.8** tables list-OR semantics and **§2.9** negative-list
  semantics were footguns of the old HCL matcher. Operators now
  write `sql.tables.all(t, t in ["public"])` /
  `!sql.tables.exists(t, t == "secret")` explicitly. The parser
  fixes above ensure those CEL predicates see the right candidates.

## Tests

- `TestParseSQL_Audit143` — every in-scope payload from #143
  exercised against `parseSQL`.
- `TestPgEvaluate_Audit143` — multi-statement / CTE-shadow /
  DO-shadow paths end-to-end through `pgEvaluate`.
- `TestPgClientToServerDeniesFunctionCall` — `'F'` wire frame
  fails closed.
- Existing wire-protocol tests still pass.

## Test plan

- [x] `go test ./...` passes locally
- [x] `gofmt -l .` clean
- [x] `go vet ./...` clean
- [ ] CI green


arnauorriols:
I prefer to use an off-the-shelf sql parser, like we do for clickhouse

arnauorriols:
[bot-reply]
Want to confirm direction before swapping out ~760 LOC. Go-native, CGO-free postgres parsers are thin on the ground; the realistic options are:

1. **`github.com/auxten/postgresql-parser`** — pure-Go port of CockroachDB's parser. Closest shape-match to how `clickhouse_native_sql.go` uses `AfterShip/clickhouse-sql-parser` (AST walk for tables/functions, verb from top-level node type). Trade-off: pulls in ~15MB of mod cache and ~340 lines of go.sum (`cockroachdb/errors`, `gogo/protobuf`, `sirupsen/logrus`, `pingcap/errors`, etc.), last release 2021 (v1.0.1).
2. **`github.com/pganalyze/pg_query_go`** — wraps the real libpg_query via CGO. Most accurate. Rules out by current `CGO_ENABLED=0` deploy/release builds (db.go pins `modernc.org/sqlite` for this reason).
3. **`github.com/wasilibs/go-pgquery`** — libpg_query compiled to wasm, no CGO. Real parser, but adds a wasm runtime to the binary.

The PR description specifically motivated the hand-roll with the parser-dependency cost; happy to swap to (1) if you're OK with the dep drag, or (3) if you'd rather pay binary size for fidelity. Prefer (1) given the clickhouse precedent?

Plan if (1) is fine:
- Replace `postgres_sql.go` with an AST-driven extractor mirroring `clickhouse_native_sql.go`'s shape (verb from top-level node, walk for tables/functions, multi-statement = multi-AST).
- Keep `analyseAll` / `parseSQL` / `pgInfo` signatures stable so `postgres.go` and the runtime tests don't churn.
- Keep `extractCTEInner` / `extractDOInner` semantics — surface inner DML statements (audit §1.2 / §6.5) via AST walks instead of token re-tokenisation. DO bodies still need a secondary `parser.Parse` on the body string.
- Rewrite tests in `postgres_runtime_test.go` only where the AST surfaces tokens differently (e.g. `SELECT*FROM x` parses, `BEGIN;` parses, `SELECT;` parses — all the §1.1 / §1.4 cases collapse onto "parser handles it"). Keep the audit §-tagged structure.
- Fall back to a verb-sniff stub on parse failure (mirrors `parseChSQL`'s degraded path).

Confirm parser choice and I'll push.

arnauorriols:
Go with option 1

arnauorriols:
Swapped to `github.com/auxten/postgresql-parser` in 7ddd4a5 — AST walk for the happy path, verb-sniff stub for the postgres-specific syntax cockroach's grammar rejects (LOCK / VACUUM / ANALYZE / REINDEX / REFRESH / CLUSTER / SET ROLE / DO / CALL / trailing-`;` shells). All audit §-tagged tests still pass; updated expectations where the AST surfaces tokens differently (Tables sorted, §1.2 outer verb is `select` with the inner DELETE on analysedStmt.Inner, no more SQL-keyword phantom funcs). `postgres_sql.go` is ~50% smaller. go.sum grows ~290 lines from the parser's transitive deps.

arnauorriols:
[bot-reply]
Read this as an informational status update (commit 7ddd4a5 is the current tip of `polecat/dust-mp2w0uvq`) — no question or change request to action. Standing down; reply or react if a follow-up is wanted.

arnauorriols:
Use pgplex/pgparser instead

arnauorriols:
[bot-reply]
Swapped to `github.com/pgplex/pgparser` (pure-Go port of postgres' own gram.y at REL_17_STABLE) in d5bfc74. pgplex models PG's full grammar, so the parse-tree path now covers LOCK / VACUUM / ANALYZE / REINDEX / REFRESH MATERIALIZED VIEW / CLUSTER / SET ROLE / SET SESSION AUTHORIZATION / DO blocks / CALL natively — no degraded sniff fallback for any of them. The sniff still ships as the fallback for genuinely malformed input (`DROP;` / `SELECT;` audit §1.1 shells, syntax errors).

Differences from the previous extractor:
- Verb derived from concrete node Go type (no `StatementTag`): switch on `*nodes.SelectStmt` → `"select"`, `*nodes.DropStmt` → `"drop"`, etc. `TransactionStmt.Kind` maps BEGIN/COMMIT/ROLLBACK/SAVEPOINT/RELEASE individually; `VariableSetStmt.Name` + `IsLocal` surface `"set role"` / `"set session authorization"` / `"set local role"` / `"set local session authorization"` directly (§6.4).
- Table walker is hand-rolled — pgplex ships no generic `Walk`, so `visit()` switches on each concrete node type. Statement-shaped targets (`DropStmt.Objects`, `TruncateStmt.Relations`, `VacuumStmt.Rels` via `VacuumRelation`, `GrantStmt.Objects` when `Objtype` is table-shaped, `CommentStmt.Object` on tables) emit at the case site; expressions recurse so subqueries / CTE-internal references still show up.
- DO bodies: `DoStmt.Args` carries the body in a `DefElem` named `"as"` whose `Arg` is a `*String`. We pull that out and recurse through `analyseAll` so inner statements still walk the matcher (§6.5).
- Top-level split-then-parse: pgplex returns bare Stmt nodes (no `*RawStmt` wrapper / byte offsets), so we split on top-level `;` ourselves first and parse each piece. `splitTopLevelStatements` / sniff tokenizer / fallback are reused as-is.

All audit §-tagged tests pass (`TestParseSQL`, `TestParseSQL_Audit143`, `TestPgEvaluate_Audit143`); the test expectations didn't need to change. Dep accounting: drop `github.com/auxten/postgresql-parser` and its 19 transitive `cockroachdb/*` / `sentry-go` / `logrus` / etc deps — pgplex/pgparser is dependency-free. `go.sum` shrinks ~294 lines.

arnauorriols:
sync with main and fix conflict

arnauorriols:
[bot-reply]
Pushed cde3078 addressing this comment:
- Merged `origin/main` into the PR branch.
- Resolved the conflict in `config/plugins/endpoints/postgres_runtime_test.go` by keeping both main's truncatable-facet tests (`pgEndpointFromHCL`, `TestPgClientToServer{Denies,Forwards}OversizeFrame…`) and the PR's `Audit143` / `TestPgClientToServerDeniesFunctionCall` tests.
- Made the test-local `passThrough` implement `InspectsTruncatableFacet() bool { return false }` so it satisfies the `match.Matcher` interface that main extended.
- `gofmt -l .`, `go vet ./...`, and `go test ./...` all clean.

PR #295, opened 2026-05-12
feat(config): token_pool block — pool same-type credentials behind one binding

arnauorriols:
## Summary

Adds a new top-level `token_pool` HCL block. Pools group N same-type credentials behind one logical handle so an endpoint binding `credential = X` can reference the pool by name and have the dispatcher spread requests across underlying members per the chosen strategy.

```hcl
credential \"anthropic_oauth_subscription\" \"alice\" {}
credential \"anthropic_oauth_subscription\" \"bob\"   {}
credential \"anthropic_oauth_subscription\" \"carol\" {}

token_pool \"team\" {
  credentials = [alice, bob, carol]
  strategy    = \"round_robin\"  # or \"least_loaded\"
}

endpoint \"https\" \"anthropic\" {
  hosts      = [\"api.anthropic.com\"]
  credential = team
}
```

## Use case

An operator with N personal LLM-subscription credentials (e.g. one Claude OAuth per teammate) declares them as a pool so agent traffic spreads across all N rather than burning through a single credential's monthly quota.

## Scope clarification (important)

This is **resource pooling within a single operator's control** — every member is a credential the operator owns or has consent to use. It is **not**:

- A marketplace between unrelated parties.
- A money-handling system.
- A way to circumvent provider Terms of Service.

Provider terms-of-service may restrict shared subscription use (e.g. some Claude/ChatGPT subscription tiers prohibit multi-user sharing). The gateway flags the pattern in this PR description but does not enforce ToS — the operator decides whether their use qualifies.

## Phase 1 evidence

- Existing dispatch path: `runtime.ResolveCredential` → `cc.Credential.Body.(HTTPCredentialRuntime)` → `secrets.Get(name, profile)` → `InjectHTTP(req, sec)` (`main.go:1823`, mirrored at `config/plugins/endpoints/postgres.go:228`, `config/plugins/endpoints/clickhouse_native_runtime.go:171`, `config/plugins/endpoints/ssh.go:214`, `dial.go:48-58/89-108`).
- Per-credential consumption signal: only `bytes_in/bytes_out` on the `actions` table (`migrations/sqlite/0001_initial.sql:68`); no `credential` column today. v1 falls back to in-memory counters.
- Dashboard surface: `agents.go:621 statusList` walks `policy.Credentials` → `IntegrationRow`. Pools get a parallel `poolsList` returning `PoolRow{name, strategy, plugin_type, members}`.

## Phase 2 design positions

| Question | v1 position | Why |
|---|---|---|
| **Block shape** | `token_pool \"<name>\" { credentials = [bare, names], strategy = \"...\" }`. Single label; lives in the same flat namespace as credentials/etc. | Mirrors `rule` (one-label, one plugin). The endpoint's `credential = X` field accepts either a credential or a pool name — extended `RefSpec` with `AltKinds`. |
| **Strategy default** | `round_robin` | Fair when all members have equal quota; deterministic; trivially correct under contention (atomic counter). |
| **Other strategies** | `least_loaded` ships in v1 (in-memory request count). `exhaust_first` is **rejected at compile** with a clear pointer to use the other two — it requires per-member failure tracking that v1 doesn't yet wire. | Better to fail loudly than silently always-pick-member-0. |
| **Failure handling** | Surface upstream errors honestly. No mid-flight retry across members. No automatic cooldown marking yet. | Mid-flight retry risks double-charging a partially-succeeded request. Cooldown needs the response interception this PR doesn't add (see open questions). |
| **Quota tracking** | In-memory `uint64` counter per member, incremented on `Pick`. Resets on gateway restart. | Sufficient for `least_loaded` decisions inside one process lifetime. Persisted quota across restarts is v2. |
| **Same-(kind, type) constraint** | Compile-time enforced: pooling Claude with Codex fails to load with \"members must share one (kind, type)\". | Different APIs need different injection shapes — the dispatcher can't meaningfully spread across them. |
| **Visibility** | New `/api/state.pools` slice: name, strategy, plugin type, members `[{name, requests, last_use_ns}]`. | Operator can see which pool is active and how its members are sharing load without per-credential observability work. |
| **Multi-pool / shared credential** | A credential can be both pool-member and direct endpoint binding; both share its quota at the provider. The pool's `requests` counter only tracks requests dispatched *through the pool*. | Documented behaviour; not enforced. |

## Open questions for review

1. **Failure handling** — happy to add an opt-in `on_failure = \"retry_next\"` or `on_failure = \"cooldown\"` knob if you'd prefer transparent failover. Skipped here to keep semantics honest.
2. **Cooldown / exhaust_first** — these need response interception (HTTP status, retry-after header). Out of scope for v1; ship as a follow-up once we want per-member failure tracking.
3. **Persisted counters** — adding a `pool_member` column on `actions` would let the dashboard show real lifetime usage. Easy to add when the in-memory counter feels insufficient.
4. **Dashboard UI** — backend now returns a `pools` slice on `/api/state`. Front-end rendering is not in this PR; happy to add a basic card if you want it before merge.
5. **`oauth_dir` field** — unrelated; noting that I added `oauth_dir` to `Gateway.JoinConfig` mapping in a previous bead — not touched here.

## Implementation surface

- `config/plugin.go`: new `KindTokenPool` constant; `RefSpec.AltKinds` for the credential-or-pool ref site.
- `config/token_pool.go`: `CompiledTokenPool` with strategy-aware `Pick(req)`; in-memory per-member counts; `Stats()` for dashboard.
- `config/plugins/pools/pools.go`: HCL plugin (one label, no type), schema-level validation.
- `config/compile.go`: `compileTokenPools` resolves members + same-type constraint; `compileEndpoint` substitutes pool refs into `CompiledCredential.Pool`.
- `CompiledCredential.Resolve(req)`: single helper for all dispatch sites — singular bindings return `Credential`, pool bindings ask the pool to pick a member.
- Dispatch sites updated: `main.go` (HTTPS / WS), `dial.go` (TLS), `config/plugins/endpoints/{postgres,clickhouse_native_runtime,ssh}.go`.
- Dashboard: new `/api/state.pools` slice, profile-filtered.
- Docs: `gateway.example.hcl` example block; `site/doc/config-reference.md` regenerated; `tools/docgen` walks the new `pools` package.

## Test plan

- [x] `go test ./...` — green (config/runtime/all plugin packages, including new TokenPool unit tests covering compile, round-robin, least-loaded, mixed-type rejection, single-member rejection, exhaust-first rejection, unknown-strategy rejection, unknown-member rejection, endpoint→pool→member resolve, stats accumulation, and singular-binding regression).
- [x] `gofmt -l .` — clean.
- [x] `go vet ./...` — clean.
- [x] Emit round-trip — `feature_token_pool.hcl` parses → emits → reparses byte-equivalent.
- [x] Golden refresh — `feature_token_pool.want.json` checked in.
- [ ] Manual smoke against a live gateway with two real Anthropic OAuth subscriptions (out of polecat scope; flagging for reviewer if local credentials available).

### DIFF
--- a/agents.go
+++ b/agents.go
@@ -740,7 +740,7 @@ type PoolMemberRow struct {
 	LastUseNs int64  `json:"last_use_ns,omitempty"`
 }
 
-// PoolRow is the dashboard-facing view of one token_pool block. The
+// PoolRow is the dashboard-facing view of one pool credential. The
 // dashboard renders pools alongside (not inside) integrations because
 // they're a separate concept: a pool composes credentials, it isn't
 // itself a credential.
--- a/agents.go
+++ b/agents.go
@@ -710,10 +710,104 @@ func credentialsInProfile(policy *config.CompiledPolicy, profile string) map[str
 			if cb.Credential != nil {
 				out[cb.Credential.Symbol.Name] = true
 			}
+			// Pool bindings expand to every member — the dashboard's
+			// integration card renders one row per credential, so the
+			// pool's members must each appear in the allowed set.
+			if cb.Pool != nil {
+				for _, m := range cb.Pool.Members {
+					if m != nil && m.Symbol != nil {
+						out[m.Symbol.Name] = true
+					}
+				}
+			}
 		}
 		for tun := ep.Tunnel; tun != nil; tun = tun.Via {
 			if tun.Credential != nil {
 				out[tun.Credential.Symbol.Name] = true
+			}
+		}
+	}
+	return out
+}
+
+// PoolMemberRow is one member's view of a token pool's per-process
+// state. Requests is the in-memory pool counter (resets on gateway
+// restart); LastUseNs is the wall-clock nanosecond of the last
+// dispatch through this member (0 if never used).
+type PoolMemberRow struct {
+	Name      string `json:"name"`
+	Requests  uint64 `json:"requests"`
+	LastUseNs int64  `json:"last_use_ns,omitempty"`
+}
+
+// PoolRow is the dashboard-facing view of one token_pool block. The
+// dashboard renders pools alongside (not inside) integrations because
+// they're a separate concept: a pool composes credentials, it isn't
+// itself a credential.
+type PoolRow struct {
+	Name       string          `json:"name"`
+	Strategy   string          `json:"strategy"`
+	PluginType string          `json:"plugin_type"`
+	Members    []PoolMemberRow `json:"members"`
+}
+
+// poolsList returns the pools slice for /api/state. Filtered by
+// profile when provided — only pools referenced (directly via
+// `credential = pool` or via `credentials = [...]` entries) by an
+// endpoint in the named profile come back. Unfiltered (empty profile)
+// returns every declared pool.
+func (w *webMux) poolsList(r *http.Request) []PoolRow {
+	out := []PoolRow{}
+	policy := w.g.policy.Load()
+	if policy == nil || len(policy.TokenPools) == 0 {
+		return out
+	}
+	profile := r.URL.Query().Get("profile")
+	allowed := poolsInProfile(policy, profile)
+
+	names := make([]string, 0, len(policy.TokenPools))
+	for name := range policy.TokenPools {
+		if allowed != nil && !allowed[name] {
+			continue
+		}
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for _, name := range names {
+		pool := policy.TokenPools[name]
+		row := PoolRow{
+			Name:       pool.Name,
+			Strategy:   string(pool.Strategy),
+			PluginType: pool.PluginType,
+		}
+		for _, ms := range pool.Stats() {
+			row.Members = append(row.Members, PoolMemberRow{
+				Name:      ms.Name,
+				Requests:  ms.Requests,
+				LastUseNs: ms.LastUseNs,
+			})
+		}
+		out = append(out, row)
+	}
+	return out
+}
+
+// poolsInProfile mirrors credentialsInProfile for token pools: returns
+// the bare-name set of pools any endpoint in profile references via
+// its credential bindings. nil means "no filter — return everything."
+func poolsInProfile(policy *config.CompiledPolicy, profile string) map[string]bool {
+	if profile == "" || policy == nil {
+		return nil
+	}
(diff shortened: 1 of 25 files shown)


piscisaureus:
Can you just make this another credential type? 
I don't think it requires another top level concept

arnauorriols:
[bot-reply]
Pushed 0c15914 addressing this comment:
- Dropped `KindTokenPool` and re-registered the pool plugin as `credential "pool" "<name>"`, so HCL now reads `credential "pool" "team" { credentials = [...], strategy = "..." }` and `token_pool` is gone from the surface.
- `Policy.TokenPools` removed; pools live in `Policy.Credentials` alongside everything else. `CompiledPolicy.TokenPools` survives as the runtime-friendly subset (round-robin counter, per-member stats) populated by filtering credentials with `Plugin.Type == "pool"`.
- `RefSpec.AltKinds` removed — its only consumer was the endpoint `credential = X` widening; that ref is now plain `KindCredential` since pools are credentials.
- Pool-of-pools rejection moved from the compile pass into the plugin's `Validate` so the diagnostic points at the offending block.
- Downstream injection path unchanged: `CompiledCredential.Pool.Pick()` still resolves to a real member and runs through the existing per-credential runtime.

arnauorriols:
[bot-reply]
Picked this up as a PR-body dispatch. No code change — the body is a design write-up, and the substantive review ask has already been handled. Flagging two things for the maintainer:

1. **Description is now stale.** The body still leads with a top-level `token_pool` HCL block, but per @piscisaureus' review that concept was dropped — the head commit `0c159144` folds pooling into the credential kind (`credential "pool" "team" { credentials = [...], strategy = "..." }`), and `token_pool` no longer exists on the surface. The opening Summary/HCL example and the "Block shape" row of the design table no longer match what ships. Worth refreshing the description so a reviewer reading top-down isn't evaluating the rejected shape. Happy to rewrite the description to match `credential "pool"` if you'd like — say the word.

2. **Branch conflicts with main.** The PR is currently `CONFLICTING` / `DIRTY` and is 2 commits behind a moved `main`; it needs a rebase before it can merge.

The "Open questions for review" (opt-in `on_failure` failover, cooldown marking, persisted quota) read as product decisions for the maintainers rather than anything actionable from automation — leaving those for you.

piscisaureus:

PR #296, opened 2026-05-13
error: clawpatrol join failed; aborting install

littledivy:
Closes example/orchestrator#38


littledivy:
@ops-bot I don't think this is the right fix? How does it fix the origional issue

PR #353, opened 2026-05-13
fix: download binaries from GitHub Releases instead of Pages

littledivy:
GitHub Pages returns 401 for private repos — `install.sh` was pointing at `denoland.github.io/clawpatrol` which is inaccessible.

Switches to `github.com/denoland/clawpatrol/releases/download` (public GitHub Releases URL) and updates the release workflow to use `gh release create` instead of deploying to Pages.


littledivy:
Closing — the gh-releases switch was intentionally reverted in #342. Need to investigate why before re-landing.

PR #371, opened 2026-05-14
settings credentials: group cards by plugin type

arnauorriols:
On the Settings page's credentials section, cluster credential cards
by plugin type instead of rendering one flat row. Each group gets a
small heading with the plugin display name + a count of cards in the
group; groups are ordered alphabetically by display name, and cards
within a group are sorted by their HCL block name so the layout is
stable across refreshes.

Layout-only — card content/behaviour, the device-page row, and the
overflow modal are unchanged. Grouping is gated on `showAll`, so only
the Settings page picks it up; the device page keeps its compact
connected-first ordering with the "+ K more" overflow.

The group label uses the same `credentialTypeLabel` resolution as the
card title, falling back to the raw HCL type for unknown plugins.
Empty types render nothing (we only iterate over types that appear in
the integration list).

### DIFF
--- a/agents.go
+++ b/agents.go
@@ -568,12 +568,20 @@ func shortHash(s string) string {
 }
 
 type IntegrationRow struct {
-	ID               string                 `json:"id"`
-	Name             string                 `json:"name"`
-	Type             string                 `json:"type"` // credential plugin type
-	HasOAuth         bool                   `json:"has_oauth"`
-	OAuth            *OAuthIntegrationUI    `json:"oauth,omitempty"`
-	Slots            []config.SecretSlot    `json:"slots,omitempty"`
+	ID       string              `json:"id"`
+	Name     string              `json:"name"`
+	Type     string              `json:"type"` // credential plugin type
+	HasOAuth bool                `json:"has_oauth"`
+	OAuth    *OAuthIntegrationUI `json:"oauth,omitempty"`
+	Slots    []config.SecretSlot `json:"slots,omitempty"`
+	// Subtitle is the plugin's account-level identity rendered under
+	// the card heading: the connected-OAuth display_name for
+	// OAuth-flow credentials post-connect, or the HCL-derived
+	// CardSubtitle for plugins like Postgres / Clickhouse that carry
+	// a user field on the body. Empty when the plugin has no useful
+	// identity to surface — the dashboard then omits the subtitle
+	// slot entirely rather than render a placeholder.
+	Subtitle         string                 `json:"subtitle,omitempty"`
 	Connected        bool                   `json:"connected"`
 	ExpiresAt        int64                  `json:"expires_at,omitempty"`
 	DisplayName      string                 `json:"display_name,omitempty"`
@@ -664,6 +672,18 @@ func (w *webMux) statusList(r *http.Request) []IntegrationRow {
 				ConnectURL:    "/api/tailscale/connect?id=" + name,
 				StatusURL:     "/api/tailscale/status?id=" + name,
 				DisconnectURL: "/api/tailscale/disconnect?id=" + name,
+			}
+		}
+		// Card subtitle: prefer the OAuth-supplied display_name when
+		// the credential is a connected OAuth flow; fall back to any
+		// HCL-derived identity (Postgres / Clickhouse user, etc.).
+		// Leaves Subtitle empty when neither source has anything to
+		// say — the dashboard then drops the subtitle slot rather
+		// than rendering a misleading "Saved" placeholder.
+		row.Subtitle = row.DisplayName
+		if row.Subtitle == "" {
+			if sp, ok := ent.Body.(config.CardSubtitleProvider); ok {
+				row.Subtitle = sp.CardSubtitle()
 			}
 		}
 		out = append(out, row)
--- a/config/oauth.go
+++ b/config/oauth.go
@@ -65,3 +65,20 @@ type OptionalScope struct {
 type OAuthFlowProvider interface {
 	OAuthFlow() *OAuthIntegration
 }
+
+// OAuthProfileFetcher is the optional interface an OAuth-flow
+// credential plugin's body implements when it knows how to enrich a
+// freshly-issued access token with the connected account's identity
+// (display name + avatar). The host calls this once at exchange time
+// and persists the result on the credentials row; the dashboard
+// surfaces the display name as the card subtitle so two same-type
+// OAuth credentials can be told apart at a glance.
+//
+// Plugins that don't implement this leave the identity empty — the
+// dashboard then shows no subtitle (the green dot + status text still
+// indicate the connection state). Implementations should be
+// best-effort: a failed userinfo fetch returns ("", "") rather than
+// surfacing an error to the connect flow.
+type OAuthProfileFetcher interface {
+	FetchOAuthProfile(accessToken string) (displayName, avatarURL string)
+}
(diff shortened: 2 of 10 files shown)


arnauorriols:
Start a gateway with full.hcl and take a screenshot of the settings page

arnauorriols:
[bot-reply]
Started a gateway against `config/testdata/full.hcl` (with the operational fields `listen` / `state_dir` / `insecure_no_dashboard_secret` / `control = "wireguard"` / `wg_subnet_cidr` prepended so the file loads as a runnable config), then captured the settings page on PR #371's head (`e6ad1e1`).

Screenshot: settings page

It shows the 31 declared credentials grouped under per-plugin-type section headers (AWS, BEARER TOKEN — 12, CLAUDE — 1, CLAUDE (API KEY) — 1, CLICKHOUSE — 1, CODEX — 2, COOKIE TOKEN — 1, GEMINI — 1, HEADER TOKEN — 2, MTLS — 2, NOTION — 1, POSTGRES — 3, SLACK — 3, TELEGRAM — 2), with the count next to each header and the gateway.hcl editor below the credentials section as before.

arnauorriols:
1. First, group by "Not Configured" / "Connected"
2. Instead of stacking the groups vertically, group them in parent boxes that are stacked horizontally, wrapping on line overflow
3. the boxes should show the username when connected, right now rthey show saved

arnauorriols:
[bot-reply]
Pushed eddfb28 addressing items 1 and 2:

- Top-level grouping is now **Connected / Not configured**, with the
  plugin-type clusters nested inside each state section. Empty
  sections are dropped.
- The per-plugin-type clusters are now boxed containers laid out
  horizontally with `flex-wrap`, replacing the vertical stack.

**Question on item 3** — "the boxes should show the username when
connected, right now they show saved":

The "Saved" subtitle is the fallback in `contextualSubtitle()` for
connected credentials whose `display_name` is empty. OAuth flows
populate `display_name` from the connected account (GitHub login,
Slack workspace, etc.) and that's surfaced already. For non-OAuth
credentials (Postgres / Clickhouse / SSH / mTLS / bearer-token / …)
the server-side `IntegrationRow` doesn't carry a username field
today — and most of those plugins don't have a uniform "username"
concept either (SSH carries no user on the credential, mTLS has a
CN, Postgres has an HCL `user` field, raw bearer tokens have nothing).

Which of these did you have in mind?

  (a) Surface the HCL `user` field for Postgres / Clickhouse
      specifically (and show "Saved" or nothing elsewhere) — needs a
      small `IntegrationRow` extension on the server.
  (b) Drop the "Saved" fallback entirely and let the green dot +
      heading carry the "connected" signal — minimum change, no new
      data on the wire.
  (c) Show the credential's HCL bare name (already in the heading) —
      i.e. just remove the redundant subtitle.
  (d) Something else — e.g. a specific credential type whose card
      you're looking at where a username clearly exists?

Happy to implement whichever you pick — leaning toward (a) for
Postgres/Clickhouse plus (b) for everything else, but flagging
before I touch the server.

arnauorriols:
Currently all credentials are showing "saved". The contextualSubtitle is not working, not even for oauth credentials. This needs to be fixed

arnauorriols:
[bot-reply]
Pushed edff0a1 — the contextual subtitle is now resolved server-side
into a new `IntegrationRow.Subtitle` field that the dashboard surfaces
verbatim. Two sources feed it:

- **OAuth credentials**: the connected-account display_name. The host
  previously keyed the userinfo fetch on the credential's *bare name*
  (only the literal id `github` fired), so any operator-chosen name
  (`github-ops-pat`, `claude`, `codex`, …) silently skipped it and
  the card fell through to "Saved". Userinfo fetching now lives on
  the plugin body via a new `OAuthProfileFetcher` interface; the
  github_oauth plugin implements it, so any github_oauth credential
  populates display_name on connect regardless of bare name.
- **Postgres / Clickhouse**: the HCL `user` field, via a new
  `CardSubtitleProvider` interface implemented by those two plugins.
  So `pg-corp-ro` and `pg-corp-rw` against the same DB now read
  apart by their respective postgres users.

When neither source has anything to say (bearer tokens, mTLS,
anthropic_oauth_subscription, openai_codex_oauth without a userinfo
implementation, …), Subtitle is empty and the dashboard omits the
subtitle slot — the green connection dot + the existing status row
(`connected` / `expires …`) carry the connection signal without the
misleading "Saved" placeholder.

Follow-up worth filing separately: `anthropic_oauth_subscription` and
`openai_codex_oauth` could implement `OAuthProfileFetcher` against
their userinfo endpoints (Anthropic's `user:profile` scope, OpenAI's
OIDC `/userinfo`) — left untouched here so this PR stays scoped to
the subtitle plumbing.

arnauorriols:
Take a new screenshot

arnauorriols:
[bot-reply]
Rebuilt the dashboard against the PR head (`edff0a1`) and re-captured the settings page running against `config/testdata/full.hcl`.

Screenshot: settings page

Catalog: 33 declared credentials, all currently in **Not configured** (no slots filled in this fixture). The per-plugin-type clusters now sit inside that status section as bordered boxes flowing horizontally and wrapping: AWS — 1, BEARER TOKEN — 12, CLAUDE — 1, CLAUDE (API KEY) — 1, CLICKHOUSE — 1, CODEX — 2, COOKIE TOKEN — 1, GEMINI — 1, HEADER TOKEN — 2, MTLS — 2, NOTION — 1, POSTGRES — 3, SLACK — 3, TELEGRAM — 2. Cards show the server-resolved subtitle ("Not configured" here, since nothing is connected) instead of the old "Saved" fallback.

arnauorriols:
Make the credential type box less noisy. Take another screenshot after the change

arnauorriols:
[bot-reply]
Pushed 9278c95 addressing this comment:

- Dropped the heavy `border-2 border-navy bg-canvas-light` box on each plugin-type cluster. The cluster heading (e.g. `BEARER TOKEN — 12`) now sits directly above its cards, separated from neighbours by a wider `gap-x-6 gap-y-4`. The cards themselves carry the visible weight; the captions just delimit the groups.

Screenshot: settings page

arnauorriols:
[bot-reply]
Confirming the screenshot request from this comment was answered earlier in the thread and is current.

- Original screenshot reply: https://github.com/denoland/clawpatrol/pull/371#issuecomment-4446820411
- Latest screenshot against current PR head (`9278c95`): https://github.com/denoland/clawpatrol/blob/screenshots/pr-371-9278c95-vault/pr-371-settings.png?raw=true (posted in #issuecomment-4453788877)

I re-ran the same setup locally (gateway built from `9278c95`, `config/testdata/full.hcl` with operational fields — `listen` / `info_listen` / `state_dir` / `insecure_no_dashboard_secret` — prepended so the file loads as a runnable config) and captured `/#/settings`. The dashboard matches the linked image: 33 declared credentials in the **Not configured** section, plugin-type clusters (AWS · BEARER TOKEN — 12 · CLAUDE · CLAUDE (API KEY) · CLICKHOUSE · CODEX — 2 · COOKIE TOKEN · GEMINI · HEADER TOKEN — 2 · MTLS — 2 · NOTION · POSTGRES — 3 · SLACK — 3 · TELEGRAM — 2) laid out horizontally with `gap-x-6 gap-y-4` and no heavy bordered box around each cluster — i.e. the `9278c95` "drop heavy borders on type-clusters" change is reflected. No new screenshot upload needed.

PR #403, opened 2026-05-14
sql_rule: add `database` match facet

arnauorriols:
## Summary

Adds a session-scoped `sql.database` binding so a `sql_rule`'s CEL condition can scope by which database the agent is talking to:

```hcl
rule "sql_rule" "no-prod-deletes" {
  endpoint = pg-cluster
  match    = sql.database == "prod" && sql.verb == "delete"
  verdict  = "deny"
}
```

## Population per protocol

- **postgres**: pulled from the agent's StartupMessage `database` parameter at session start, threaded through `pgClientToServer` and stamped on every `match.Request.Meta` (including the truncated-frame path).
- **clickhouse_native**: pulled from `Hello.Database` at session start and threaded through `chRunSession` → `chAgentToServer` → `chEvaluateSQL`.
- **clickhouse_https**: extracted from the `database` URL query parameter or `X-ClickHouse-Database` header (query parameter wins when both are present).

## Case sensitivity

Postgres treats database names as case-sensitive identifiers, so the want-side string literal in the CEL condition is **not** auto-lowercased the way `sql.verb` is. `sql.database == "Prod"` matches `"Prod"` only.

## Limitations (deferred to v2)

- Mid-session database changes via postgres `\connect` or dialect `USE` statements are not tracked; the session-start value is canonical for the lifetime of the connection.
- The `clickhouse_https` endpoint has no wire-level MITM dispatch today (Transport=`""`), so the helper added here is exercised by unit tests but not yet plumbed into a request loop. The extraction surface is in place for when MITM wiring lands.

## Test plan

- [x] `gofmt -l .` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] New unit tests:
  - `TestSQLMatcherDatabaseCaseSensitive` — case-sensitive match across the facet.
  - `TestPgStartupParamExtractsDatabase` + `TestPgEvaluateThreadsDatabaseIntoMeta` — postgres population + matcher fires.
  - `TestChEvaluateSQLThreadsDatabaseIntoMeta` — clickhouse_native Hello → matcher.
  - `TestClickhouseHTTPSDatabase` — URL/header extraction order.


arnauorriols:
Database changes mid-session should be tracked. It is not acceptable that actions report the wrong database (which happens for all queries after the database change, if we don't track the database change)

arnauorriols:
[bot-reply]
Pushed 6ce0c84 addressing the mid-session-database-change comment:

- **ClickHouse native**: `USE db` statements are now tracked. A connection-scoped database tracker seeds from `Hello.Database` and rolls forward when the matcher allows a `USE` through; subsequent queries on the same connection report the new value as `sql.database` to the matcher. Denied `USE` leaves the tracker untouched (the statement never reached upstream, so the session is still on the old database). The USE statement itself evaluates against the pre-swap value — the rule decides whether the swap is allowed based on the database the agent is leaving.
- **ClickHouse HTTPS**: already per-request via the `database` query param / `X-ClickHouse-Database` header, so no change needed.
- **Postgres**: no code change. Postgres has no in-protocol way to switch the active database on an open connection — there's no `USE` analogue, and `\connect newdb` is a psql *client* meta-command that closes the current TCP connection and opens a fresh one (which arrives at the gateway as a new `ConnHandle` with its own `StartupMessage`, so the session-start database is always authoritative). The pgClientToServer godoc now spells this out instead of mislabelling it as a \"v1 limitation\".

New tests:
- `parseChSQL("USE metrics")` surfaces `UseDatabase=metrics`.
- `TestChAgentToServerUseDatabaseUpdatesSessionScope` runs `USE metrics; DROP TABLE events` against a `sql.database == "metrics" && verb == "drop"` rule and verifies the DROP is denied (pre-fix it slipped through) and that the per-statement meta sees `default` for the USE and `metrics` for the DROP.
- `TestChAgentToServerUseDatabaseStickyOnDeny` verifies a denied `USE` does not poison the session — the next SELECT still sees the original database.

arnauorriols:
Sync with main and resolve conflicts

arnauorriols:
[bot-reply]
Pushed d4b3a56 syncing with main and resolving the conflict in `config/plugins/endpoints/clickhouse_https.go`:
- main #404 landed an exported `ClickhouseHTTPSDatabaseFromRequest` + `PickClickhouseHTTPSEndpointByDatabase` + a `DispatchDatabase` method, covering the same query/header extraction this PR's v1 `clickhouseHTTPSDatabase` did. Took main's version and folded this branch's matcher note ("returns empty so `sql.database == \"...\"` won't match") into the doc.
- Removed the now-redundant `config/plugins/endpoints/clickhouse_https_test.go` — main's new `clickhouse_database_routing_test.go::TestClickhouseHTTPSDatabaseFromRequest` covers the same cases (query/header precedence, case preserved, nil request).
- All other touched files (clickhouse_native_runtime.go, etc.) auto-merged cleanly.

Verified locally: `gofmt -l .` clean, `go vet ./...` clean, `go build ./...` clean, `go test ./...` green.

PR #407, opened 2026-05-14
clickhouse_native: parse CTE-prefixed INSERTs

arnauorriols:
AfterShip's parser refuses `WITH … INSERT INTO X SELECT …`, which is routine in ClickHouse analytics. Today the fallback sniffs "WITH" as the verb and produces empty tables / functions — silently disabling every write-gating `sql_rule` on these queries.

When the parser fails on an input that begins with `WITH`, locate the top-level `INSERT` keyword (string-/comment-/paren-aware), strip the `INSERT INTO X [(cols)]` clause, and re-parse the residual `WITH … SELECT`. Pin `verb = insert` and graft the target onto the table set; the existing AST walker pulls source tables and functions out of the body.

CTE-local names stay in the tables list (v1 decision).

Test fixture: the multi-CTE INSERT reported by the user that triggered this bug.


arnauorriols:
This solution is not generic enough. We want to handle soundly any query that the sql parser is not able to properly parse. 

What we should do instead, is that when the parser fails to parse the statement, we don't set the sql facet fields verb, functions and tables (we still set statement though). A new boolean flag in the Request is set to true (similar to the Truncated flag), indicating that the request could not be parsed. Similarly to the matching semantics when the statement is truncated, when a rule tries to match a Request where Unparseable = true, and it tries to match against an sql facet field that was not set and therefore has the zero value, the rule should automatically be considered a match, and the verdict should be "deny: unparseable query". However, rule priority should be honored: if there are rules with higher priority that do not try to match against these sql fields (for example, a rule that matches against sql.statement), the action should still be evauated by the matching rule as normal.

arnauorriols:
[bot-reply] Pushed f942c5c that rips out the CTE-rewrite hack and replaces it with a generic `Unparseable` flag — same shape as `Truncated`, applied to the parser-failure axis.

**Plumbing** (mirrors Truncated end-to-end):

- `match.Request.Unparseable bool`, set by the wire frontend when its parser refuses the Query bytes (Statement is still populated; verb/tables/functions are zero).
- `Matcher.InspectsUnparseableFacet() bool`, pre-computed at compile time the same way `InspectsTruncatableFacet()` is — `CompileCondition` gains an `unparseablePaths` argument and walks the AST once.
- Dispatcher: when `req.Unparseable && rule reads an unparseable facet`, synthesize a deny attributed to that rule. Rule priority is walked first; a higher-priority statement-only rule still gets its honest Match call.

**SQL facet** declares `unparseablePaths = ["sql.verb", "sql.tables", "sql.functions"]`. `sql.statement` is intentionally excluded — the raw text is populated regardless of parse success, so `statement.contains(...)` / `statement.matches(...)` rules continue to evaluate honestly.

**ClickHouse parser**: `parseChSQL` returns `(chSQLInfo, bool)`. Parse failure is one branch — Statement-only, every other facet zero. The CTE+INSERT shape, exotic SYSTEM commands, and future syntax the grammar doesn't admit all travel this one fail-closed path. No more bespoke rewrites or verb-sniffing.

**`SQLParser.ParseStatement`** widens to `(any, bool)` so `clawpatrol test` fixtures see the same shape live dispatch sees. Postgres returns `false` for now; **cl-o4xf** tracks migrating its `sniffStmt` path onto the contract (its current design from cl-1y3's audit work deliberately sniffs verb/tables on failure, so a coordinated migration is the right shape).

**Tests** cover all six PR-comment scenarios:

- Unparseable + high-priority statement rule that matches → allow.
- Unparseable + statement rule that doesn't match + verb rule → synth deny on the verb rule (priority honored).
- Unparseable + only parser-facet rules → synth deny on the highest-priority one; lower-priority rules don't run.
- Unparseable + no rules → fall through (no auto-deny when no rule references an unset facet).
- Parseable query → unchanged.
- CTE+INSERT regression at the `chEvaluateSQL` boundary → still denies, via the new generic path.

config/runtime/dispatch.go
```diff
@@ -107,6 +125,21 @@ func synthesizeTruncatedDeny(r *config.CompiledRule) *config.CompiledRule {
 	return &synth
 }
 
+// synthesizeUnparseableDeny mirrors synthesizeTruncatedDeny for the
+// parser-failure gate. The reason names the rule whose contract the
+// unparseable-query case broke, so logs / dashboard cards attribute
+// the synthesized deny to the matching rule rather than to an opaque
+// "unparseable" line item.
+func synthesizeUnparseableDeny(r *config.CompiledRule) *config.CompiledRule {
+	reason := "rule \"" + r.Name + "\" references a SQL facet (verb / tables / functions) that the gateway's parser could not derive from the unparseable query; failing closed"
```
arnauorriols: This error message should be generic, not specific to SQL

config/plugins/endpoints/postgres.go
```diff
@@ -127,14 +127,19 @@ func (PostgresEndpointRuntime) DetectPlaceholder(req *runtime.Request, candidate
...
 // Returns *sqlfacet.Meta as `any` to keep the runtime interface free
 // of the facets-package import.
-func (PostgresEndpointRuntime) ParseStatement(sql string) any {
+func (PostgresEndpointRuntime) ParseStatement(sql string) (any, bool) {
 	info := parseSQL(sql)
+	// Postgres hasn't adopted the Unparseable contract yet — the
+	// parser-failure path sniffs verb / tables from a token scanner
+	// rather than reporting "parser refused, fields zero, set the
+	// flag." Until the follow-up bead migrates that path, fixtures
+	// always see unparseable=false on the postgres runtime.
```
arnauorriols: Adopt Unparseable contract in Postgres now, in this PR.

config/extplugin/celenv.go
```diff
@@ -55,7 +55,12 @@ func newPluginFacetMatcher(facetName, condition string, streamFields []string) (
 	for _, f := range streamFields {
 		truncatable = append(truncatable, facetName+"."+f)
 	}
-	inner, err := match.CompileCondition(env, condition, buildAct, nil, truncatable)
+	// Plugin facets have no parser stage on the gateway — their action
+	// payload is decoded JSON, validated by the plugin itself, not
+	// parsed by a grammar that could refuse the bytes. Pass nil for
+	// unparseablePaths so the Unparseable gate is a no-op for plugin
+	// facet matchers.
```
arnauorriols, line 62: We need to add the unparseable contract to plugins

arnauorriols (commented):
feedback

arnauorriols:
[bot-reply]

Three review comments addressed in 3 commits on top of the existing
Unparseable contract:

- [r3257958948](https://github.com/denoland/clawpatrol/pull/407#discussion_r3257958948) — `dispatch.go:134` synthesised deny string is now generic ("facet that the gateway's parser could not derive from the unparseable request"). Surrounding doc comments updated to say "parser" rather than "SQL parser" so other facet families inherit the wording. (8a9e4b1)

- [r3257992020](https://github.com/denoland/clawpatrol/pull/407#discussion_r3257992020) — postgres adopts the Unparseable contract: `parseSQL` returns `(pgInfo, unparseable bool)` matching `parseChSQL`'s shape, the sniff verb-fallback is removed (its purpose is now served end-to-end by the dispatcher's fail-closed gate), `pgEvaluate` stamps the flag onto each per-statement `match.Request`, and `ParseStatement` returns it so fixtures see the same shape live dispatch sees. New tests pin pgEvaluate's synth-deny + statement-only-rule-still-allows parity end-to-end. (9a01425)

- [r3258016524](https://github.com/denoland/clawpatrol/pull/407#discussion_r3258016524) — `EvaluateAction` gains `bool unparseable = 6`; the adapter propagates it to `match.Request.Unparseable` for both plugin and built-in facet flows; `newPluginFacetMatcher`'s CEL env exposes `unparseable` as a top-level bool binding, and `pluginMatcher.InspectsUnparseableFacet` forwards the inner answer so plugin-facet rules that read it fail-close on Unparseable=true. `match.CompileCondition` learns the bare-ident path shape since the plugin binding lives at the env top level. (50590c1)

arnauorriols:
Revert 50590c196b092be2c954989a64460b49580607df

arnauorriols:
[bot-reply] Reverted 50590c1 in 6ec5410.

arnauorriols:
sync with main and fix git conflicts

arnauorriols:
[bot-reply]

Synced with `origin/main` (34 commits ahead). Conflicts resolved per-file:

- **`action_file.go`** — *hand-merged*. Kept the cte-branch's `parseSQL func(string) (any, bool)` contract (post-revert state). Combined with main's stricter fixture-vs-parser validation (#423) by treating fixture facets as fillers when the parser produced nothing (clears `unparseable`), and as assertions that must agree otherwise (`sql.verb / tables / functions mismatch`). `a.SQL.Database` override from main (#410) added as additive — parser can't see it, fixture wins.
- **`config/plugins/endpoints/clickhouse_native_runtime.go`** — *hand-merged*. Took main's `chEvaluateSQL(ctx, ch, sql, credName, database, truncated) (verdict, reason, nextDB)` signature + `defer` USE-tracking (#404/#410), kept HEAD's `info, unparseable := parseChSQL(sql)` and `Unparseable: unparseable` propagation on `match.Request`. Doc comment combines both — main's three-return contract plus the unparseable note from cl-3lsy.
- **`config/plugins/endpoints/clickhouse_native_sql.go`** — *hand-merged*. Kept HEAD's `(chSQLInfo, bool)` return + `chWalkSQL` helper (cl-myyc). Inserted main's `USE db` head-statement detection (sets `info.UseDatabase`) before the walk — additive.
- **`config/plugins/endpoints/postgres.go`** — *hand-merged*. `pgEvaluate` now takes `database` (from main) and `pgEvaluateInfo` takes both `database` (main) and `unparseable` (HEAD/cl-3lsy). Body was auto-merged correctly — fix was just signature reconciliation.

Knock-on test edits (necessary for the file to compile after the merged signatures):

- `action_file_test.go` — `TestFixtureSQLStrictAgainstParser`'s `parser` lambda updated to `func(string) (any, bool)` returning `(meta, false)`.
- `config/plugins/endpoints/clickhouse_native_test.go` — `TestChEvaluateSQLCTEInsertSynthDenies` call updated to the new 6-arg / 3-return form.
- `config/plugins/endpoints/postgres_runtime_test.go` — two `pgEvaluate` call sites updated to pass `database=""`; dropped invalid `database = "app"` HCL fields from the test endpoints (#422 removed that arg from the postgres endpoint schema).

Verification:

- `go build ./...` clean.
- `go vet ./...` clean.
- `go test ./...` all pass (including `TestFixtureSQLStrictAgainstParser`, `TestChEvaluateSQLCTEInsertSynthDenies`, `TestPgEvaluateUnparseableSynthDeny`, `TestPgEvaluateUnparseableLetsStatementOnlyRuleRun`).
- `gofmt -l .` clean.
- `cd www && npx oxfmt --check …` clean.
- `cd www && npm run build` clean.

No substance changes beyond conflict resolution and the knock-on test edits above. (cl-uiv4)

Issue #401, opened 2026-05-14
Dual-mode gateway: tailscale-control + WG-control simultaneously, so clawpatrol run works on tailnet-only deployments


ry:
## Background

`cfg.Control` is today an enum: `"wireguard"` xor `"tailscale"`. Picking one foregoes a capability the other has:

- **`wireguard` control**: `clawpatrol run` per-process tunneling works (the WG-mode onboarder writes a wg.conf that the namespaced child consumes). But device onboarding has to invent its own device-flow.

- **`tailscale` control**: tsnet listener on the tailnet IP, identity via tsnet whois, automatic auth-key minting for clients via OAuth. But `clawpatrol run` is unavailable — there's no wg.conf for the namespaced child to consume; the device joins the tailnet as an exit-node consumer and routes whole-machine.

For tailnet-only deployments (no public IP — see `security-refactoring.md`'s Phase 0 migration plan), the natural posture is:

- Tailscale as the transport / identity layer
- clawpatrol WG as the policy-enforcement layer

The today-path is to run **two layers of WG** (Tailscale's WG carrying clawpatrol-WG payloads). The overhead is essentially free at modern CPU speeds, but it's architecturally redundant.

## Proposal

Allow `cfg.Control` to be set to both. Concrete shape:

```hcl
control = ["tailscale", "wireguard"]   # OR a new enum value like "dual"
```

Effects:

- `StartWGServer` still runs (clawpatrol WG endpoint on tailnet IP, listening UDP/51820 inside the tailnet)
- `openListener` still returns a tsnet listener for the dashboard / MITM surface
- The onboarder dispatcher mints both a Tailscale auth key (for the device to join the tailnet) and a wg.conf (for `clawpatrol run`'s per-process model)
- Clients use the Tailscale auth key for whole-machine, the wg.conf for per-process via `clawpatrol run`

## Scope

Looking at the code (`tailscale.go::openListener`, `main.go::runGateway`, `onboard.go::newOnboarder`), the two control modes are largely independent code paths that just happen to be mutually exclusive at dispatch. Relaxing the exclusivity is roughly:

- HCL schema: accept `[]string` or a new enum
- `runGateway`: run both startup paths in parallel
- `newOnboarder`: return a `dualOnboarder` that calls both
- `clawpatrol join`: handle the dual response (both keys returned)

Rough estimate: ~50–150 LOC + tests + docs.

## Alternatives considered

[...]

PR #408, opened 2026-05-15
clawpatrol run: tsnet path + gateway advertise-exit-node

piscisaureus:
* `gateway`: when tsnet comes up (HCL `authkey` or `TS_AUTHKEY`),
  EditPrefs to advertise both `0.0.0.0/0` and `::/0` so the
  gateway appears as a usable exit node. Failures are logged but
  don't gate the listener (the tailnet admin still has to approve
  the routes).
* `clawpatrol join --tsnet-authkey KEY --tsnet-exit-node NAME`
  brings up an embedded tsnet node, validates the named gateway
  is visible as a peer, tears down, and persists the node
  identity under `<ca-dir>/tsnet-client/` plus the operator's
  exit-node / control-url / hostname next to ca.crt. The
  device-flow / WG onboard is skipped — tsnet clients reach the
  gateway via tailnet identity, not a registered WG peer. The
  authkey is one-shot and never written to disk.
* `clawpatrol run` auto-detects the persisted tsnet bundle and
  selects the tsnet path (no flags needed); otherwise it stays
  on the wg.conf flow. The wg-up signal pipe carries a 1-byte
  length plus the netns-side address line so the parent can hand
  the tsnet-discovered tailnet IPs to the namespaced child (the
  WG path also routes its address through this pipe; the
  env-based fallback in the child is preserved for compatibility).

Tests cover exit-node prefs resolution by IP / hostname (case +
FQDN with/without trailing dot) / no-match / peer-without-IP, and
the join-config persist / load round-trip.

### DIFF
--- a/main.go
+++ b/main.go
@@ -2395,10 +2395,10 @@ usage:
                   --hostname NAME        device name to register (default: os.Hostname)
                   --profile NAME         suggest a profile for the approver
                   --whole-machine        bring up wg-quick (route all traffic)
+                  --tsnet-authkey KEY    onboard via embedded tsnet client (skips wg/host-tailscale)
+                  --tsnet-exit-node NAME gateway tailnet hostname or IP (required with --tsnet-authkey)
   clawpatrol login                       onboard this machine (tailscale path)
-  clawpatrol run -- <cmd> [args...]      route one process tree through gateway (WG via wg.conf)
-                  --tsnet-authkey KEY    use embedded tsnet client instead of WG
-                  --tsnet-exit-node NAME tailscale exit-node hostname or IP (required with --tsnet-authkey)
+  clawpatrol run -- <cmd> [args...]      route one process tree through gateway
   clawpatrol status                      report install + tunnel state
   clawpatrol uninstall                   remove local join state and tunnel config
   clawpatrol env                         print shell exports for sourcing
--- a/main.go
+++ b/main.go
@@ -2396,7 +2396,9 @@ usage:
                   --profile NAME         suggest a profile for the approver
                   --whole-machine        bring up wg-quick (route all traffic)
   clawpatrol login                       onboard this machine (tailscale path)
-  clawpatrol run -- <cmd> [args...]      route one process tree through gateway
+  clawpatrol run -- <cmd> [args...]      route one process tree through gateway (WG via wg.conf)
+                  --tsnet-authkey KEY    use embedded tsnet client instead of WG
+                  --tsnet-exit-node NAME tailscale exit-node hostname or IP (required with --tsnet-authkey)
   clawpatrol status                      report install + tunnel state
   clawpatrol uninstall                   remove local join state and tunnel config
   clawpatrol env                         print shell exports for sourcing
--- a/run_linux.go
+++ b/run_linux.go
@@ -79,15 +79,10 @@ func runRun(args []string) {
 	fs := flag.NewFlagSet("run", flag.ExitOnError)
 	confPath := fs.String("conf", defaultRunConf(), "path to wg conf written by `clawpatrol join`")
 	noAutoExpose := fs.Bool("no-auto-expose", false, "disable the seccomp relay that mirrors TCP listeners inside the netns back to the host")
-	tsnetAuthKey := fs.String("tsnet-authkey", os.Getenv("CLAWPATROL_RUN_TSNET_AUTHKEY"), "tailscale authkey for the embedded tsnet client (selects tsnet path; otherwise the WG `wg.conf` path is used)")
-	tsnetExitNode := fs.String("tsnet-exit-node", os.Getenv("CLAWPATROL_RUN_TSNET_EXIT_NODE"), "tailscale exit-node hostname or IP — required with --tsnet-authkey")
-	tsnetControlURL := fs.String("tsnet-control-url", os.Getenv("CLAWPATROL_RUN_TSNET_CONTROL_URL"), "override tailscale control-plane URL")
-	tsnetHostname := fs.String("tsnet-hostname", os.Getenv("CLAWPATROL_RUN_TSNET_HOSTNAME"), "tailscale hostname for the embedded tsnet client (default: clawpatrol-run-<pid>)")
-	tsnetStateDir := fs.String("tsnet-state-dir", os.Getenv("CLAWPATROL_RUN_TSNET_STATE_DIR"), "tsnet state dir (default: <clawpatrol-dir>/run-tsnet)")
 	_ = fs.Parse(args)
 	cmd := fs.Args()
 	if len(cmd) == 0 {
-		fail("usage: clawpatrol run [--conf <path>] [--no-auto-expose] [--tsnet-authkey KEY --tsnet-exit-node NAME-OR-IP] -- <cmd> [args...]")
+		fail("usage: clawpatrol run [--conf <path>] [--no-auto-expose] -- <cmd> [args...]")
 	}
 	// Honour both the flag and the env var so children re-execed by this
 	// process see the same setting without round-tripping the flag.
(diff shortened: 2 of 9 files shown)


ry:
closes https://github.com/denoland/clawpatrol/issues/401 (theoretically)

piscisaureus:
Two gaps in the `clawpatrol run` story worth thinking about before
this lands:

**1. Concurrent `clawpatrol run` from one device (Linux).**

In WG mode every `clawpatrol run` mints its own keypair, POSTs it to
`/api/peer/ephemeral`, gets a fresh `10.55.x.y/32` from the gateway,
and DELETEs it on exit (`ephemeral_peer.go` + `run_linux.go`'s call
to `ephemeralPeer(cfg)`). Concurrent runs on one device are isolated
identities + IPs.

This branch skips `ephemeralPeer` when `useTsnet` is true and points
every invocation at the same persistent tsnet state dir:

```go
// run_linux.go
if !useTsnet {
    cleanupEphemeral, _ := ephemeralPeer(cfg)
    defer cleanupEphemeral()
}
...
// run_tsnet_linux.go
srv := &tsnet.Server{
    Dir: opts.stateDir,           // <ca-dir>/tsnet-client, shared across runs
    Tun: tunDev,
    ...
}
```

`tsnet.Server` wants exclusive access to `Dir` (it holds the node
identity + IPN state), and the join-time authkey produces a
non-ephemeral device. A second concurrent `clawpatrol run` on the
same device either trips the lockfile or both processes thrash the
same tailnet session.

Closing this gap basically requires the gateway-side ephemeral
authkey endpoint from #413 (`POST /api/peer/ephemeral/tsnet`) plus a
per-invocation `os.MkdirTemp` + `tsnet.Server{Ephemeral: true}` on
the client. Worth pulling that piece in here rather than landing the
exit-node mechanism and discovering the concurrency regression later.

**2. macOS `clawpatrol run` (NE process filter).**

`run_darwin.go` is untouched. The macOS path goes through
`Clawpatrol.app`, whose `NETransparentProxyProvider` extension reads
`~/.config/clawpatrol/wg.conf` and speaks userspace WireGuard to
`cfg.WGEndpoint`. If the gateway flips to `control = \"tailscale\"`
the WG listener is gone and the extension has nothing to dial. The
`mode` marker file `runJoinTsnet` writes is never read on Darwin, so
the failure surfaces as a hang rather than a clear error.

Two options:

* Add a fail-fast stub in `run_darwin.go` analogous to #413's (\"not
  supported on macOS in tailscale mode, rejoin with WireGuard\") so
  the failure mode is at least obvious.
* Or land the dual-control gateway shape from #401 first — keep the
  WG listener up alongside tsnet so the macOS NE extension keeps
  working while Linux uses the new path. That's the only way both
  platforms work with one gateway config.

Without one of those, this PR effectively makes the gateway a
Linux-only target the moment an operator switches `control` to
`tailscale`.

piscisaureus:
Followup on the concurrency gap — there's a wider design space here
than just \"mint a fresh key per run\". Putting these out as a menu
since the right choice depends on how transparent we want `clawpatrol
run` to stay for non-HTTP protocols.

The key insight: the constraint isn't \"one tailnet identity per
outbound IP\" the way it is for WG. A single `tsnet.Server` happily
multiplexes many concurrent `Dial`s — wireguard-go demuxes by 5-tuple
and the gateway just sees N connections from one `100.64.x.y` with
different source ports. All concurrent runs from one device *are* the
same device, so collapsing them onto one tailnet identity matches
the audit/profile model rather than working around it.

Options, roughly in decreasing transparency:

**A. Daemon at TUN-fd level.** `clawpatrol-tsnetd` owns one
`tsnet.Server`. Each `run` hands its netns TUN fd over via
SCM_RIGHTS; daemon attaches it as a new NIC on a shared gVisor
netstack; the netstack's TCP/UDP forwarder calls `srv.Dial`. One
tailnet identity, N concurrent NICs.
- Pros: fully transparent inside the netns (any app, any protocol),
  one device in the admin panel, tsnet bring-up amortized across
  every run for the rest of the session.
- Cons: netstack in the daemon, version-skew handling on binary
  upgrade, daemon crash takes out all tunneled processes.

**B. Daemon as a SOCKS5 / HTTP proxy.** Daemon exposes a SOCKS5
endpoint backed by `tsnet.Server.Dial`. `clawpatrol run` sets
`ALL_PROXY` / `HTTPS_PROXY` in the child's env. No TUN, no netstack
— the daemon is ~150 lines.
- Pros: dramatically simpler than (A), trivial to test.
- Cons: not transparent. Apps that don't honor `HTTPS_PROXY` (raw
  `psql`, ClickHouse native binary, ssh without `ProxyCommand`)
  bypass it. Fine for `clawpatrol run claude`, broken for
  `clawpatrol run psql -h prod-db`.

**C. Daemon as a TCP-dial RPC.** Run keeps its own gVisor netstack
(transparency preserved), but instead of running its own tsnet it
asks the daemon to dial: \"open `gateway:443`, here's the PROXY-v1
header, splice this FD.\" Daemon returns a connected conn via
SCM_RIGHTS or relays bytes.
- Pros: per-run netstack isolation, shared tsnet identity, daemon
  stays small.
- Cons: bytes traverse the Unix socket twice; two layers of TCP
  state to debug.

**D. Leader/follower without a daemon.** First run wins a flock and
becomes leader; subsequent runs hand their TUN fds over the socket.
Leader exit triggers re-election; remaining runs stall while a new
leader does `tsnet.Server.Up`.
- Pros: no separate daemon binary, no systemd unit.
- Cons: every leader exit is a synchronized ~5-10s stall across all
  followers; long-lived in-flight connections die during handoff;
  crash-leader case is messy.

**E. Per-run ephemeral keys (the #413 model).** Fresh authkey per
`run`, fresh `tsnet.Server{Ephemeral: true}` in a temp `Dir`.
- Pros: no daemon, full isolation, identity model parallels WG's
  per-run ephemeral peers.
- Cons: ~5-10s tsnet bring-up on every invocation; transient device
  churn in the admin panel.

**F. Opportunistically reuse the user's `tailscaled` if present.**
If `/var/run/tailscale/tailscaled.sock` is accessible, talk to it
via `LocalClient.Dial` — the daemon role is already filled. Fall
back to spawning our own tsnet otherwise.
- Pros: zero new infrastructure for users who already run Tailscale.
- Cons: two code paths to maintain; sub-identity story (audit logs
  attributing `clawpatrol run` traffic separately from the user's
  personal device) is unclear via the local API.

**G. Don't do tsnet for `run` at all; keep WG.** Gateway runs both
control modes (the dual-mode shape #401 proposes). `clawpatrol run`
still uses wg.conf + per-run ephemeral WG peers, same as today.
- Pros: zero new client code.
- Cons: doesn't solve #401 — still running two layers of WG on
  tailnet-only deployments. macOS works unchanged.

If I had to pick two to prototype: **A** as the long-term answer,
**B** as the \"ship something next week\" answer. B is small enough
that we could land it under a flag and see how many real workloads
break; if HTTP-ish workloads cover 95% of `clawpatrol run` usage,
(A)'s netstack work may never be needed.

**F is also worth an hour of investigation** before committing to
any of the above — if `tailscaled.LocalClient().Dial(ctx, \"tcp\",
\"gateway:443\")` works with the right socket permissions, the
daemon question might just go away for the common case.

piscisaureus:
Closing — superseded by landed work:

* Gateway side (embedded tsnet, advertise `0.0.0.0/0`+`::/0`) landed
  in #443, using `RegisterFallbackTCPHandler` for native original-dst
  instead of the exit-node prefs approach here.
* macOS `clawpatrol run` in tsnet mode landed in #441 (NE extension
  `mode=tailscale` branch).
* Concurrent `clawpatrol run` on Linux fixed in #446 (per-invocation
  `MkdirTemp` + `Ephemeral: true`, addressed #448).

PR #412, opened 2026-05-15
postgres + clickhouse_native: honor agent cancel during approval

arnauorriols:
## Summary

When a SQL query is parked waiting for HITL approval, an agent-side
`Ctrl+C` (psql, clickhouse-client) printed "Cancel request sent" but
the cancel didn't take effect — the query kept waiting on the
approver. This change wires both protocols' cancel paths through to
the approval flow.

### postgres

Cancel arrives on a **separate** TCP connection to port 5432, not on
the existing session. The dispatcher entry now distinguishes
StartupMessage (`196608`), SSLRequest (`80877103`), and CancelRequest
(`80877102`) at the first-8-bytes check. CancelRequest reads
`[pid:4][secret_key:4]` and looks up the live session in a
process-wide registry.

To make the registry safe, the gateway now mints its own
BackendKeyData (the `'K'` post-auth frame) before relaying it to the
agent: a random `(pid, key)` pair the agent uses for cancel, with the
upstream's real `(pid, key)` retained on the registry entry for the
forwarding path.

Dispatch on cancel:

- session parked for approval → abort the approval chain via the new
  `ApproveCallRequest.Cancel` channel → `pgWriteDeny` on the
  original agent connection (`E` + `Z`);
- session running upstream → open a one-shot TCP connection to the
  upstream postgres and forward the CancelRequest with the real
  `(pid, key)`;
- unknown `(pid, key)` → silent drop (mirrors what real postgres
  does).

Precedence: a cancel concurrent with an `allow` verdict still
denies — `pgEvaluateInfo` re-checks the channel after `Approve`
returns and overrides the verdict.

### clickhouse_native

Verification confirmed the bug: the inspector loop synchronously
calls `ch.Approve` from inside `chHandleQuery`, so an inbound
`ClientCancel` byte sits in the kernel socket buffer until approval
resolves. Fixed by spawning a one-shot peek goroutine while the
approver is parked:

- if the byte is `ClientCodeCancel`, close
  `ApproveCallRequest.Cancel` so the approver returns immediately
  and `chHandleQuery` overlays a "canceled by agent" deny that the
  agent sees as a `ServerException`;
- otherwise (typical case: `ClientCodeData` from an INSERT's
  trailing block), the byte is rewound onto a fresh
  `chgoproto.Reader` via the existing `chRewindReader` helper so the
  main pump keeps streaming;
- if no byte arrives by the time the approver resolves,
  `SetReadDeadline` unblocks the peek with no side effects.

### Shared plumbing

- `runtime.ApproveCallRequest` gains a `Cancel <-chan struct{}`
  channel.
- `main.go` derives a cancellable context for `runApproveChain` from
  that channel so existing approvers' `<-ctx.Done()` paths fire
  unchanged.

### Out of scope

clickhouse_https cancellation (HTTP queries aren't cancellable
client-side); operator-driven cancel from the dashboard; idle-timeout
cancel of long-parked approvals.

## Test plan

- [x] `go test ./...` and `go test -race ./...` green
- [x] `gofmt -l .` clean, `go vet ./...` clean
- [x] new unit tests cover: `(pid, key)` registry register / lookup /
      unregister, park / cancel-parked / unpark transitions, unknown
      cancel silently dropped, parked cancel fires the close, not-
      parked cancel forwards upstream with the real key, cancel-
      beats-allow precedence, ClickhouseClientCancel mid-approval
      aborts and is swallowed, non-cancel byte rewinds, no-byte case
      cleans up via SetReadDeadline
- [ ] manual end-to-end with psql against a gateway-fronted postgres
      (out of scope for this PR — requires test infrastructure)

### DIFF
--- a/config/plugins/endpoints/clickhouse_native_runtime.go
+++ b/config/plugins/endpoints/clickhouse_native_runtime.go
@@ -14,6 +14,7 @@ import (
 	"log"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
 	chgoproto "github.com/ClickHouse/ch-go/proto"
@@ -394,7 +395,7 @@ func chAgentToServer(ctx context.Context, ch *runtime.ConnHandle, agentReader *c
 		}
 		switch chgoproto.ClientCode(code) {
 		case chgoproto.ClientCodeQuery:
-			next, fatal := chHandleQuery(ctx, ch, agentReader, upstream, revision, credName, &sessionDB)
+			next, rewound, fatal := chHandleQuery(ctx, ch, agentReader, upstream, revision, credName, &sessionDB)
 			if fatal {
 				return
 			}
@@ -405,6 +406,13 @@ func chAgentToServer(ctx context.Context, ch *runtime.ConnHandle, agentReader *c
 			// us reading a compressed frame as an uncompressed Block,
 			// which surfaces as a "data-block-decode" boolean error.
 			compression = next
+			if rewound != nil {
+				// The approval-cancel peek consumed a byte that
+				// turned out NOT to be a ClientCancel; prepend it
+				// back onto the agent stream so the next code-loop
+				// iteration sees the start of the next packet.
+				agentReader = rewound
+			}
 		case chgoproto.ClientCodeData:
 			rewound, ok := chHandleData(ch, agentReader, upstream, revision, compression)
 			if !ok {
@@ -459,7 +467,7 @@ func chAgentToServer(ctx context.Context, ch *runtime.ConnHandle, agentReader *c
 // nil pointer disables the tracking (used by tests that don't care
 // about session state and the truncated-frame path that has no SQL
 // to inspect).
-func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chgoproto.Reader, upstream io.Writer, revision int, credName string, sessionDB *string) (chgoproto.Compression, bool) {
+func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chgoproto.Reader, upstream io.Writer, revision int, credName string, sessionDB *string) (chgoproto.Compression, *chgoproto.Reader, bool) {
 	// Build the forward-bound preamble (everything up to the body
 	// length prefix) as we decode each field. Re-encoding these small
 	// fields is cheap; the body is the only field whose size warrants
@@ -470,7 +478,7 @@ func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chg
 	id, err := agentReader.Str()
 	if err != nil {
 		chEmitError(ch, "query-decode", "id: "+err.Error())
-		return chgoproto.CompressionDisabled, true
+		return chgoproto.CompressionDisabled, nil, true
 	}
 	preamble.PutString(id)
 
@@ -478,20 +486,20 @@ func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chg
 		var info chgoproto.ClientInfo
 		if err := info.DecodeAware(agentReader, revision); err != nil {
 			chEmitError(ch, "query-decode", "client info: "+err.Error())
-			return chgoproto.CompressionDisabled, true
+			return chgoproto.CompressionDisabled, nil, true
 		}
 		info.EncodeAware(&preamble, revision)
 	}
 
 	if !chgoproto.FeatureSettingsSerializedAsStrings.In(revision) {
 		chEmitError(ch, "query-decode", "unsupported settings format")
-		return chgoproto.CompressionDisabled, true
+		return chgoproto.CompressionDisabled, nil, true
 	}
 	for {
 		var s chgoproto.Setting
 		if err := s.Decode(agentReader); err != nil {
 			chEmitError(ch, "query-decode", "setting: "+err.Error())
-			return chgoproto.CompressionDisabled, true
+			return chgoproto.CompressionDisabled, nil, true
 		}
 		if s.Key == "" {
 			preamble.PutString("") // end-of-settings sentinel
@@ -504,14 +512,14 @@ func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chg
 		secret, err := agentReader.Str()
 		if err != nil {
 			chEmitError(ch, "query-decode", "inter-server secret: "+err.Error())
-			return chgoproto.CompressionDisabled, true
+			return chgoproto.CompressionDisabled, nil, true
 		}
 		preamble.PutString(secret)
 	}
 
 	if _, err := agentReader.UVarInt(); err != nil {
 		chEmitError(ch, "query-decode", "stage: "+err.Error())
-		return chgoproto.CompressionDisabled, true
+		return chgoproto.CompressionDisabled, nil, true
 	}
 	// chgoproto.Query.EncodeAware hard-codes StageComplete on the wire
 	// regardless of what the agent sent; mirror that here so the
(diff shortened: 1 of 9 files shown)


config/plugins/endpoints/clickhouse_native_runtime.go
```diff
@@ -547,30 +555,46 @@ func chHandleQuery(ctx context.Context, ch *runtime.ConnHandle, agentReader *chg
...
+	// read.
+	approveSt := &chApproveCancelState{agentReader: agentReader, ch: ch}
+	verdict, reason, nextDB := chEvaluateSQL(ctx, ch, string(head), credName, database, truncated, approveSt.run)
+	rewound := approveSt.rewound
+	if approveSt.canceled {
+		// The agent issued a ClientCancel while we were parked. Force
+		// the verdict to deny so the agent sees a terminating
+		// Exception (and the pump doesn't try to ship Data blocks
+		// that follow the canceled Query to upstream).
+		verdict, reason = "deny", "canceled by agent"
```
arnauorriols: We shouldn't return a synthetic verdict that could be mistaken by a hitl decision. Find an alternative solution that results in the request being properly cancelled instead of returning a fake deny exception

config/plugins/endpoints/clickhouse_native_runtime.go
```diff
@@ -1010,6 +1037,105 @@ func chEmit(ch *runtime.ConnHandle, ev runtime.ConnEvent) {
...
+	}
+	req.Cancel = cancelCh
+
+	type peekResult struct {
+		b   byte
+		err error
+	}
+	peekDone := make(chan peekResult, 1)
+	go func() {
+		b, err := s.agentReader.UInt8()
```
arnauorriols, line 1158: The peek goroutine here consumes one byte off `agentReader` during approval. For any modern protocol revision the Query packet ends with a Parameters block AFTER the body (`FeatureParameters`, rev ≥ 54459 — `chMaxProtocolRev = chgoproto.Version` covers this), and the agent flushes the whole Query packet before waiting on a response. So in the common case (no Ctrl+C), the peek reads the leading byte of the trailing Parameters block — typically the `0` sentinel for an empty parameter list, or the leading UVarInt of the first param's Key.

That byte then lands in `s.rewound` for the *next* chAgentToServer iteration, but `chHandleQuery` continues reading the Parameters block from the same `agentReader` it just consumed from — both the deny-drain path (`p.Decode(agentReader)` around L596) and the allow-forward path (around L628). `p.Decode` reads from a stream missing its first byte: either it errors out (chEmitError + fatal=true → session torn down post-approval), or it succeeds while interpreting garbage as a Parameter and upstream rejects later.

The `truncated == true` path is also affected: the body tail is still in the buffer right after `head`, and the peeked byte is the first byte of that tail. Same off-by-one when the deny path runs `io.CopyN(io.Discard, agentReader, bodyLen-headLen)`.

The existing tests miss this because `state.run` is exercised against an empty pipe / a single synthetic byte, never against an agentReader that still has packet-trailing bytes pending.

A few directions:
- Drain Parameters (and the truncated body tail) BEFORE invoking approval so the peek genuinely lives between packets.
- Wrap the underlying net.Conn in `bufio.Reader` and use `Peek(1)` so the byte isn't consumed (chgoproto.Reader doesn't expose peek itself, but the conn is reachable via `ch.Conn`).
- Skip the peek on the allow side entirely and only deploy it during a long operator-wait window; the rewind machinery isn't useful on the cancel path either (the cancel byte is swallowed).

config/plugins/endpoints/postgres.go
```diff
@@ -668,13 +702,30 @@ func pgEvaluateInfo(ch *runtime.ConnHandle, info pgInfo, credName, database stri
...
+			case <-parkCancel:
+				canceled = true
+			default:
+			}
+		}
+		if canceled || v.Decision != "allow" {
 			reason := v.Reason
-			if reason == "" {
+			if canceled {
+				reason = "canceled by agent"
```
arnauorriols: Same critique as the inline above on clickhouse_native: on agent cancel this synthesizes `verdict = "deny"` + `Action: "denied"` with reason `"canceled by agent"`. From the dashboard / metrics surface, an agent Ctrl+C and a HITL operator deny become indistinguishable — both fire as `denied` ConnEvents with structurally identical payloads, polluting deny-rate / SLO metrics and making the dashboard's deny-reason filter noisy.

Suggested direction: introduce a distinct verdict (e.g. `"canceled"` or `"aborted"`) and matching `Action` so the dashboard can render agent cancels separately from rule denies. Same change applies to chHandleQuery's `verdict, reason = "deny", "canceled by agent"` overlay.

arnauorriols (commented):
Two follow-ups in addition to the synthetic-deny note:

1. **clickhouse_native peek consumes the Parameters block byte** — concrete bug, not just a smell. The in-flight peek during approval reads one byte off `agentReader`, but for any protocol revision ≥ 54459 (`FeatureParameters`, well below the gateway's `chMaxProtocolRev = chgoproto.Version`) the Query packet's trailing Parameters block is already buffered. The peeked byte is sent into `s.rewound` for a future iteration, while `chHandleQuery` continues reading parameters from the now-short stream — either erroring out (fatal=true, session torn down post-approval) or accepting garbage. Truncated-body path is affected identically.

2. **postgres has the same cancel→synthetic-deny pattern** — `pgEvaluateInfo` translates an agent CancelRequest into `verdict="deny"` + `Action: "denied"` with reason `"canceled by agent"`. Identical dashboard/metrics conflation with HITL operator denies. Same fix recommended (distinct verdict + action for agent cancels).

arnauorriols:
[bot-reply]
Pushed 0e746bd addressing this comment:
- postgres: agent cancel during HITL now emits `Action: "canceled"` (Approver fields cleared) and `pgEvaluate` returns a `"canceled"` verdict; the caller still writes a deny ErrorResponse on the wire.
- clickhouse_native: `chApproveCancelState.run` substitutes `Decision: "canceled"` when the in-band ClientCancel byte fires, so `chEvaluateSQL` emits `Action: "canceled"` (instead of `hitl_deny`) and returns `verdict="canceled"`. The `chHandleQuery` overlay was removed; wire path treats `deny` and `canceled` the same.
- `runtime.ConnEvent.Action` doc updated to include `"canceled"`.

config/plugins/endpoints/clickhouse_native_runtime.go
arnauorriols: [bot-reply]
Pushed 314bc56 addressing this comment:
- Drain the Parameters block off the wire BEFORE invoking approval, so the peek goroutine in chApproveCancelState.run operates at a true packet boundary instead of consuming the leading byte of Parameters / the truncated body tail.
- Truncated queries (bodyLen > chMaxQueryBody) leave approveSt.agentReader unset → state.run falls through to plain ch.Approve. We lose in-band cancel detection for that one query, but the old peek was already off-by-one on this path so it's a net improvement; buffering a multi-MiB tail isn't safe.
- Allow path forwards the pre-drained paramBuf directly; truncated path keeps the inline decode + re-encode.
- New tests in clickhouse_native_test.go: forward-with-Parameters round-trip, deny-with-Parameters wire-alignment regression (uses a 3-char param key so its UVarInt length byte = 0x03 / ClientCodeCancel and would have tripped the previous false-cancel), and HITL-allow-with-Parameters confirming the peek no longer mis-fires.

arnauorriols:
[bot-reply]
No action taken — the PR body is a self-authored description of the completed change (Summary + Test plan + Out of scope) with no open question or change request directed at the bot. The one unchecked test-plan item (manual end-to-end with psql) is explicitly marked out of scope. Nothing to implement or reply to here.

piscisaureus:
Closing during the backlog pass: pre-launch branch, +1366 lines, no activity since May, and the HITL paths it touches (async operations, retry relay) have been rewritten since. Honouring an agent-side cancel while a SQL query waits for approval is still a good behaviour to have; a fresh PR against current main, ideally smaller, is welcome.

PR #413, opened 2026-05-15
feat: Tailscale exit-node MITM via SO_ORIGINAL_DST + clawpatrol run in tsnet mode

littledivy:
### clawpatrol run in Tailscale mode
- Client: new user+net+mnt namespace + TUN bridged to gVisor netstack; each TCP connection dials the gateway via ephemeral `tsnet.Server{Ephemeral: true}`; HAProxy PROXY v1 header carries original 4-tuple so gateway routes all ports (443→HTTPS MITM, 5432→PG, ClickHouse native, etc.)
- Gateway accept loop: peeks for PROXY header → dispatches by original dst IP/port (same table as WG forwarder); no header → `g.handle` fallback (direct dashboard/admin, backward-compat)
- API: `POST /api/peer/ephemeral/tsnet` mints ephemeral Tailscale auth keys; `mintTailscaleAuthKey` gains `ephemeral bool` param
- Join: writes `mode`/`tailnet-gateway`/`control-url` marker files; poll response includes `gateway_host`/`control_url` for Tailscale joins
- macOS: stub with clear error (NE extension + tsnet is a separate task)

### Exit-node MITM via SO_ORIGINAL_DST (Linux)
- When the gateway runs as a Tailscale exit node, iptables `PREROUTING REDIRECT` on `tailscale0` intercepts TCP on ports 443/5432 + all ConnRouter ports → gateway listen port; `SO_ORIGINAL_DST` recovers the pre-NAT destination so the accept loop can dispatch normally
- UDP FORWARD REJECT on same ports forces QUIC→TCP fallback (QUIC can't be REDIRECT-ed by iptables)
- `installExitNodeRedirect` also called on policy reload so newly-added ConnRouter ports (e.g. adding a ClickHouse native endpoint) pick up iptables rules without a service restart
- Exempt public-IP inbound: exit-node routing skipped for connections arriving on the public IP (non-tailscale interface), preventing NAT loops
- Split DNS: writes systemd-resolved override so endpoint hostnames resolve inside the gVisor netstack on the client side

### Auth / routing fixes
- `skipsTailnetGate` now returns `true` for `authSelfAuthenticating` routes (e.g. `/api/cred/` credential webhooks are HMAC-verified, no Tailscale identity needed); previously only `authPublic` routes bypassed the tailnet gate
- `isTailscaleControlMode("")` returns `true` (default mode is Tailscale, not WireGuard); WG-mode openListener tests updated to set `Control: "wireguard"` explicitly

## Known limits
- IPv4 only: `SO_ORIGINAL_DST` uses `RawSockaddrInet4`; no `ip6tables` rules installed
- Requires `iptables` binary at runtime (graceful skip on nftables-only systems like Debian 12+ without the iptables compatibility shim)

## Test plan
- [ ] `clawpatrol join <tailscale-gateway>` → `~/.config/clawpatrol/mode` = `tailscale`, `tailnet-gateway` written
- [ ] `clawpatrol run claude` → HTTPS traffic intercepted via MITM
- [ ] `clawpatrol run psql -h <pg-host>` → PostgreSQL traffic routed through gateway
- [ ] Direct browser → gateway dashboard → no PROXY header → `g.handle` fallback still works
- [ ] `clawpatrol run` in WG mode → unchanged (loopback listener, no PROXY headers)
- [ ] Policy reload with new ConnRouter endpoint → `iptables -t nat -L PREROUTING` shows new port

### DIFF
--- a/origdst_linux.go
+++ b/origdst_linux.go
@@ -1,0 +1,108 @@
+//go:build linux
+
+package main
+
+import (
+	"encoding/binary"
+	"log"
+	"net"
+	"os"
+	"os/exec"
+	"strconv"
+	"syscall"
+	"unsafe"
+)
+
+const soOriginalDst = 80 // SO_ORIGINAL_DST, linux/in.h
+
+// originalDst returns the pre-NAT destination of c when the connection
+// was redirected by an iptables REDIRECT rule. Returns ok=false on
+// non-Linux, for non-TCP conns, or when no NAT entry exists (direct
+// connection — getsockopt returns ENOPROTOOPT or ENOENT).
+func originalDst(c net.Conn) (ip string, port uint16, ok bool) {
+	tc := unwrapTCPConn(c)
+	if tc == nil {
+		return "", 0, false
+	}
+	raw, err := tc.SyscallConn()
+	if err != nil {
+		return "", 0, false
+	}
+	var sa syscall.RawSockaddrInet4
+	saLen := uint32(syscall.SizeofSockaddrInet4)
+	var soErr error
+	_ = raw.Control(func(fd uintptr) {
+		_, _, errno := syscall.Syscall6(
+			syscall.SYS_GETSOCKOPT,
+			fd,
+			syscall.IPPROTO_IP,
+			soOriginalDst,
+			uintptr(unsafe.Pointer(&sa)),
+			uintptr(unsafe.Pointer(&saLen)),
+			0,
+		)
+		if errno != 0 {
+			soErr = errno
+		}
+	})
+	if soErr != nil {
+		return "", 0, false
+	}
+	dstIP := net.IP(sa.Addr[:]).String()
+	dstPort := binary.BigEndian.Uint16((*[2]byte)(unsafe.Pointer(&sa.Port))[:])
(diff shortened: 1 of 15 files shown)


ry:
ref #408 
closes https://github.com/denoland/clawpatrol/issues/401

origdst_linux.go
```diff
@@ -0,0 +1,128 @@
...
+// installExitNodeRedirect installs iptables PREROUTING REDIRECT rules so that
+// traffic arriving via the Tailscale exit-node (tailscale0) on common service
+// ports is redirected to the gateway's own listen port. This lets the accept
+// loop's SO_ORIGINAL_DST fallback dispatch exit-node traffic exactly like the
+// WireGuard promiscuous forwarder — no PROXY header needed.
+//
+// Idempotent: each rule is checked with -C before insertion. Failures are
+// logged but not fatal; operators can install the rules manually if the
+// process lacks CAP_NET_ADMIN.
+func installExitNodeRedirect(listenPort int, extraPorts []string) {
```
littledivy, line 65: https://github.com/tailscale/tailscale/issues/10940

PR #415, opened 2026-05-15
hitl: add async operation store

magurotuna:
## Summary
- Adds the async HITL `hitl_operations` SQLite migration for durable metadata-only operation tracking.
- Adds `HITLOperationStore` APIs for create, owner-scoped lookup, compare-and-swap state transitions, and one-shot retry-grant consumption.
- Covers non-enumerating owner checks, unknown-state rejection, stale transition conflicts, retry mismatch/expiry safety, and the regression that generic transitions cannot consume retry grants.

This is stacked on #414 and intentionally stays schema/store-only. It does not add polling endpoints, raw-body HMAC computation, retry header handling, Slack/dashboard state, or upstream relay integration.

## Security notes
- The operation table persists metadata and fingerprints only.
- It does not store raw request bodies, `Authorization`, `Cookie`, upstream secrets, request headers, replay bodies, or other replayable request material.
- `ConsumeRetryGrant` is the only path that marks a retry grant consumed, and it checks profile/principal ownership, state, expiry, auth binding, fingerprint version/key, and request fingerprint before atomically moving to upstream execution.

## Test plan
- `npm --prefix www run build`
- `gofmt -l hitl_operation_store.go hitl_operation_store_test.go`
- `git diff --check`
- static added-line scan for hardcoded secrets / replayable-material storage patterns
- `go test . -run 'TestHITLOperationStore|TestHITLOperationTransitionCannotConsumeRetryGrant' -count=1`
- `go test ./...`

## Review
- Independent follow-up review after the retry-consumption blocker fix: `APPROVED_NO_FINDINGS`.

### DIFF
--- a/hitl_operation_store.go
+++ b/hitl_operation_store.go
@@ -1,0 +1,422 @@
+package main
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+)
+
+type HITLOperationState string
+
+const (
+	HITLOperationStateSyncWaiting             HITLOperationState = "sync_waiting"
+	HITLOperationStatePendingApproval         HITLOperationState = "pending_approval"
+	HITLOperationStateApprovedWaitingForRetry HITLOperationState = "approved_waiting_for_retry"
+	HITLOperationStateDenied                  HITLOperationState = "denied"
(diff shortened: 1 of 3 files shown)


magurotuna:
Replacement PR opened: #419 (based directly on `main`). This replaces #415, which was closed after its stacked base branch was deleted.

PR #433, opened 2026-05-16
docs(glossary): merge Facet/Family entries; add Facet field

arnauorriols:
## Summary

The previous glossary treated **Facet** and **Family** as distinct
concepts. They are synonyms: both name the protocol class an action
belongs to (`http`, `sql`, `ssh`, `k8s`). The matchable properties a
rule's CEL condition reads are **facet fields** (alias: family fields).

## Changes (`site/doc/glossary.md`)

- Rewrote **Facet** to describe the protocol class, with **Family**
  called out as the alias. Lists the built-in facets.
- Added **Facet field** (alias: **Family field**) covering the
  per-facet field sets, with examples for `sql` (including
  `sql.database`), `http`, and `k8s`.
- Updated the **Rule** and **CEL condition** cross-references so links
  to matchable properties resolve to `#facet-field` rather than
  `#facet`.

Build: `cd site && npm run build` is clean.

## Follow-ups (not in this PR)

The same Facet-vs-Family-vs-facet-field framing isn't reflected in
`site/doc/approval-rules.md` or `config/README.md` — they still mix
"facet" and "family" loosely. Worth a separate pass.

### DIFF
--- a/site/doc/glossary.md
+++ b/site/doc/glossary.md
@@ -72,7 +72,7 @@ One unit of agent work the gateway sees and applies policy to — one
 HTTP call, one SQL query, one `kubectl` invocation, one SSH command.
 Each action targets an [endpoint](#endpoint), is gated by the matching
 [rule](#rule)'s [outcome](#outcome), and surfaces in the dashboard's
-live request feed (action [families](#facet): `http`, `sql`, `k8s`,
+live request feed (action [families](#family): `http`, `sql`, `k8s`,
 `ssh`) with its own detail page. "Action" is the operator-visible concept of "the
 thing the agent did."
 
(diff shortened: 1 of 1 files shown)


site/doc/glossary.md
arnauorriols: These "record kinds" are actually "action families", and should be referenced as such

arnauorriols (commented):
Families and facets are not really the same:
- an action belongs to an action family. 
- each action family includes one or more facets into the action's data. The sql family includes the sql facet. The http family includes the http facet. The k8s family includes the http and k8s facets. 
- each facet is a collection of fields: sql facet includes sql.verb, sql.functions, sql.tables, and sql.statment.

site/doc/glossary.md
arnauorriols: [bot-reply]
Pushed 3d60b1c addressing this comment:
- Renamed "record kinds" to "action [families](#facet)" in the Action entry, linking to the existing Facet glossary entry (which already documents "Family" as an alias).

arnauorriols:
Families and facets are not really the same:

an action belongs to an action family.
each action family includes one or more facets into the action's data. The sql family includes the sql facet. The http family includes the http facet. The k8s family includes the http and k8s facets.
each facet is a collection of fields: sql facet includes sql.verb, sql.functions, sql.tables, and sql.statment.

arnauorriols:
[bot-reply]
Pushed 40606b5 addressing this comment:
- Split the merged Facet entry into a Family entry (the protocol class an action belongs to) and a Facet entry (a named collection of fields a Family includes into the action's data).
- Documented that the `sql` Family includes the `sql` Facet, the `http` Family includes the `http` Facet, and the `k8s` Family includes both the `http` and `k8s` Facets.
- Dropped the "alias: Family" framing on Facet / Facet field; repointed cross-refs in Action, Rule, and CEL condition to Family where the protocol-class inference is what's described.

arnauorriols:
Don't like the families parenthesis in the action definition. Thread the explanation better. 
Also, update the rule definition. The concept of rule families is wrong

arnauorriols:
[bot-reply]
Pushed 03eca1e addressing this comment:
- Action: dropped the parenthetical "(action families: http, sql, k8s, ssh)". Family is now threaded into the definition: every action belongs to exactly one family — the protocol class (`http`, `sql`, `ssh`, `k8s`) the gateway used to intercept it — and that family fixes the data the gateway extracts from the wire and exposes to policy.
- Rule: removed the "rule's family" framing. The CEL condition now matches the facet fields *the rule's endpoints' family* exposes; the load-time constraint is reframed as "all endpoints listed by one rule must share a family" rather than "the rule has a family inferred from its endpoints".
- Family: an endpoint's type carries an implied family, which propagates to rules through the shared-family constraint on the endpoint list. Dropped the "A rule's family is inferred from its endpoint(s)" sentence.
- CEL condition: "facet fields exposed by the family of the rule's endpoints" instead of "the rule's inferred family exposes".

site/doc/glossary.md
arnauorriols: [bot-reply]
This was already addressed before the PR was squash-merged into main (a45e013). The merged `site/doc/glossary.md` uses "family"/"families" throughout — e.g. line 73: "Every action belongs to exactly one [family](#family) — the protocol class (`http`, `sql`, `ssh`, `k8s`)…" — and no "record kind" wording remains. No further commits needed.

PR #461, opened 2026-05-19
Add insecure_no_dashboard_auth flag for local dev

ry:
## Summary

Adds an `insecure_no_dashboard_auth = true` config field that disables dashboard authentication entirely. Loopback-bind only — gateway refuses to start otherwise.

After #454 made the root password mandatory in WireGuard control mode, the dev loop for `clawpatrol gateway dev.hcl` got noticeably worse: launch gateway → open browser → set a password → dismiss the form → run the actual test. With this flag the loop is: launch gateway → run the test.

## Safety

- Refuses to start unless `info_listen` is `127.0.0.1` / `::1` / `127.x.y.z` — can't accidentally enable on a real deploy.
- Startup log is loud: `dashboard auth: DISABLED via insecure_no_dashboard_auth — every request gets root-level access. Never use this in production.`
- The `insecure_` name prefix and `Never use this in production` doc comment make it unmistakable.

## Implementation

- `internal/config/config.go`: new `InsecureNoDashboardAuth bool` field on `Gateway`.
- `cmd/clawpatrol/web.go`: bypass at the top of `dashboardAuthGate` — injects the dashboard-password principal so downstream tailnetGate short-circuits.
- `cmd/clawpatrol/main.go`: startup safety check before `logDashboardAuthState`; refuses non-loopback bind with a `log.Fatalf`.
- `site/doc/config-reference.md`: regenerated by `go run ./internal/tools/docgen`.

26 insertions, 1 deletion.

## Test plan
- [x] `make test` passes (including docgen freshness check)
- [x] gateway starts with `insecure_no_dashboard_auth = true` and prints the DISABLED warning
- [x] API requests bypass auth (no 401)
- [x] gateway refuses to start when `info_listen` is `0.0.0.0:8080` with the flag set
- [ ] CI green


piscisaureus (approved):
You can also just run `clawpatrol gateway --set-dashboard-password bla`

PR #479, opened 2026-05-19
Design proposal: credentials and profiles to the database

ry:
Doc-only PR. Proposes moving credentials and profiles out of the HCL config file and into the database, edited via the dashboard.

## The principle

> When adding a new agent, I'm not changing the rules, therefore I shouldn't have to edit the config.

## What changes (proposed, not implemented in this PR)

- HCL keeps: rules, approvers, endpoints, tunnels, plugins, gateway operational config
- DB owns: credentials (declaration + secret), profiles (identity → endpoint+credential pairs)
- Profile schema becomes `(profile, endpoint, credential)` triples, so endpoint declarations no longer carry a `credential` field. That collapses the per-person endpoint duplication in `gateway.hcl` (four `anthropic-*` endpoints become one `anthropic`).

## Phasing

1. Profiles to DB (schema unchanged) — minimum-friction win
2. Credentials to DB
3. Schema change to `profile_endpoints` with pairs
4. Reject HCL `credential` / `profile` blocks

## Review request

@alice @Operator — this captures the Slack thread plus the per-person endpoint duplication smell I noticed in `gateway.hcl`. Looking for:

- Sanity check on the HCL-vs-DB cut (does anything I'm leaving in HCL actually belong in DB? does anything I'm moving to DB carry infrastructure semantics I'm missing?)
- Whether the four-phase sequencing is the right granularity, or if some phases should collapse
- The "two sources of truth during migration" trade-off and how aggressively to push toward Phase 4

Nothing here is committed to implementation yet. PR is just to get the design under version control and review-able.

### DIFF
--- a/doc/credentials-profiles-to-db.md
+++ b/doc/credentials-profiles-to-db.md
@@ -1,0 +1,304 @@
+# Credentials and profiles to the database (design proposal)
+
+**Design principle.** Adding a new agent should not require editing HCL.
+
+Today every personnel change (new team member, new bot, granting an
+existing user access to one more service) ends with a git commit and a
+gateway restart. That makes the file a record of every Slack handle on
+the team, scales config noise with headcount instead of policy
+complexity, and locks personnel changes behind code review even though
+nothing about the policy itself is changing.
+
+The proposal: keep rules, approvers, endpoints, and gateway operational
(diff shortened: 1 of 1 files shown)


ry:
rejected

PR #489, opened 2026-05-19
toolgate: exploratory LLM tool-call gating (draft)

arnauorriols:
## Summary

Exploratory implementation of **tool-call gating** — rules applied to a model's `tool_use` blocks before the agent ever sees them. Three verdicts: allow, deny, HITL (human-in-the-loop). For HITL, the gateway parks the call and runs a follow-up LLM call so the **model picks a polling tool from the agent's own advertised `tools[]`**; the agent executes that tool to poll clawpatrol's approval endpoint and waits for an operator decision.

**Marked draft. Not for merge.** Bead: cl-umxf. Design doc: `doc/tool-call-gating.md`.

> **Body refreshed 2026-06-18** to match the implementation as it stands after review. Streaming (SSE) gating and the gateway-initiated polling-tool choice ("option 2") have landed since the first draft and are no longer stubbed; the design decisions and out-of-scope list below reflect current `HEAD`. See the PR thread for the review history.

## What's implemented

- `internal/toolgate` — standalone package, independently testable
  - `Rule` / `RuleSet` matcher (tool name + args substring; first match wins)
  - `Store` — in-memory token-keyed parked-call registry with background sweep (`StartSweeper`)
  - `GateAnthropicResponse` — buffered-JSON path: parses `content[]`, rewrites `tool_use` blocks per verdict
  - `GateAnthropicSSE` (`anthropic_sse.go`) — **streaming path**: per-block buffering, holds a `tool_use`'s frames from `content_block_start` to `content_block_stop`, evaluates at stop, then replays / rewrites / parks. Non-`tool_use` frames stream straight through (time-to-first-token preserved). **Fail closed** — a `tool_use` that can't be evaluated is blocked or the stream is terminated; the raw call never reaches the agent.
  - `followup.go` — **gateway-initiated polling-tool choice** ("option 2"): on HITL, removes the parked `tool_use`, fabricates a user message telling the model to poll the approval endpoint, makes clawpatrol's own `/v1/messages` call (via an injected `LLMCaller`), and forwards the model's chosen tool to the agent. Degrades to an "approval pending" text block if no caller is wired or the call fails.
  - Three approval endpoints: `POST /api/approval/poll` (real, long-poll, 30s default), `GET /api/approval/sse` (wired, single-event), `GET /api/approval/ws` (stub — returns 501 with fallback hint)
  - Dashboard-side `POST /api/approval/decide`, `GET /api/approval/pending`
- `cmd/clawpatrol/toolgate.go` — MITM hook glue (body buffer, gunzip, swap, dashboard event marker), wires the real `LLMCaller` for the follow-up dance
- MITM integration in `mitmHTTPS`: fires on `api.anthropic.com` `/v1/messages`, both buffered-JSON and streaming-SSE responses, behind a rule-set gate
- Rules loaded from `CLAWPATROL_TOOLGATE_RULES` (JSON env var). Empty = gating is a no-op. Approval base URL from `CLAWPATROL_TOOLGATE_APPROVAL_URL`.
- Tests (`go test ./internal/toolgate/` — green):
  - allow / deny rewrite paths over fixture (buffered + SSE)
  - HITL multi-turn dance E2E through the live HTTP endpoint (park → follow-up picks tool → agent polls → dashboard approves → poller wakes)
  - HITL deny half of the same dance; long-poll timeout returns `pending`
  - SSE allow-replays-verbatim, deny-rewrites, HITL-parks-and-emits-model-choice
  - SSE fail-closed cases: malformed delta, bad block-start, truncated mid-tool_use
  - Follow-up unavailable / error fallbacks; `BuildFollowupRequest`
  - Store sweep GCs aged decided entries

## Design decisions

1. **Standalone package not glued to cl-1yh.** cl-1yh's `llm` facet + `llm_rule` plugin isn't on `main`. Follow-up: when cl-1yh lands, replace `toolgate.Rule` with a thin shim consuming the compiled `llm_rule`. Until then, `CLAWPATROL_TOOLGATE_RULES` is the configuration surface.
2. **The model picks the polling tool, not the gateway.** On HITL the gateway runs a follow-up LLM call (acting as one agent-loop iteration) so the model selects a polling tool from the agent's *own* advertised `tools[]` — the agent's dispatcher is built from the tools it registered, so a tool clawpatrol invents has no handler. (This replaces the first draft's unilateral, unexecutable `clawpatrol_poll` synthesis.)
3. **Both response shapes handled.** Buffered JSON (`GateAnthropicResponse`) and streaming SSE (`GateAnthropicSSE`). Streaming is what makes the gate fire against typical agents, which stream. The streaming gate is fail-closed.
4. **One `tool_use` per turn is the supported shape.** N>1 `tool_use`s are evaluated individually; mixed allow/deny/HITL across siblings is the studied gap — a HITL block's follow-up replaces the whole assistant turn, so siblings aren't independently forwarded. Flagged as a v2 follow-up.
5. **Deny replaces the `tool_use` with a text block carrying the reason** (and flips `stop_reason` to `end_turn`), instead of synthesising a `tool_result`. Simpler — no fabricated round-trip. Synth-`tool_result` (round-trip preservation) is the v2 alternative.
6. **Gateway-initiated LLM follow-up is implemented.** clawpatrol makes its own upstream `/v1/messages` call for the HITL polling-tool choice, reusing the agent's credentials and tool set. Anthropic uses header credential injection (not body signing), so swapping the request body is safe; the follow-up never re-enters the MITM handler, so it isn't re-gated.
7. **Approval surface is independent of the existing HITL chain.** Tokens are gateway-issued; operators decide through `/api/approval/decide` behind the same dashboard auth gate as `/api/hitl/decide`. cl-99t's `rule.template` integration is out of scope for the draft.
8. **Tokens are 32-byte hex from crypto/rand**, treated as bearer credentials on the polling endpoint, auto-GC'd 5 minutes after a verdict.

## What's stubbed / out of scope

- SSE approval endpoint (`/api/approval/sse`): wired (single-event flush, pending-then-verdict), not production-grade (no retry hint, no last-event-id).
- WebSocket approval endpoint (`/api/approval/ws`): returns 501 with a fallback hint pointing at the long-poll path.
- Per-provider parsers: Anthropic only. OpenAI/Codex/OpenRouter are explicit follow-ups.
- N>1 `tool_use`s per turn with mixed verdicts — individually evaluated; sibling independence unstudied.
- Re-gating the follow-up's chosen polling tool — currently trusted, not re-evaluated.
- Dashboard UI: no React work in this PR. Operators interact via `curl POST /api/approval/decide`; the existing HITL surface can render `GET /api/approval/pending` in a follow-up.
- HCL config for rules. Env-var loader is the draft's stand-in.

## Follow-ups (filed as new beads)

- Wire `toolgate.Rule` to cl-1yh's `llm_rule` HCL plugin once cl-1yh lands.
- Independent handling of allow/deny siblings in a HITL turn (N>1).
- Re-gate (or explicitly trust) the follow-up's chosen polling tool.
- Synth-`tool_result` deny path for round-trip preservation.
- SSE / WS production-grade approval endpoints.
- Dashboard UI for the approval queue.
- Per-provider parsers.

## Test plan

- [x] `go test ./internal/toolgate/ -count=1` — passes (buffered + streaming + follow-up + E2E)
- [x] `go test ./cmd/clawpatrol/ -count=1` — passes (no regressions)
- [x] `go test ./... -count=1 -short` — passes
- [x] `gofmt -l .` — clean
- [x] `go vet ./...` — clean
- [x] `cd dashboard && deno task format:check` — clean
- [x] Manual E2E: real Claude subscription creds through the real MITM + WireGuard pipeline — all three verdicts (allow / deny / HITL) observed working (see PR thread).


arnauorriols:
**Pre-test review notes**

I'm setting up an end-to-end test of this and read through the implementation first. A few things worth capturing for the v2 follow-ups — none are merge-blockers given the draft status, but a couple affect whether the feature does anything against real agents.

**1. Streaming bypass means real agents are effectively never gated.** The trigger at `cmd/clawpatrol/main.go:2393` requires `Content-Type: */json` and explicitly excludes `event-stream`. Claude Code, the Anthropic SDKs, and openclaw all stream `/v1/messages` by default, so in a realistic deployment the gate never fires — it only triggers on `"stream": false` responses. This is noted as a known limitation, but it's worth flagging that it isn't a corner case: it's essentially all real traffic. Any manual test has to force non-streaming, and the streaming rewrite is the thing standing between this and being usable.

**2. `Store.Sweep()` is never wired up.** `internal/toolgate/store.go` defines the GC sweep and it's unit-tested, but nothing in the gateway calls it on a tick (compare `agents.startSessionSweeper`, invoked at `main.go:2873`). In a long-lived gateway, decided `PendingCall`s accumulate in the store map forever — a slow leak. A ticker next to the existing session sweeper would close it.

**3. Deny: doc vs. behavior mismatch.** The package doc and the `Verdict` comment in `toolgate.go` say deny "synthesises a `tool_result` the model sees next turn," but `GateAnthropicResponse` replaces the `tool_use` with a **text block** and flips `stop_reason` to `end_turn` (design note #5 in the PR body confirms text-block is the intended draft behavior). The doc comments describe the not-taken design — worth aligning so a reader doesn't expect a synthesized `tool_result` on the wire.

**4. `docs/tool-call-gating.md` doesn't exist.** `PollingToolName` and several design-decision comments reference it for the long-poll rationale and the v2 plan, but the file isn't in the PR.

**5. Minor: `toolgate.Mux(...)` is rebuilt per request** inside each `apiApproval*` wrapper in `web.go` — a fresh `http.ServeMux` per call. Harmless at current volume; cache one on the gateway if the approval surface ever gets hot.

E2E results to follow.

arnauorriols:
[bot-reply]
Pushed d2a80ca addressing this review:

- **(2) `Store.Sweep()` now wired.** Added `Store.StartSweeper`, invoked from `newGateway` next to `startSessionSweeper` (1-min tick, first sweep 30s after boot). Decided `PendingCall`s no longer accumulate forever.
- **(3) Deny doc/behavior mismatch fixed.** The package doc and the `Verdict` comment now describe the actual draft behavior — `tool_use` → text block + `stop_reason: end_turn` — and note the synth-`tool_result` design as the deferred v2 alternative.
- **(4) `doc/tool-call-gating.md` added** (the dir is `doc/`, not `docs/`); the dangling reference in `anthropic.go` now points to it. The doc captures the long-poll-only rationale, the deny-shape tradeoff, and the v2 plan.
- **(5) `toolgate.Mux(...)` cached.** Built once on the gateway (`g.toolgateAPI`); the `apiApproval*` wrappers dispatch through a shared `serveToolgateAPI` helper instead of constructing a fresh `ServeMux` per request.

Not addressed here:

- **(1) Streaming bypass.** Left as the v2 streaming-response rewrite — it's the largest item and the draft is explicitly non-streaming-only. Documented in `doc/tool-call-gating.md` as the gap between this and a feature that fires against real agents. Happy to file/scope it as a dedicated follow-up.

`go build ./...`, `go vet`, `gofmt -l .`, and `go test ./internal/toolgate/... ./cmd/clawpatrol/...` all pass.

arnauorriols:
**E2E results — real Claude subscription creds through the real MITM + WireGuard pipeline**

Stood up the gateway from this branch and ran a real client through it: `clawpatrol run -- curl ...` → WireGuard netns (joined as `10.55.0.2`) → MITM listener → real `anthropic_oauth_subscription` token injected → real `api.anthropic.com` → toolgate rewrite. All three verdicts confirmed against genuine model output:

| Case | Request | Result |
|---|---|---|
| **allow** | `read_file` (no matching rule) | tool_use forwarded unchanged ✓ |
| **deny** | `run_command` w/ `rm -rf` (matches deny rule) | tool_use → `text` block (`"...denied by clawpatrol: destructive filesystem command..."`), `stop_reason` → `end_turn` ✓ |
| **hitl** | `run_command` `ls -la` (matches hitl rule) | tool_use → `clawpatrol_poll` tool_use carrying the opaque token, `original_tool`/`original_tool_id` preserved, `stop_reason` kept as `tool_use` ✓ |

HITL multi-turn dance, end-to-end through the live HTTP endpoints:
- `GET /api/approval/pending` → parked call with the **real model-generated args** `{"command":"ls -la"}`, reason, timestamp
- `POST /api/approval/poll` (1s) before decision → `{"state":"pending"}`
- `POST /api/approval/decide` `approve` → `{"state":"approved","by":"arnau"}`
- `POST /api/approval/poll` after decision → `{"state":"approved","by":"arnau"}`; pending queue drains to `[]`

Endpoint contract: `/pending` → 401 unauth / `[]` authed; `/poll` unknown token → 404; `/ws` → 501 `{"fallback":"/api/approval/poll"}`. Credential injection verified independently (a bad model id returned Anthropic's `not_found_error` with a real `request_id`, i.e. auth succeeded upstream).

**Confirmed limitation #1 empirically:** the same HITL-triggering request with `"stream": true` was **not gated** — the raw `run_command` tool_use streamed straight through as SSE, no `clawpatrol_poll`. So the gate is a no-op for any streaming client (≈ all real agents) until streaming-response rewriting lands.

Note on harness: the `listen` (`:18443`) socket serves the dashboard/web mux and 302s to `/__login`; the transparent MITM only fires for traffic arriving via the WireGuard netstack (`any dst → :443=mitm`), so a real client must go through `clawpatrol run` / the tunnel, not a direct dial to the listen port.

A real `openclaw` agent routes through the gateway correctly (its outbound traffic hit the MITM), but its embedded agent needs a provider token in `auth-profiles.json` (interactive, TTY-only) to actually call Anthropic — staged for the streaming follow-up.

arnauorriols:
[bot-reply]
Thanks for the end-to-end run against the real MITM + WireGuard pipeline — that's exactly the validation this draft needed, and it's great to see all three verdicts (allow / deny / HITL), the full HITL multi-turn dance, the endpoint contracts (`/pending`, `/poll`, `/decide`, `/ws` 501-with-fallback), and independent credential injection all confirmed against genuine model output.

No code change here — this comment confirms the design behaves as intended rather than requesting one. On the two notes:

- **Streaming not gated** — confirmed and expected. This is design decision #3 in the PR description: gating only fires on buffered, non-streaming `/v1/messages` responses (`Content-Type: */json`, status 200); SSE bodies bypass it entirely. You're right that this makes the gate a no-op for ~all real (streaming) agents until streaming-response rewriting lands. I've filed that as a concrete follow-up so it's tracked rather than buried in the PR prose: it covers rewriting the Anthropic SSE event stream byte-by-byte (parse `content_block_start`/`delta`/`stop` events and apply the verdict to `tool_use` blocks mid-stream). That's the highest-priority gap for making this useful in practice.
- **Harness note** — correct and intended. The `listen` socket (`:18443`) only serves the dashboard/web mux and 302s to `/__login`; the transparent MITM is wired to fire on traffic arriving via the WireGuard netstack (`any dst → :443=mitm`), so a real client has to go through `clawpatrol run` / the tunnel rather than dialing the listen port directly. A direct dial to `:18443` will never hit the gate by design.

Leaving the PR as draft pending the design review it asks for; streaming-response gating is the first thing to build on top once the shape here is agreed.

arnauorriols:
can we implement stream support?

arnauorriols:
[bot-reply]
Yes — and it's the #1 follow-up already flagged in `doc/tool-call-gating.md` ("most `/v1/messages` traffic streams, so the non-streaming-only draft doesn't fire against typical agents"). It's the gap between this draft and a usable feature. Here's the concrete plan; I'd like a steer on scope before I touch code.

**The core tension.** Gating needs the *complete* `tool_use` block (name + full `input` JSON) before it can pick allow/deny/hitl. Anthropic's SSE delivers a tool_use incrementally: `content_block_start` (type=tool_use, id, name) → N× `content_block_delta` (`input_json_delta` fragments) → `content_block_stop`, with `stop_reason` arriving later in `message_delta`. So the input is only complete at `content_block_stop`. We can't decide mid-block.

**Touch points.**
- `cmd/clawpatrol/main.go:2398` — the gate guard explicitly *excludes* `event-stream`. Add a streaming branch that routes `text/event-stream` to a new `gateAnthropicSSE`.
- The inline rewrite currently swaps `resp.Body` for a `bytes.Reader` and sets Content-Length. Streaming must instead wrap `resp.Body` in a transforming `io.ReadCloser` (or `io.Pipe`) and keep chunked / no Content-Length so `resp.Write` flushes frames as they're produced.
- New `internal/toolgate` SSE scanner (split on `\n\n`, parse `event:`/`data:`), reusing `Rule.Evaluate` + `store.Park`. Deny/hitl block synthesis mirrors the JSON path but emits SSE frames.

**Two ways to build it:**
1. **Reassemble → reuse → re-emit** (small, correct-first): tee the whole stream, rebuild the equivalent JSON message, run the existing `GateAnthropicResponse`, then replay original SSE bytes if it didn't rewrite or synthesize a fresh SSE stream if it did. ~reuses the battle-tested rewriter, but negates streaming (TTFT = full-response latency) for every gated-eligible turn.
2. **Per-block buffering** (production shape): pass `message_start` + text blocks straight through (preserves TTFT for the common path); only *hold* a tool_use block's events, accumulate `input_json_delta` until `content_block_stop`, then replay verbatim (allow), or drop+synthesize a text block (deny) / `clawpatrol_poll` tool_use (hitl).

I'd do (1) first to get correctness behind the same `CLAWPATROL_TOOLGATE_RULES` flag, then optimize to (2).

**Gotchas I'd handle:** content-block `index` consistency on drop/replace, parallel tool_use blocks, pass-through of `ping`/`message_delta` usage, fail-open on scanner errors (raw passthrough, matching the JSON path), and a per-block input-accumulation cap to replace the JSON path's 8MB `MaxBytesReader` (which doesn't map onto a stream).

**Question:** want me to implement it in this PR (approach 1 first), or land it as the separate v2 follow-up bead so this draft stays scoped to the non-streaming core? Happy either way — just don't want to balloon the draft without a nod.

arnauorriols:
Implement the stream support in this PR. I'm interested in the optimal approach 2. Do not fail open on errors

arnauorriols:
[bot-reply]
Pushed cc360ad implementing streaming (SSE) tool-call gating — approach 2 (per-block buffering), fail-closed on errors as requested.

**What it does** (`internal/toolgate/anthropic_sse.go`, `GateAnthropicSSE`):
- Non-`tool_use` frames (`message_start`, text blocks, `ping`, usage) stream straight through — **TTFT preserved** for the common path.
- A `tool_use` block's frames are **held** from `content_block_start` to `content_block_stop`, accumulating `input_json_delta` fragments. At stop, with the complete input known, the rule set runs:
  - **allow / unmatched** → held frames replayed **verbatim** (output is byte-identical to upstream, interleaved pings included).
  - **deny** → held frames dropped; synthesised `text` block carries the reason at the same content index.
  - **hitl** → held frames dropped; call parked; `clawpatrol_poll` `tool_use` synthesised at the same index (same token/shape as the JSON path).
- `stop_reason` rewritten `tool_use` → `end_turn` in the trailing `message_delta` iff no `tool_use` survives (all denied/blocked).

**Do not fail open — honored.** The streaming gate never forwards a `tool_use` it couldn't evaluate. An unparseable `content_block_start`, a malformed delta on a held block, an input past the 8 MB cap, or a stream truncated mid-block all resolve to the call being **blocked** (refusal text block) or the stream **terminated** — the raw tool call never reaches the agent. A held `tool_use` is emitted only after a clean *allow*. (The pre-existing JSON path still fails open by design — a parse bug there shouldn't brick a legitimate non-tool turn — and that contrast is now documented in `anthropic.go` and `doc/tool-call-gating.md`.)

**Wiring** (`cmd/clawpatrol/`): the gate guard at `main.go:2398` now branches on Content-Type — JSON → existing buffered path; `event-stream` → `gateAnthropicSSEStream`, which wraps `resp.Body` in an `io.Pipe` transform so frames gate and forward incrementally (chunked, no Content-Length). Dashboard gating notes are attached after `resp.Write` drains.

**Tests** (`internal/toolgate/sse_test.go`): allow-verbatim, deny-rewrite + stop_reason flip, hitl park + poll-token emission, fail-closed on malformed delta, fail-closed on bad `content_block_start`, fail-closed on mid-block truncation, and a mixed allow+deny turn keeping `stop_reason: tool_use`.

`go build ./...`, `go vet`, `gofmt -l .`, `go test ./internal/toolgate/... ./cmd/clawpatrol/...` (incl. `-race` on toolgate) all pass.

arnauorriols:
Sync with origin's main

arnauorriols:
[bot-reply]
Pushed 8974bb0 addressing this comment:
- Merged origin/main into `polecat/ghoul-mpd8ddrp` (was 83 commits behind, now 0 behind / up to date with main).
- Merge was clean — no conflicts. Branch retains the 3 toolgate commits on top of current main.
- Verified post-merge: `gofmt -l .` clean, `go build ./...` and `go test ./internal/toolgate/...` pass. (The `cmd/clawpatrol` embed step still requires the dashboard `dist/` to be built first, which CI handles — unchanged by this sync.)

arnauorriols:
[bot-reply]
Pushed 7b6f279 fixing the red `lint` check — `golangci-lint` had 6 issues on the streaming commit (cc360ad). All mechanical, **no behavior change**:

- **errcheck** (`endpoints.go`) — the two SSE approval-endpoint `fmt.Fprintf` writes weren't error-checked; now `_, _ =` like the sibling `handleWS` write.
- **nilerr** (`anthropic_sse.go`) — the malformed-delta and oversized-input **fail-closed** paths set `blk.errored = true` then `return nil` immediately after a non-nil error, which `nilerr` flags. Restructured the `content_block_delta` handler to mark `errored` and fall through to the existing unconditional `bufferHeld`/`return nil` (a no-op while errored). **Fail-closed semantics are identical** — an errored block is still blocked at `content_block_stop`; only the error-conditioned `return nil` is gone.
- **revive unused-parameter** — `gateAnthropicSSEStream`'s `ev *Event` was unused (toolgate notes flow back via the returned `SSEOutcome` and are attached at the call site in `main.go`), so the param is removed and the one caller updated; `handleWS`'s unused `r` renamed to `_`.
- **revive exported-comment** — added a doc comment to the `Verdict` const block.

Local gates green: `gofmt -l .` clean, `go vet`, `golangci-lint run ./...` (0 issues), and `go test ./internal/toolgate/...` (incl. `-race`) `./cmd/clawpatrol/...` all pass.

arnauorriols:
The tool that polls the status of the parked tool_call cannot be named `clawpatrol_poll`, since the agents don't have that tool. The llm needs to figure out which is the right tool to perform the polling, given the set of tools that the agent declares

arnauorriols:
[bot-reply]
Agreed — the current draft is broken here, and for exactly the reason you give. The HITL path synthesises a `tool_use` named `clawpatrol_poll` (`internal/toolgate/anthropic.go:29`, and the SSE `emitPollBlock`), but the agent's tool dispatcher is built from the tools *it* registered locally. It has no `clawpatrol_poll`, so it cannot execute the call. Worth noting injecting the tool *definition* into the request's `tools[]` wouldn't fix this either — dispatch is agent-local, so the upstream model would emit the call but the agent harness still has no handler for it.

So the synthesised poll call must name a tool the agent already has. Two realistic ways to do that:

1. **Gateway introspects the request's `tools[]` and picks.** The request body is already buffered at the gate callsite (`trackedReqBody`, `cmd/clawpatrol/main.go:2445`); it just isn't threaded into `GateAnthropicResponse`/`GateAnthropicSSE` yet. The gate parses the declared tools, finds an HTTP-capable one (fetch / http_request / a shell tool that can `curl`), and rewrites the HITL call to a `tool_use` for *that* tool, targeting `POST /api/approval/poll`. Deterministic, no extra LLM call — but the "which tool can do HTTP" match is a heuristic and varies per agent.

2. **Let the model choose (your phrasing).** This is the richer design already listed as a v2 follow-up in `doc/tool-call-gating.md` ("No gateway-initiated LLM call"). It needs clawpatrol to make its own upstream call with the agent's tool set so the model selects the polling tool — which means re-injecting the per-credential secret into a gateway-initiated request. Correct, but a meaningfully bigger change.

In both cases there has to be a safe fallback for when the agent declares **no** tool that can poll an HTTP endpoint: in that case fall back to the deny-style `text` block ("a human approval is pending for X; stop and retry once approved") rather than emitting an unexecutable tool call. That keeps the turn coherent for any agent.

Since this is a draft and the fork is architectural, which do you want me to build now?
- (1) the introspect-and-pick path with the text-block fallback (smaller, lands on this branch), or
- (2) go straight for the gateway-initiated LLM-choice path (v2), accepting the credential-injection work.

I'll hold off changing code until you point at one.

arnauorriols:
implement option 2: gateway initiated llm choice. The gateway should act as an agent loop iteration. This means it should get all the llm response, removing the parked tool call from the tool call blocks, and fabricating a user message that tells the llm to chose the right tool for the polling. Then the llm responds with the right tool call to perform the polling. This response is forwarded to the agent, that executes the tool call and polls the status of the tool call that was parked.

arnauorriols:
[bot-reply]
Pushed 3782386 implementing **option 2 (gateway-initiated LLM choice)** for the HITL polling tool. The gateway now acts as one agent-loop iteration instead of synthesising the unexecutable `clawpatrol_poll`.

**Flow** (`internal/toolgate/followup.go`):
1. On a HITL verdict the call is parked, and the parked `tool_use` is **removed** from the assistant turn (text rationale kept).
2. A **user message is fabricated** telling the model to long-poll `POST <base>/api/approval/poll` with `{"token": …}`, choosing the right tool **from the agent's own `tools[]`**.
3. The gateway makes its **own upstream `/v1/messages` call** (`stream:false`, tools/model/system carried over) via an `LLMCaller` callback.
4. The follow-up response — naming a tool the agent actually has — is **forwarded to the agent**, which executes it and polls.

**Credential injection solved without re-fetching the secret** (the worry in design note #6): the follow-up clones the request that's *already* been credential-injected at the MITM hook and round-trips it on the same upstream transport. Anthropic uses header injection (not body signing), so swapping the body is safe. The follow-up never re-enters the MITM handler, so it isn't re-gated.

**Both response shapes:**
- JSON path forwards the follow-up response verbatim.
- SSE path makes the follow-up non-streaming and re-emits the chosen tool_use at the held block's index (TTFT for earlier text preserved).

**Fail-closed fallback (honoring "do not fail open"):** if the follow-up is unavailable (no caller, or the request body was too large to rebuild) or the call errors, the HITL path degrades to a coherent "approval pending" text block. The parked tool call is dropped either way — the raw call never reaches the agent.

**Config:** `CLAWPATROL_TOOLGATE_APPROVAL_URL` sets the poll base URL (must be clawpatrol's agent-reachable address; placeholder default won't connect).

**Known limitations (draft, documented in `doc/tool-call-gating.md`):** N=1 tool_use per turn is the supported shape — a turn mixing HITL with allow/deny siblings is replaced wholesale; each HITL block in a multi-tool stream triggers its own follow-up; the chosen polling tool is not itself re-gated.

**Tests:** HITL JSON + SSE tests rewritten to drive the follow-up via a fake `LLMCaller` and assert a real agent tool (`http_request`) is forwarded (not `clawpatrol_poll`); added follow-up-unavailable, follow-up-error, and `buildFollowupRequest` cases. `go test ./internal/toolgate/...` (incl. `-race`) and `./cmd/clawpatrol/...` pass; `golangci-lint run` 0 issues; `gofmt -l .` clean.

arnauorriols:
[bot-reply]
Refreshed the PR body to match the implementation as it stands at `HEAD` (`3782386`). No code change — this is a description-only sync, since the body had drifted from the code over the review cycle.

What the old body got wrong vs. current code:
- **Streaming was listed as out-of-scope** — but `GateAnthropicSSE` (`internal/toolgate/anthropic_sse.go`, per-block buffering, fail-closed) landed in `cc360ad`. Now documented as implemented.
- **HITL was described as a synthesised `clawpatrol_poll` tool the gateway injects** — that approach was replaced by the **gateway-initiated polling-tool choice ("option 2")** in `followup.go` (`3782386`): the model picks a polling tool from the agent's own `tools[]`. Summary + design decisions #2 and #6 rewritten accordingly.
- **Design decision #3 ("non-streaming only")** corrected to "both buffered JSON and streaming SSE".
- **What's-implemented / stubbed / follow-ups / test plan** reconciled with the green test suite (added the SSE + follow-up + fail-closed tests) and checked the **manual E2E** box (the real-creds MITM + WireGuard run reported earlier in this thread).

The body now matches `doc/tool-call-gating.md`. Still a draft / not-for-merge.

piscisaureus:

PR #475, opened 2026-05-19
dashboard: device detail page with credential pills

arnauorriols:
## Summary

Adds credential pills to the **v1 dashboard** device page — one pill per credential bound to the device's profile, shown in the existing device-header card (`dashboard/src/components/DevicePage.tsx`). Each pill shows the plugin-type brand logo + bare name; a green border means the credential is connected, and clicking a pill opens the same v1 `ConnectModal` the IntegrationsCards already use.

Resolution follows the credential→profile inversion (cl-mtvb / cl-ab3r): device → profile → credentials, read from `/api/profiles`. The handler now emits the richer `ProfileInfo` shape (`{name, credentials}`) the dashboard consumes, instead of a flat `[]string`.

Brand logos are inlined as SVGs in `dashboard/src/components/Logos.tsx` (sourced from the simple-icons set) rather than loaded from the `api.iconify.design` CDN — that host is on common ad-blocker lists, so the previous `` requests silently failed and pills rendered text-only.

Screenshots of the three credential states (none / some / all connected) live under `docs/screenshots/pr-475/`.

### Out of scope
- v2 dashboard — completely removed from this PR; v2 is orthogonal.
- `dashboard/src/components/AgentTypeIcon.tsx` still uses the iconify CDN for agent-type icons in the sessions table; it will hit the same ad-blocker issue but is out of scope here (credential pills only). Follow-up worth a bead.

## Test plan

- [ ] `npx oxfmt --check src` + `npx oxlint src` clean (verified locally)
- [ ] `npm run build` succeeds (verified locally; `tsc -b --noEmit && vite build`)
- [ ] `go test ./... && go vet ./...` clean (verified locally)
- [ ] Manual: open a device page in the v1 dashboard; the header card shows one credential pill per credential on the device's profile
- [ ] Manual: a connected credential pill has a green border; clicking a pill opens the v1 ConnectModal
- [ ] Manual: all brand logos render (no ad-blocker / CDN dependency)

PR #500, opened 2026-05-20
dashboard: device detail page with credential pills (cl-pe7y)

arnauorriols:
Rewrite the PR #475 branch to drop the v2 dashboard work — that scope is orthogonal and will land separately. Keeps only the v1 device-page credential-pill feature:

- DevicePage renders a profile→credentials pill row below the device meta (green border = connected, click jumps to settings).
- /api/profiles returns ProfileInfo{name, credentials} so the page resolves device → profile → credentials in one round-trip.
- Brand logos inlined as SVGs so ad-blockers don't break pills.


arnauorriols:
Is this PR any different than #475 ?

arnauorriols:
[bot-reply]
No — as they currently stand, #500 and #475 are identical. Both branches point to the exact same commit (`4c8f890`), the same tree, and the same merge-base with `main`:

- `git diff origin/polecat/ghoul-mpcpl2ah origin/polecat/deathclaw-mpe4k042` → empty
- Both: 8 files changed, +198 −49 (same files, including the `docs/screenshots/pr-475/*` images)

The only real difference is the **PR description**. #475's body describes a *v2* device page (`#/v2/devices/<ip>`) and a new `v2/components/CredentialPills` component — but its diff contains no `v2/` files. The committed code in both PRs is the **v1** change to `dashboard/src/components/DevicePage.tsx`, which is exactly what #500's description ("drop the v2 work, keep only the v1 credential-pill feature") says.

So: #500's description accurately matches the code; #475's description is stale. Recommend keeping **#500** and closing **#475** as a duplicate.

PR #513, opened 2026-05-20
fix: scope device credential cards to the profile's own declared list

arnauorriols:
credentialsInProfile walked CompiledEndpoint.Credentials (the global endpoint→credentials list) for every endpoint reached by a profile, which leaks sibling-profile credentials onto a device's card whenever two profiles share an endpoint. With the postgres "pg" endpoint bound to both pg-readonly and pg-writer, every device on profile "data" (declares only pg-readonly) saw pg-writer too, and Agents.Integrations (used by the agents table's red "needs setup" dot) flagged credentials the device never actually uses.

Read CompiledProfile.Credentials directly — the profile's HCL-declared list, populated by Compile() from pr.Credentials. Keep walking ep.Tunnel.Via for tunnel-attached auth so tailscale node-auth and the like still surface on profiles whose endpoints route through a tunnel.

Runtime dispatch (ResolveCredential / EndpointCredentials) was already profile-scoped — pinned by a new TestResolveCredentialIsolatesProfilesSharingEndpoint test. Dashboard-only fix.

### DIFF
--- a/cmd/clawpatrol/agents.go
+++ b/cmd/clawpatrol/agents.go
@@ -818,6 +818,13 @@ func dashboardTailscaleState(label tailscaleproto.NodeStateLabel, hasPersistedSl
 // apiStatus and the device-page card render so per-device views only
 // show credentials the device's profile actually uses.
 //
+// Reads strictly from CompiledProfile.Credentials (the profile's own
+// HCL-declared list). Walking ep.Credentials would leak credentials
+// from *other* profiles that share the same endpoint — e.g. the
+// postgres `pg` endpoint carries both `pg-readonly` and `pg-writer`,
+// but profile "data" declares only `pg-readonly` and must not see
+// `pg-writer` on its card.
+//
 // Walking tunnel-attached credentials in addition to the profile's
 // own list lets the dashboard surface tunnel-bound auth (e.g. the
 // tailscale node-auth credential) on profiles whose endpoints reach
@@ -833,14 +840,14 @@ func credentialsInProfile(policy *config.CompiledPolicy, profile string) map[str
 		return map[string]bool{} // unknown profile → empty set, not nil
 	}
 	out := map[string]bool{}
-	for _, ep := range prof.Endpoints {
-		for _, ent := range ep.Credentials {
-			if ent != nil && ent.Symbol != nil {
-				out[ent.Symbol.Name] = true
-			}
+	for _, ent := range prof.Credentials {
+		if ent != nil && ent.Symbol != nil {
+			out[ent.Symbol.Name] = true
 		}
+	}
+	for _, ep := range prof.Endpoints {
 		for tun := ep.Tunnel; tun != nil; tun = tun.Via {
-			if tun.Credential != nil {
+			if tun.Credential != nil && tun.Credential.Symbol != nil {
 				out[tun.Credential.Symbol.Name] = true
 			}
 		}
--- a/cmd/clawpatrol/tailscale_dashboard_test.go
+++ b/cmd/clawpatrol/tailscale_dashboard_test.go
@@ -44,7 +44,8 @@ func TestCredentialsInProfileWalksTransitivelyViaTunnelChain(t *testing.T) {
 	policy := &config.CompiledPolicy{
 		Profiles: map[string]*config.CompiledProfile{
 			"default": {
-				Endpoints: map[string]*config.CompiledEndpoint{"grafana": ep},
+				Credentials: []*config.Entity{endpointCred},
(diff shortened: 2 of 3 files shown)


arnauorriols:
Verify across all the repo that the following spec is honored:
1. When a device performs a request, the profile of the device is selected
2. from the profile of the device, we list all the endpoints that can claim requests by that profile. We can do that because the profile lists credentials, and each credential is linked to an endpoint
3. Among the endpoints to which the credentials in the device's profile are linked, maybe one is found to match the request of the device
4. If the endpoint matches the request, it parses the request, produces an action, rules are matched against that action, and if the verdict is allow, the credential linked to that endpoint is injected into the request. Finally, the request is forwarded upstream
5. if there are more than one credential linked to the same endpoint, a desambiguating field (placeholder, database, user...) must have been specified in the config, and it is used to desambiguate which credential is injected.

arnauorriols:
[bot-reply]
Verified each spec step against the current branch (`polecat/dust-mpekye4h`). The spec is honored end-to-end. No code changes required — the PR's `TestResolveCredentialIsolatesProfilesSharingEndpoint` already pins the only place dispatch could have leaked across profiles.

**1. device → profile** — `Gateway.profileFor(peerIP)` in `cmd/clawpatrol/main.go` (called at lines 1335, 1349, 1403, 1517) maps the WireGuard peer IP to a profile via the onboarding map, falling back to `defaultProfileName(policy)` when nothing is bound.

**2. profile → endpoints, via credentials** — `Compile()` in `internal/config/compile.go:335-407` builds each `CompiledProfile` by walking the profile's `credentials = [...]` list and following `credentialEndpointTargets(credEnt)` for every named credential. That populates `profile.Endpoints`, `profile.HostIndex`, `profile.HostPatterns`, and `profile.EndpointCredentials[epName]`. Endpoint membership is exactly the transitive closure profile → credentials → endpoints.

**3. matching endpoint for the request**
- HTTPS/k8s (https-mitm family): `runtime.HostEndpoint(policy, profile, host)` in `internal/config/runtime/dispatch.go:23-53` first checks `prof.HostIndex[host]` (exact, lowercased, including compile-time bare-host aliases for default-port HTTPS), then walks `prof.HostPatterns` longest-suffix-first. Strictly profile-scoped.
- Postgres / VIP / wire-protocol: `pickEndpointForProfile()` (`main.go:1685-1702`) filters `connIdx.Lookup(dstIP)` / `dnsvip.LookupVIP(dstIP)` candidates to `prof.Endpoints[c.Name]`. No fallback to "any endpoint claiming this IP" — an unclaimed dst relays verbatim rather than cross-profile routing.

**4. parse → action → rules → allow → inject → forward**
- HTTPS: `main.go:1982` calls `runtime.MatchRequest(ep, mreq)` after `fac.ParseRequest` builds the action; on allow, `main.go:2219` calls `runtime.ResolveCredential(policy, profile, ep, mreq)`, fetches the secret, invokes `HTTPCredentialRuntime.InjectHTTP` / `HTTPRequestSigner.SignHTTPRequest` / `WebSocketCredentialRuntime` as appropriate, then forwards.
- Postgres: `postgres.go:238` calls `ResolveCredential` with `Family:\"sql\", Database, User, Meta.Statement=user`, fetches the secret, then dials upstream and sends a real-credential StartupMessage. Rule evaluation runs on `MatchRequest(ep, mreq)` per query (`postgres.go:549,661`).
- Clickhouse-native follows the same shape via its `*_runtime.go`.

**5. multiple credentials on the same endpoint → disambiguator** — `ResolveCredential` (`dispatch.go:188-242`) and `disambiguatorMatches` (`dispatch.go:248-270`) implement this. Supported fields: `placeholder` (via the endpoint plugin's `PlaceholderDetector`), `user` (from `req.User`), `database` (from `req.Database`). Unknown field names fail closed. Most-specific entry wins by count of declared constraints; compile-time uniqueness validation in the profile dispatch table rejects ambiguous configs, and the runtime returns nil on the unreachable tie rather than silently mis-routing. The PR's `TestResolveCredentialIsolatesProfilesSharingEndpoint` pins the "two profiles share an endpoint with different credentials, neither sees the other's" case.

One related dashboard-side note: `credentialsInProfile` (`cmd/clawpatrol/agents.go:834`) — the function this PR fixed — is the only consumer that walks the profile's credential graph for display. It now reads `prof.Credentials` (the profile's own HCL list) plus tunnel-attached auth, matching the dispatch semantics. The previous version walked `ep.Credentials` (the global per-endpoint list), which is why device cards leaked sibling-profile credentials. Tunnel-attached auth is intentionally included so tailscale node-auth-style credentials still surface on profiles whose endpoints route through a tunnel.

PR #532, opened 2026-05-21
fix: populate ConnHandle.Blobs in conn dispatch (unblocks SSH)

littledivy:
## Summary
SSH connections through the gateway have been failing with a TCP reset mid-kex (\`kex_exchange_identification: read: Connection reset by peer\`) since at least #455. The gateway log shows the cause:

\`\`\`
ssh direct 65.20.66.139:22: ssh runtime needs a BlobStore to persist host keys
\`\`\`

\`ssh.HandleConn\` rejects \`ConnHandle\`s whose \`Blobs\` is nil — it needs the BlobStore to load / mint the per-endpoint host key. Both ConnHandle construction sites in \`main.go\` left \`Blobs\` unset:
- the postgres-only path (\`g.handlePGConn\`, line 1448),
- the generic \`dispatchConnEndpoint\` shared by VIP + direct-IP paths (line 1602).

This stores the gateway's BlobStore on the \`Gateway\` struct (next to \`sink\`) and wires it through both sites. Postgres doesn't currently use it, but the field is documented as the API for plugin persistence so future endpoints don't trip the same wire.

## Test plan
- [x] \`go build ./...\` + \`go test ./...\` green.
- [x] \`gofmt -l .\` clean.
- [ ] Deploy to clawpatrol-gateway and confirm \`ssh user@orchid\` reaches the upstream banner instead of getting RST.

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -313,10 +313,15 @@ type Gateway struct {
 	certs    *CertCache
 	dialer   *net.Dialer
 	sink     *Sink
-	oauth    *OAuthRegistry
-	agents   *AgentRegistry
-	hitl     *HITLRegistry
-	onboard  *onboardRegistry
+	// blobs is the gateway-side plugin blob store (sqlite-backed).
+	// Used by endpoint plugins that need per-endpoint persistent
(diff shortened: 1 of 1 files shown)


littledivy:
Folding into #531.

PR #550, opened 2026-05-21
Re-fetch env-pushdown on every session start (fix #546)

### LINKED ISSUE #546: clawpatrol run does not pick up newly connected credentials until daemon restart
## Summary

`clawpatrol run -- env` does not pick up newly added credentials while the daemon is already running. Restarting/killing the daemon makes the same command show the expected environment variables.

## Repro

Reported by @aornn in Slack:

1. Start with the daemon already running.
2. Add a new credential and connect it.
3. Run `clawpatrol run -- env`.
4. The new environment variable is missing.
5. Kill the daemon.
6. Run `clawpatrol run -- env` again.
7. The environment variable is now present.

## Expected

`clawpatrol run -- env` should reflect newly connected credentials without requiring a daemon restart.

## Notes

This surfaced while testing first-run UX around `eval $(clawpatrol env)` / shell integration. Arnau confirmed the issue appears daemon-related: once the daemon is restarted, `clawpatrol run -- env` sees the credential.

piscisaureus:
Closes #546.

The daemon cached the gateway's env-pushdown response for 30s. A
credential connected on the dashboard wasn't visible to
\`clawpatrol run -- env\` until either the TTL expired or the daemon
was restarted — the user-reported symptom in #546.

* Drop the TTL. Every START / ENV call now hits the gateway. The
  gateway lives on-tailnet so the round-trip is sub-50ms; agent
  sessions are not noticeably slower. Pre-warm at daemon boot goes
  away with the cache — the first session pays the same round-trip
  as every subsequent one.
* Keep the "transient gateway error" guard in a smaller form:
  \`daemonFetchEnvPushdown\` can't distinguish "no vars declared"
  from "gateway down" (both encode as \`[]\`), so if a fetch comes
  back empty and a previous fetch in the same daemon lifetime was
  non-empty, serve the last known-good list. New \`lastGoodEnv\`
  field, guarded by the existing \`envMu\`, replaces \`envVars\` +
  \`envFetchedAt\`.
* macOS NE path already re-fetched per session via the \`getenv\`
  IPC verb, so this change brings Linux to behavioral parity.

Net diff: -54 / +31 LOC.

### DIFF
--- a/cmd/clawpatrol/daemon_linux.go
+++ b/cmd/clawpatrol/daemon_linux.go
@@ -53,14 +53,6 @@ import (
 const (
 	daemonIdleTimeout  = 5 * time.Minute
 	daemonHelloTimeout = 2 * time.Second
-
-	// envCacheTTL bounds how stale a cached env-pushdown response can
-	// be before the next START / ENV call re-fetches it from the
-	// gateway. Trades off "profile changes propagate instantly" against
-	// "every bursty agent session hits the gateway". 30s keeps a fleet
-	// of short-lived `clawpatrol run` invocations to ≤2 fetches/min
-	// while making a manual profile change visible inside one prompt.
-	envCacheTTL        = 30 * time.Second
 	daemonSpawnTimeout = 30 * time.Second
 	daemonMagicLine    = "CLAWPATROL/1\n"
 )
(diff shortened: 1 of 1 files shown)


piscisaureus:
Superseded by #549, which landed first with a background-polling approach (refresh every 2s, keep 30s TTL as stall guard) instead of fetch-on-every-call. Bug fixed on main either way; closing this one.

PR #560, opened 2026-05-21
join: --login bootstraps tsnet for tailnet-only gateways

piscisaureus:
Onboard a fresh machine to a clawpatrol gateway that has no public
URL. tsnet is already statically linked into the binary, so the CLI
can drive an interactive Tailscale login itself — no system
tailscaled install, no \`tailscale up\`, no PATH dependency.

### `--login` flag

When a gateway has no public Funnel (or this machine simply can't
reach it from its current network), pass \`--login\` to bring up a
temporary tsnet node, walk the operator through Tailscale's
standard browser-auth flow, reach the gateway over the tailnet,
run the existing device-flow approval, then tear the bootstrap
node down. The agent's persistent identity is the gateway-minted
tagged auth-key; the human's tailnet credentials never persist on
the agent machine.

Auto-fallback (no flag needed) kicks in when the gateway URL is
tailnet-shaped — CGNAT \`100.64.0.0/10\` literal or \`*.ts.net\` host —
and the initial \`/ca.crt\` probe fails with a network-unreachable
error. Operators dragging a fresh laptop onto a tailnet don't have
to know the flag exists.

### Auto-approve via operator identity

The bootstrap dials the gateway as a Tailscale-authenticated peer.
When the operator who completed the browser-login step is in the
gateway's dashboard operators allowlist, the gateway's
\`tailnetGate\` accepts \`/api/onboard/approve\` from this client just
like it would from the dashboard \"Approve\" button — no separate
human-in-the-loop click needed. The CLI prints \"Auto-approved via
tailnet operator identity.\" and the polling loop resolves
immediately. Non-operator users hit the same 403 the dashboard
gate produces; we silently fall through to the standard \"Verify
code in browser:\" prompt with no warning, since that's the
expected outcome for any user not on the allowlist.

### Credentials in RAM

The bootstrap tsnet uses a private \`ramStateStore\` implementation
of \`ipn.StateStore\` so the machine key, login profile, and netmap
snapshot live in memory only — process exit by any means leaves
nothing sensitive on disk.

tsnet ships its own in-memory store at \`ipn/store/mem\`, but
\`tsnet.Server\` gates that one behind \`Ephemeral=true\` (\`isMemStore\`
type-assert in \`tsnet/tsnet.go\`). Setting \`Ephemeral=true\` sends
\`LoginEphemeral\` on the registration RPC, which Tailscale SaaS
only honors on the auth-key flow — empirically the browser auth
completes but the resulting node never transitions to
\`BackendState=Running\` and the join times out. By presenting our
own non-\`*mem.Store\` implementation we satisfy the same interface,
dodge the gate, and keep browser auth working.

tsnet still creates a directory to write its log buffer
(\`tailscaled.log.conf\` + \`tailscaled.log*.txt\`) regardless of
\`Store\`. Those aren't credentials, and they land in
\`os.MkdirTemp(\"\", …)\` — i.e. \`/tmp\` or \`$TMPDIR\`, both tmpfs on
Linux / per-user volatile on macOS. The happy-path \`Close\` removes
them; an uncaught signal leaves a few KB of log noise the OS
reaps. No signal handler.

### URL validation

\`clawpatrol join clawpatrol-gateway-1\` (no scheme) used to fail at
the dial layer with \"unsupported protocol scheme\", but only after
\`preJoinFetchCA\` had already overwritten \`~/.clawpatrol/gateway\`
with the bogus string. Subsequent \`clawpatrol env\` / \`clawpatrol
run\` invocations would then read the broken URL out of disk and
produce a fresh round of confusing errors.

\`validateGatewayURL\` now rejects scheme-less hostnames upfront,
before any I/O. The error message names the most likely correct
form so the operator can copy-paste the fix:

\`\`\`
gateway URL \"clawpatrol-gateway-1\" is missing an http:// or
https:// scheme — for a tailnet-mounted gateway, try
http://clawpatrol-gateway-1:8080
\`\`\`

No auto-promotion: the right port differs between Funnel (\`:443\`)
and tailnet (\`:8080\`) and we don't have enough context at this
layer to guess.

### Plumbing

\`preJoinFetchCA\`, \`fetchCAHTTP\`, and \`onboardViaDeviceFlow\` now
accept an optional \`*http.Client\` so the bootstrap path can thread
its tsnet-dialing client through without forking the request
plumbing. \`nil\` keeps the existing TOFU-permissive default.
\`isNetworkUnreachableErr\` is deliberately narrow — host/net
unreachable, connection refused, DNS miss, and dial timeout flip
the auto-fallback; TLS handshake timeouts, HTTP 5xx, and cert
errors mean the gateway IS reachable but misbehaving at a higher
layer, and a tailnet bootstrap wouldn't help.

### Verified end-to-end

On a fresh Ubuntu VM (\`ssh exe.dev new\`) against the dev gateway:

* \`clawpatrol join --login http://clawpatrol-gateway-1:8080\`
  → \"Tailnet login complete.\" → \"Auto-approved via tailnet
  operator identity.\" → \"Approved.\" → \`~/.clawpatrol/\` populated,
  tsnet-daemon mode persisted, bootstrap dir removed, no
  \`tailscaled.state\` ever written.
* \`clawpatrol join --login --whole-machine
  http://clawpatrol-gateway-1:8080\` → same auto-approve flow →
  \`tailscale up --authkey=<gateway-minted>\` placed the tagged
  identity (100.104.72.120) on the tailnet → exit-node + DNS pin
  + SSH exemption applied; SSH session survived the exit-node
  flip.

### DIFF
--- a/cmd/clawpatrol/fetch_ca_test.go
+++ b/cmd/clawpatrol/fetch_ca_test.go
@@ -22,7 +22,7 @@ func TestFetchCAHTTPReturnsFingerprintAndPersistsCert(t *testing.T) {
 	}
 
 	dst := filepath.Join(t.TempDir(), "ca.crt")
-	got, err := fetchCAHTTP(srv.URL, dst)
+	got, err := fetchCAHTTP(srv.URL, dst, nil)
 	if err != nil {
 		t.Fatalf("fetchCAHTTP: %v", err)
 	}
@@ -40,7 +40,7 @@ func TestFetchCAHTTPRejectsNonPEMBody(t *testing.T) {
 	}))
 	defer srv.Close()
 	dst := filepath.Join(t.TempDir(), "ca.crt")
-	if _, err := fetchCAHTTP(srv.URL, dst); err == nil {
+	if _, err := fetchCAHTTP(srv.URL, dst, nil); err == nil {
 		t.Fatal("expected error for non-pem body")
 	}
 	// installCATrust must never see a malformed file. If a future
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -2721,6 +2721,9 @@ usage:
                   --hostname NAME        device name to register (default: os.Hostname)
                   --profile NAME         suggest a profile for the approver
                   --whole-machine        bring up wg-quick (route all traffic)
+                  --login                interactive tailnet login for gateways
+                                         with no public URL (creds discarded
+                                         once join completes)
   clawpatrol run -- <cmd> [args...]      route one process tree through gateway
   clawpatrol status                      report install + tunnel state
   clawpatrol uninstall                   remove local join state and tunnel config
--- a/cmd/clawpatrol/setup.go
+++ b/cmd/clawpatrol/setup.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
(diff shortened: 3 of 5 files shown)


littledivy (approved):
LGTM

PR #561, opened 2026-05-22
macos: auto-recover from stale tsnet state on auth-key reject

littledivy:
## Bug

Mac mini install of v0.2.1 (and presumably every other host that ran v0.2.0 first) can't bring up NE's tsnet. The fresh \`clawpatrol join\` mints a valid auth key, NE.startProxy fires \`ts_netstack_init\`, and Tailscale immediately rejects the key:

\`\`\`
ts_netstack_init: tsnet up: tsnet.Up: backend: invalid key: API key kgZXnD5TAr11CNTRL not valid
\`\`\`

Re-joining doesn't help — every fresh key gets the same rejection.

The only workaround is to manually \`sudo rm -rf "/var/root/Library/Application Support/clawpatrol"\` and re-join. After that the NE comes up cleanly and stays up.

## Root cause

#544 introduced a persistent state dir for tsnet. Upgrading from a build that wrote a half-initialised \`tailscaled.state\` into that path (v0.2.0 → v0.2.1 in the wild, but also any prior \`startProxy\` that died mid-Up) leaves stale state behind. \`tsnet.Server.Up\` reads the stale state, resumes against whatever control plane / node identity it names, and tries to validate the freshly-minted auth key against THAT identity. Control returns \"invalid key\". NE never reports a tsnet IP, every \`clawpatrol run\` lands in the default profile with no tunnel.

## Fix

If \`tsnet.Server.Up\` fails on the first attempt with an \"invalid key\" / \"not valid\" error, \`os.RemoveAll(stateDir)\` + recreate the dir + retry \`Up\` once with the supplied fresh auth key. Lets the upgrade path self-heal without operator sudo.

Subsequent restarts pick up the now-good saved identity and never hit the retry path — single-use key still consumed exactly once.

## Test plan
- [x] \`CGO_ENABLED=1 GOOS=darwin GOARCH=arm64 go build -buildmode=c-archive\` clean.
- [x] Full \`build-macos-app.sh\` produces signed-locally \`Clawpatrol.app\`.
- [x] Manual confirmation that wiping state dir + re-join is what unblocks on Mac mini (the workaround this PR automates).
- [ ] Install on the still-broken Mac mini, watch \`ts_netstack_init: stale state detected\` log line, then successful Up.

### DIFF
--- a/macos/netstack/wgnetstack.go
+++ b/macos/netstack/wgnetstack.go
@@ -29,6 +29,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net"
 	"net/http"
 	"net/netip"
@@ -802,6 +803,21 @@ var (
 	tsStarted bool
 )
 
+// isStaleStateErr matches tsnet.Server.Up errors that indicate the
+// persisted state in Dir is from an incompatible past (different
+// control plane, expired/revoked auth key tied to old node identity,
+// half-written tailscaled.state from a crashed prior init). Recovery
+// is to wipe the state dir and retry with the supplied fresh auth
+// key. Matches on the message Tailscale's control returns:
+// "backend: invalid key: API key XXXX not valid".
+func isStaleStateErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "invalid key") || strings.Contains(msg, "not valid")
+}
+
 // ts_netstack_init joins the tailnet with the given ephemeral auth key
 // and configures the gateway as the tsnet exit node so subsequent dials
 // route through it. Blocks until the tsnet node has a 100.x.x.x address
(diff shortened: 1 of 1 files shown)


piscisaureus:
Reviewed against current main (rebases cleanly, builds). The symptom is real, since the manual `rm -rf` of the tsnet state dir fixed the Mac, but I don't think this shape should land:

- The diagnosis can't be right as written: v0.2.0 predates the persisted state dir (it used a temp dir with `Ephemeral: true` and removed it), so whatever stale state that Mac had came from a later build. Realistic triggers are the node being deleted in the tailnet admin console, node-key expiry, or the gateway moving to another tailnet while the Mac kept its old identity.
- The retry only helps on the join path. The NE re-sends the stored single-use auth key on every start, so when control rejects during a plain `clawpatrol run` restart, the wipe deletes the machine key and the retry fails with the same consumed key. The node is then orphaned and the next join creates a brand-new device instead of re-keying the old one, which is worse than today.
- `"not valid"` as a second substring adds false positives (it matches the local machinekey parse error), and `os.RemoveAll`'s error is ignored, so a half-wiped dir is retried silently.
- Nothing in `macos/netstack` is logged in a way that reaches unified logging, so the recovery is invisible in production, and the "verify on the broken Mac mini" step in the test plan is unchecked.

The deterministic shape is better: when join hands the NE a fresh auth key, reset the state dir before the first `Up`, and keep the string match at most as a narrow (`backend: invalid key`) logged fallback. If you want to pursue it that way I'll review promptly; otherwise I'll close this in a while.

PR #572, opened 2026-05-22
Replace demo image with live iframe

josh-collinsworth:
Instead of having a static image of the dashboard, let's use a live iframe to show visitors the demo firsthand.

This keeps it in sync with any changes we might make, and users can still opt to click through to it.

CleanShot 2026-05-22 at 15 22 54@2x


ry (approved):
LGTM

PR #573, opened 2026-05-23
human_approver: per-request channel override via X-HITL-Channel

littledivy:
## Summary

`human_approver` blocks today hard-code a single destination channel. An agent that investigates incidents across multiple Slack channels needs the approval prompt to land in whichever channel the originating session came from — which today requires N approver blocks and N rules with header-routing CEL on each rule.

This adds the same per-request override that `ThreadTS` already has, just for channel. When the inbound request carries `X-HITL-Channel: <id>`, the validated value replaces `h.Channel` at notification time. Empty/invalid header → fall back to the static block field, so existing configs are unchanged.

## Why mirror ThreadTS rather than add a CEL facet

The full-CEL approach (CEL-evaluable `channel` reading a `slack.session.channel` facet) was the first design considered — but `ThreadTS` already plumbs an identical per-request value via the `X-HITL-Thread-TS` header, validated and dispatched the same way. Reusing that shape:

- ~150 LOC instead of ~500
- No new facet, no grammar / parser changes, no doc burden on the operator
- Behaves intuitively next to `ThreadTS` (same source, same lifetime, same fallback story)
- Composes with `ThreadTS` cleanly: a request can pin both the channel AND the thread

The trade-off is the value is set by whoever proxies the request (today: the agent's own openclaw, mirroring how it sets `X-HITL-Thread-TS`), not declaratively in HCL.

## Plumbing

- `ApproveRequest.Channel` field, doc'd alongside `ThreadTS` (`internal/config/runtime/runtime.go`)
- `runApproveCtx.Channel` + read from `req.Header.Get(\"X-HITL-Channel\")` at the HTTP entry point + plumbed through `runApproveChain` into `ApproveRequest` (`cmd/clawpatrol/main.go`)
- `sanitizeSlackChannelHeader` validates: shape `^[CG][A-Z0-9]{8,23}$`, length-bounded, whitespace-trimmed. DMs (D…) and user IDs (U…) rejected — not valid HITL destinations. Lowercase / punctuation / injection shapes dropped to empty (caller falls back to static channel)
- `X-HITL-Channel` added to the upstream-strip list alongside `X-HITL-Thread-TS` so it never leaks to the real upstream
- `HumanApprover.Approve` prefers `req.Channel` over `h.Channel` for `HITLTarget.Channel`; falls back when empty

## Test plan

- [x] `go build ./...` clean
- [x] New: `TestSanitizeSlackChannelHeader` — public/private accepts, DM/user/lowercase/injection rejects, length bounds, whitespace trim
- [x] New: `TestHumanApproverRequestChannelOverridesBlockChannel` — `req.Channel != \"\"` wins over `h.Channel = \"C-STATIC\"`
- [x] New: `TestHumanApproverEmptyRequestChannelFallsBackToBlockChannel` — `req.Channel == \"\"` keeps existing behavior
- [x] Full `internal/config/plugins/approvers` + `cmd/clawpatrol` suites pass (~28s)
- [x] `gofmt -l .` clean

## Caller-side rollout (separate from this PR)

The agent / openclaw will start emitting `X-HITL-Channel: <slack channel id>` on outbound HTTP requests originating from a given Slack session. Once shipped, a single `human_approver \"incident-response\" {}` block can route prompts to whichever channel triggered the agent — no per-channel approver duplication.

### DIFF
--- a/cmd/clawpatrol/hitl_channel_header_test.go
+++ b/cmd/clawpatrol/hitl_channel_header_test.go
@@ -1,0 +1,54 @@
+package main
+
+import "testing"
+
+// sanitizeSlackChannelHeader gates the X-HITL-Channel header before
+// it's threaded into the approver request. The header is
+// agent-controlled, so anything that doesn't shape like a Slack
+// channel ID must be dropped — otherwise an attacker-shaped value
+// could land in chat.postMessage verbatim.
+func TestSanitizeSlackChannelHeader(t *testing.T) {
+	cases := []struct {
+		in   string
+		want string
+	}{
+		// Public + private channels accepted.
+		{"C0AH1SJGHAP", "C0AH1SJGHAP"},
+		{"C03P8NG2Q06", "C03P8NG2Q06"},
+		{"GABC12345", "GABC12345"}, // private channel / mpim
+		// Whitespace trimmed.
+		{"  C0AH1SJGHAP  ", "C0AH1SJGHAP"},
+
+		// Empty in, empty out — caller falls back to static block field.
+		{"", ""},
+
+		// DMs and user IDs rejected — not valid HITL destinations.
+		{"D12345678", ""},
+		{"U12345678", ""},
+
+		// Too short / too long.
+		{"C123", ""},
+		{"C" + repeat("A", 25), ""},
+
+		// Wrong shape — lowercase, punctuation, injection attempts.
+		{"c0ah1sjghap", ""},
+		{"C0AH1S JGHAP", ""},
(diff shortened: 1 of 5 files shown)


littledivy:
Closing — design relies on the agent setting X-HITL-Channel, which means a prompt-injected or misbehaving agent can route approvals to the wrong channel. HITL routing must be infrastructure-enforced, not agent-cooperative. Reopening once a proper design lands: gateway-side slack.session facet whose channel is derived from infrastructure (mTLS cert identity / openclaw session registry), not request headers.

PR #582, opened 2026-05-27
fix: scope settings credential Profiles column to each profile's own list

arnauorriols:
## Summary

`credentialBindings` (cmd/clawpatrol/agents.go) populated the settings-page credentials table's **Profiles** cell by collecting every endpoint globally bound to the credential and then claiming any profile whose endpoint set intersected. That walks the global endpoint→credentials list — the inverse of the bug fixed in commit b0a3813 (#cl-lgwg) for the device page.

With the postgres `pg` endpoint binding both `pg-readonly` and `pg-writer`, profile `data` (declares only `pg-readonly`) showed up under `pg-writer`'s row on the settings credentials table, and vice versa.

Fix: match each profile against `CompiledProfile.Credentials` directly (the profile's HCL-declared list, populated by `Compile()` from `pr.Credentials`) and keep the tunnel-chain walk for tunnel-attached auth — same data sources as `credentialsInProfile`, so the per-credential view and the per-device view stay consistent. The **Endpoints** column is unchanged: it's genuinely the global endpoint→credentials projection.

Runtime dispatch (`ResolveCredential`) was already profile-scoped — re-ran the existing `TestResolveCredentialIsolatesProfilesSharingEndpoint` pin, still green.

## Test plan

- [x] `go test ./...` — full suite green
- [x] `TestCredentialBindingsDoesNotLeakSiblingProfiles` — new, pins credential→profiles direction with the shared-endpoint repro
- [x] `TestCredentialBindingsSurfacesTunnelAttachedCredentials` — new, pins the tunnel-walk path
- [x] `TestCredentialBindings` — existing happy-path cases still pass
- [x] `TestResolveCredentialIsolatesProfilesSharingEndpoint` — re-run, runtime matcher still isolates profiles
- [x] `gofmt -l .` clean

### DIFF
--- a/cmd/clawpatrol/agents.go
+++ b/cmd/clawpatrol/agents.go
@@ -715,26 +715,36 @@ func (w *webMux) statusList(r *http.Request) []IntegrationRow {
 // column for profiles and one for endpoints so an operator can see
 // at a glance where a credential is in play.
 //
-// Profiles are matched strictly against each CompiledProfile's own
-// HCL-declared `credentials = [...]` list (plus the tunnel chain on
-// the profile's endpoints for tunnel-attached auth). Walking the
-// global endpoint→credentials list would leak sibling profiles
-// onto the row whenever two profiles share an endpoint — e.g. the
-// postgres `pg` endpoint carries both `pg-readonly` and `pg-writer`,
-// and profile "data" (declares only `pg-readonly`) must not appear
-// in `pg-writer`'s Profiles cell. Mirror of the device-page fix in
-// cl-lgwg / commit b0a3813.
+// Both columns read in the canonical "credentials bind endpoints"
+// direction:
 //
-// Endpoints stay on the global binding — that column is genuinely
-// "which endpoints declare this credential", profile-agnostic.
+//   - Endpoints: the credential's own `endpoint` / `endpoints`
+//     framework attr (via config.CredentialEndpointTargets), plus any
+//     endpoint whose tunnel chain attaches this credential.
+//   - Profiles: each CompiledProfile's own HCL-declared `credentials`
+//     list, plus the tunnel chain on the profile's endpoints for
+//     tunnel-attached auth.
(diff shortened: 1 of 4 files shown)


arnauorriols:
1. is `endpointBindsCredential` used anywhere anymore? It shouldn't since nowadays credentials are the ones binding endpoints. 
2. I'm surprised there's nothing to reuse from https://github.com/denoland/clawpatrol/commit/b0a3813ac9b08c401a1c7e21ae6aaad1221878a5, since the bug is essentially the same. Not saying there's anything to reuse necessarily, but evaluate the possibility.

arnauorriols:
[bot-reply]
Pushed 2e6c05f addressing both points:

1. **endpointBindsCredential dropped.** You were right — `credentialBindings` was the last caller, and it re-derived `CompiledEndpoint.Credentials` (the inverted index) instead of asking the credential which endpoints it binds. The Endpoints column now reads `config.CredentialEndpointTargets(ent)` directly (canonical direction), with the tunnel-chain walk kept inline for tunnel-attached auth (credentials pulled in by a tunnel rather than declared on any endpoint). Exported `CredentialEndpointTargets` so callers outside the package have a single source of truth.

2. **Reuse from b0a3813 (cl-lgwg).** The profile-side mirror was already in this PR (`profileBindsCredential` reads `prof.Credentials` + tunnel chain, same shape as `credentialsInProfile`). What was left to mirror was the *direction-of-lookup* lesson on the Endpoints column — that's what this refactor lands. Same data sources across `credentialsInProfile` / `profileBindsCredential` / `credentialBindings` now, so all three views stay consistent.

Behavior unchanged: `ep.Credentials` is just the inverted index of `CredentialEndpointTargets`, so direct bindings produce the same set. Existing tests pass; hand-built test 2 needed `policy.Credentials` + `Framework.RefLists` populated to feed the canonical lookup.

PR #584, opened 2026-05-27
fix: wait for daemon transport readiness before exec'ing wrapped command

arnauorriols:
## Summary
- Adds `WaitReady(ctx)` to the `daemonTransport` interface; wg polls wireguard-go's `last_handshake_time_sec`, tsnet re-runs the gateway:53 exit-node probe. Both cache the first observed success in an atomic so warm-path callers pay one atomic load.
- The daemon's session handler runs `WaitReady` between the TUN handoff and the ATTACHED reply with a 30s timeout; on failure it ships a new `READYERR <n>\n<body>` frame so the wrapper exits with the actual reason instead of a misleading "expected ATTACHED, got ..." parse error.
- Client's ATTACHED read deadline grows from 5s → 35s to outlast the daemon's own readiness budget; parser now handles both `ATTACHED` and `READYERR`.

Closes cl-lmuk.

## Test plan
- [x] Cold start: `pollWGReady` / `pollTsnetReady` keep retrying until a stubbed `getCfg`/`dial` flips to ready, then return without blowing the budget.
- [x] Warm start: both pollers return after a single call with no sleeps (huge interval, short ctx — would expose any unconditional pre-sleep).
- [x] Timeout: pollers return `context.DeadlineExceeded`; daemon writes `READYERR`; client parser surfaces the body verbatim as the returned error.
- [x] `READYERR` with empty body: client still returns a "not ready" error rather than reading zero bytes.
- [x] `go test ./...` clean, `gofmt -l .` clean, `go vet ./...` clean.

### DIFF
--- a/cmd/clawpatrol/daemon_linux.go
+++ b/cmd/clawpatrol/daemon_linux.go
@@ -91,10 +91,27 @@ type daemonTransport interface {
 	// Each `clawpatrol run` repeats it on stderr so configuration
 	// issues surface without tailing daemon.log.
 	BootWarning() string
+	// WaitReady blocks until the transport can actually carry user
+	// traffic — wg: a handshake has completed; tsnet: exit-node
+	// routing is up (the gateway is reachable on the tailnet). The
+	// daemon calls this once per session, after the TUN handoff and
+	// before signalling ATTACHED, so the wrapped child never execs
+	// into a tunnel that silently drops its first packets. Returns
+	// nil immediately on the warm path once readiness has been
+	// observed at least once this daemon lifetime. Returns ctx.Err()
+	// (wrapped) on timeout.
+	WaitReady(ctx context.Context) error
 	// Close tears down the transport. Best-effort — the daemon is
 	// about to exit.
 	Close() error
 }
+
+// daemonReadyTimeout caps how long a session START waits for the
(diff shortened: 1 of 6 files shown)


arnauorriols:
[bot-reply]
The PR body itself contains no open question or change request — it's a descriptive summary. But while reviewing it I noticed the body's test plan claims lint is clean, whereas CI's `lint` job was failing.

Pushed `09730ad4` to fix it:
- `golangci-lint` (revive `unused-parameter`) flagged the unused `ctx` parameter on the two `dial` closures in `cmd/clawpatrol/daemon_ready_linux_test.go` (lines 122, 148). Renamed both to `_`.

The original "gofmt/go vet clean" check didn't exercise `golangci-lint`, which is what CI runs (`make lint`), so the revive failure slipped through. Verified locally with the same `golangci-lint v2.12.2`: `0 issues`; `gofmt -l .` clean, `go vet ./...` clean, package tests pass.

piscisaureus:
Landed on main as 0da7964d, attributed to you. The fakeTransport in the newer session tests needed the WaitReady method after #786, the client deadline is now derived from daemonReadyTimeout, and there is a test driving handle() through the READYERR path. Thanks.

PR #585, opened 2026-05-27
fix: drive relay recv/send through SyscallConn.Read/Write (cl-0jsq)

arnauorriols:
## Summary

The relay's worker recv loop and supervisor accept-loop send both used the pattern `int(file.Fd())` + raw `unix.Recvmsg`/`unix.Sendmsg`, handing the int to a goroutine while letting Go's reference to the `*os.File` lapse. Two failure modes followed from the same anti-pattern; the previous iteration of this PR fixed one and exited on the other.

**EAGAIN.** `os.NewFile`'s poller integration sets `O_NONBLOCK` on the underlying file description. The inheriting process reads via raw `Recvmsg` (bypassing the poller) and surfaces EAGAIN whenever the kernel queue is briefly empty. Previously surfaced as `[clawpatrol relay-worker] recv: resource temporarily unavailable` followed by silent worker death. Previously addressed by classifying as transient + looping (`isTransientRecvErr`).

**EBADF.** Under heap pressure the `*os.File` becomes liveness-dead immediately after `.Fd()` returns; the runtime collects it; its `runtime.SetFinalizer`-installed `Close()` closes the very fd the goroutine is reading from. `Recvmsg` then returns EBADF. The previous fix explicitly exited on this errno (`isTransientRecvErr`'s test pinned `{"EBADF", syscall.EBADF, false}`). Reproducible: probing with `fcntl(F_GETFL)` from the failing recv site shows the syscall *also* returns EBADF — i.e. the fd is closed in our own fd table, not the peer's.

Both modes stem from the int-handoff. Both go away by holding a `syscall.RawConn` instead and driving `recvmsg`/`sendmsg` inside `RawConn.Read`/`Write` callbacks:

- **RawConn carries an internal `*os.File` reference** for the goroutine's lifetime. GC can't collect the file out from under us, so EBADF can't surface from inside our own process. No `runtime.KeepAlive` needed.
- **RawConn.Read/Write integrate with the runtime poller.** `EAGAIN`/`EWOULDBLOCK`/`EINTR` returned by `Recvmsg`/`Sendmsg` inside the callback turn into "return false" → park on the poll wait → re-run when the fd becomes ready. The outer loop never observes these errno classes.

`isTransientRecvErr` and the retry helper become dead weight. `sendMu` protecting concurrent `Sendmsg` across multiple `acceptLoop` goroutines becomes unnecessary too — `RawConn.Write` takes a per-RawConn lock internally and serialises concurrent writes for free.

## Why this matters for #589

PR #589 introduces a second recv loop in the **supervisor** (host-loopback direction). Empirical testing of #589 reproduced the EBADF mode at ~60% under 20-way concurrent load. The supervisor's new loop uses the same int-handoff anti-pattern this PR removes; once #589 rebases onto this branch, applying the same `RawConn.Read`/`Write` pattern to its loopback loop closes the same family of bugs there with no further plumbing.

## Test plan

- [x] `go test ./...` — full suite green.
- [x] `gofmt -l cmd/clawpatrol/relay_linux.go cmd/clawpatrol/relay_linux_test.go` — clean.
- [x] `go vet ./cmd/clawpatrol/` — clean.
- [x] New tests against real `SOCK_SEQPACKET|SOCK_NONBLOCK` socketpairs:
  - `TestRecvJobReceivesFrame` — happy path, extracts port + SCM_RIGHTS fd.
  - `TestRecvJobAbsorbsEAGAINViaPoller` — recv on empty non-blocking sock parks in the poller rather than surfacing EAGAIN; unblocks when a frame arrives. **This is the property that replaces `isTransientRecvErr`.**
  - `TestRecvJobReturnsEOFOnPeerClose` — clean shutdown semantics.
  - `TestRelayWorkerLoopDispatchesEndToEnd` — two frames in, two dispatch callbacks out, clean loop exit when sender closes.
  - All four pass 5× in a row, no flakes.
- [x] End-to-end with built binary: `clawpatrol run` + 20-way concurrent host → agent-netns traffic across 5 runs; **zero relay-recv errors** logged. The same workload against the prior tip of this branch reproduced EBADF reliably when combined with #589's loopback loop.

## What's removed

- `isTransientRecvErr` + its test.
- `TestRelayWorkerLoopRetriesTransient`, `TestRelayWorkerLoopExitsOnFatal`, `TestRelayWorkerLoopDispatches` (the recv-fake-driven loop tests — replaced with socketpair-driven end-to-end tests that exercise the real RawConn path).
- `TestRecvJobReturnsEAGAINOnEmptyNonblockingSocket` (asserted a behaviour that's now correctly *absent* — EAGAIN is absorbed by the poller, not surfaced).
- The `sendMu` mutex in `runRelaySupervisor` / `acceptLoop`.
- The `workerFD` raw int + `int(workerSock.Fd())` extraction.

Closes cl-0jsq.

### DIFF
--- a/cmd/clawpatrol/relay_linux.go
+++ b/cmd/clawpatrol/relay_linux.go
@@ -533,17 +533,46 @@ func runRelayWorker(_ []string) {
 	sockFD := int(sock.Fd())
 	ignoreSIGPIPE()
 
+	relayWorkerLoop(
+		func() (uint16, int, error) { return recvJob(sockFD) },
+		handleJob,
+	)
+}
+
+// relayWorkerLoop is the recv→dispatch loop, extracted so tests can drive
+// it with a faked recv. Transient errors (EAGAIN/EWOULDBLOCK/EINTR) are
+// retried; io.EOF / supervisor shutdown returns cleanly; other errors log
+// and return.
+//
+// Previously this loop exited on any non-EOF error, which surfaced as the
+// log line "[clawpatrol relay-worker] recv: resource temporarily unavailable"
+// (EAGAIN's canonical strerror text) followed by a silent worker death and
+// the wrapped command losing its auto-expose tunnel.
+func relayWorkerLoop(recv func() (uint16, int, error), handle func(uint16, int)) {
 	for {
-		port, fd, err := recvJob(sockFD)
+		port, fd, err := recv()
 		if err != nil {
-			if errors.Is(err, io.EOF) {
-				return
+			if isTransientRecvErr(err) {
+				continue
 			}
-			fmt.Fprintf(os.Stderr, "[clawpatrol relay-worker] recv: %v\n", err)
+			if !errors.Is(err, io.EOF) {
+				fmt.Fprintf(os.Stderr, "[clawpatrol relay-worker] recv: %v\n", err)
+			}
 			return
 		}
-		go handleJob(port, fd)
+		go handle(port, fd)
 	}
+}
+
+// isTransientRecvErr reports whether err from the SCM_RIGHTS recv loop
+// represents a transient interruption that should be retried rather than
+// treated as fatal. EAGAIN/EWOULDBLOCK can occur if the worker socket is
+// (inadvertently) nonblocking; EINTR occurs when a signal arrives during
+// the syscall. None of these indicate the supervisor has gone away.
+func isTransientRecvErr(err error) bool {
+	return errors.Is(err, syscall.EAGAIN) ||
+		errors.Is(err, syscall.EWOULDBLOCK) ||
+		errors.Is(err, syscall.EINTR)
 }
 
 func recvJob(fd int) (uint16, int, error) {
--- a/cmd/clawpatrol/relay_linux.go
+++ b/cmd/clawpatrol/relay_linux.go
@@ -611,7 +611,7 @@ func relayWorkerLoop(rc syscall.RawConn, handle func(uint16, int)) {
 func recvJob(rc syscall.RawConn) (uint16, int, error) {
 	var (
 		port  uint16
-		jobFD int = -1
+		jobFD = -1
 		rerr  error
 	)
 	err := rc.Read(func(rawFD uintptr) (done bool) {
--- a/cmd/clawpatrol/relay_linux.go
+++ b/cmd/clawpatrol/relay_linux.go
@@ -566,9 +566,21 @@ func relayWorkerLoop(recv func() (uint16, int, error), handle func(uint16, int))
 
 // isTransientRecvErr reports whether err from the SCM_RIGHTS recv loop
 // represents a transient interruption that should be retried rather than
-// treated as fatal. EAGAIN/EWOULDBLOCK can occur if the worker socket is
-// (inadvertently) nonblocking; EINTR occurs when a signal arrives during
-// the syscall. None of these indicate the supervisor has gone away.
+// treated as fatal. None of these indicate the supervisor has gone away.
+//
+// EAGAIN/EWOULDBLOCK: we create the socketpair blocking (SOCK_SEQPACKET
+// only, no SOCK_NONBLOCK), but O_NONBLOCK lives on the open file
+// description, not the fd, so anything that flips it on either end is
+// visible here after inheritance. In particular, wrapping a socket fd
+// with os.NewFile on the supervisor side hands it to Go's runtime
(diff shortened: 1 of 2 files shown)


cmd/clawpatrol/relay_linux.go
```diff
@@ -533,19 +533,48 @@ func runRelayWorker(_ []string) {
...
 			return
 		}
-		go handleJob(port, fd)
+		go handle(port, fd)
 	}
 }
 
+// isTransientRecvErr reports whether err from the SCM_RIGHTS recv loop
+// represents a transient interruption that should be retried rather than
+// treated as fatal. EAGAIN/EWOULDBLOCK can occur if the worker socket is
```
arnauorriols: Justify why EAGAIN/EWOULDBLOCK would ever be returned.

arnauorriols:
[bot-reply]
Pushed fc05bac addressing this comment:
- Expanded the `isTransientRecvErr` comment to justify why EAGAIN/EWOULDBLOCK can be returned at all: the socketpair is created blocking, but `os.NewFile` on the supervisor side registers the fd with Go's runtime netpoller, which flips O_NONBLOCK on the shared open file description; the relay-worker child inherits that fd and reads it with `unix.Recvmsg`, so EAGAIN surfaces directly. The "resource temporarily unavailable" line we saw in the wild is the empirical confirmation.
- Comment-only change; no behavior change.

cmd/clawpatrol/relay_linux.go
```diff
@@ -533,19 +533,60 @@ func runRelayWorker(_ []string) {
...
+// treated as fatal. None of these indicate the supervisor has gone away.
+//
+// EAGAIN/EWOULDBLOCK: we create the socketpair blocking (SOCK_SEQPACKET
+// only, no SOCK_NONBLOCK), but O_NONBLOCK lives on the open file
+// description, not the fd, so anything that flips it on either end is
+// visible here after inheritance. In particular, wrapping a socket fd
+// with os.NewFile on the supervisor side hands it to Go's runtime
+// netpoller, which sets O_NONBLOCK on the underlying description; the
+// relay-worker child then inherits an fd that already has O_NONBLOCK
+// set and reads it via unix.Recvmsg (bypassing the runtime poller), so
```
arnauorriols: Evaluate if using the runtime poller instead of unix.Recvmsg would be beneficial or not

arnauorriols: [bot-reply]
Evaluation — sticking with `unix.Recvmsg` + EAGAIN retry, but documenting the alternatives:

**Option A (current): direct `unix.Recvmsg` + retry EAGAIN/EWOULDBLOCK/EINTR.**
Pros: simple, single syscall per frame, no callback indirection. Cons: EAGAIN triggers a busy-spin retry; the worker OS thread is parked in `recvmsg` between frames rather than handed back to the Go scheduler.

**Option B: route through the runtime poller** via `os.NewFile(fd, ...).SyscallConn().Read(func(fd uintptr) bool { ... unix.Recvmsg ...; return err != EAGAIN })`.
Pros: poller-managed sleep on `epoll_wait`, EAGAIN handled by the runtime, no manual retry for that case. Cons: extra wrapper layer; we still need EINTR handling inside the callback; relay-worker is a dedicated process whose only job is this loop, so freeing the OS thread buys us nothing here — there is no other work for the scheduler to do.

**Option C: clear `O_NONBLOCK` on the worker side** via `unix.SetNonblock(sockFD, false)` once at startup. The worker holds the only reference to its end's open file description after the parent closes its copy in `setupRelayInChild`, and the supervisor sits on the other end of the socketpair (a separate OFD), so this is safe and doesn't disturb the netpoller on the supervisor side. EAGAIN goes away entirely; only EINTR retry remains.

**Recommendation:** keep A. The runtime poller (B) adds machinery without buying anything for a single-purpose process. C is the cleanest "real fix" but it's a behavioural change that warrants its own PR — it's also one fcntl away if EAGAIN ever shows up frequently in the wild. The retry loop in this PR is small, well-commented, and correct for all three transient cases (EAGAIN/EWOULDBLOCK/EINTR), so the cost of leaving it is near zero.

arnauorriols:
make ci green

arnauorriols:
[bot-reply]
Pushed d603899 addressing this comment:
- Renamed unused `fd` parameter to `_` in `relay_linux_test.go:223` to satisfy the `revive` unused-parameter check that was failing CI lint.

PR #587, opened 2026-05-27
fix: exit clawpatrol run when relay-supervisor or relay-worker dies

arnauorriols:
## Summary

- If `relay-supervisor` or `relay-worker` died, `clawpatrol run` kept executing the wrapped command with a dead relay — webhooks would silently fail. Now an unexpected death in either tears down the wrapped child and exits non-zero with the cause on stderr.
- Supervisor monitors the worker socketpair for EOF and coordinates with the seccomp notify loop's `ENOENT` shutdown via a 500ms grace, so the normal exit cascade (user cmd → agent child → worker via Pdeathsig) stays silent.
- Top parent waits on the wrapped child and the supervisor concurrently; if the supervisor dies first, it SIGTERM/SIGKILLs the wrapped child and surfaces the supervisor's last stderr line as the cause. Last-line capture uses a bounded (4 KB) ring-buffer tee in front of `os.Stderr` so live logs still stream in real time.

Closes cl-2qgj.

## Test plan

- [x] `go build ./cmd/clawpatrol/` and `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go test ./cmd/clawpatrol/ -count=1` passes (full package)
- [x] `TestMonitorWorkerSocketEOFOnPeerClose` — real socketpair, close peer end, assert `io.EOF`
- [x] `TestLastLineWriter` — pass-through + last-line extraction across input shapes
- [x] `TestLastLineWriterBounded` — ring buffer stays bounded under noisy writes
- [ ] Manual integration: kill `relay-worker` mid-run, verify `clawpatrol run` exits non-zero, wrapped child gets terminated, stderr contains the worker-death cause (not exercised here — would require a daemon + namespaces harness this repo doesn't have)

### DIFF
--- a/cmd/clawpatrol/relay_linux.go
+++ b/cmd/clawpatrol/relay_linux.go
@@ -41,6 +41,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
@@ -195,6 +196,22 @@ func notifSendContinue(fd int, id uint64) error {
 //
 // On each listen() trap it inspects the agent's socket, opens a host-side
 // listener on the same port, and hands accepted connections to the worker.
+//
+// Two background loops run concurrently:
+//
+//   - The seccomp notify loop (the supervisor's primary job).
+//   - A worker-socket monitor: when the worker process dies, its end of
+//     the socketpair closes and recvmsg here returns EOF. That's the only
+//     way the supervisor (host netns, host pid space) can notice the
+//     worker (agent netns) has died — there's no Wait relationship.
+//
+// On clean shutdown the agent child exits, which sends Pdeathsig=SIGTERM
+// to the worker AND empties the seccomp filter's task list. Both loops
+// fire near-simultaneously: the seccomp loop returns via ENOENT, the
+// monitor returns via EOF. We give the seccomp loop a short grace window
+// — if it closes within that window the worker death is the expected
+// shutdown cascade and we exit 0; otherwise the worker died for some
+// other reason and we exit non-zero so the top parent can surface it.
 func runRelaySupervisor(_ []string) {
 	notifyFile := os.NewFile(3, "seccomp-notify")
 	workerSock := os.NewFile(4, "worker-sock")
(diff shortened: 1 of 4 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR body — it accurately describes the single commit (cb4b9071). No code change needed; verified the checkable test-plan claims on the branch:

- `go build ./cmd/clawpatrol/` — clean
- `go vet ./cmd/clawpatrol/` — clean
- `gofmt -l cmd/clawpatrol/` — clean
- `go test ./cmd/clawpatrol/ -run 'TestLastLineWriter|TestLastLineWriterBounded|TestMonitorWorkerSocketEOFOnPeerClose'` — passes

The unchecked manual-integration item (kill `relay-worker` mid-run) remains unexercised by design — it needs a daemon + namespaces harness this repo doesn't have, as noted in the body. Nothing further to address.

piscisaureus:
Reviewed against current main (rebased onto d17afc32, all 19 relay/run tests pass on Linux). I'm closing this as superseded rather than landing it, because main went the other way on the policy question after this PR was opened:

- #688 (`1f77140d`, "relay: fail open when the auto-expose relay dies") decided that a dead relay must not take the wrapped command down: the worker removes its REDIRECT on exit, the supervisor warns loudly, and the command keeps running with in-namespace loopback restored. This PR SIGTERMs the command and exits 1, which reverses that decision.
- The spontaneous relay deaths that motivated both PRs were fixed at the source: `0be38abe` (ENOTTY from a GC-finalised notify fd) and `18e593b1` (EINTR/EAGAIN retried via RawConn).
- On main a failed worker spawn or a notify-loop error is deliberately non-fatal; with this change either becomes "relay died — exiting", so it would regress those paths too.
- There is also a tie-break gap in the parent `select`: a command that exits normally can be reported as a relay failure if the supervisor's exit is observed first.

What is worth keeping is the concurrent `relaySup.Wait()`: on main a supervisor that dies early sits defunct until the command exits, and its last stderr line is lost. I'll land that part separately as a small change that reaps and logs without killing the command. Thanks for the investigation; the socketpair-EOF monitor was the right instinct for detecting the death.

PR #589, opened 2026-05-27
run: forward wrapped-cmd 127.0.0.0/8 connects to host loopback (cl-1clt)

arnauorriols:
## Summary

The auto-expose relay already mirrors agent listeners back to the host. The reverse — a wrapped command dialing `127.0.0.1:port` and reaching a host loopback service — was missing, so `clawpatrol run` couldn't be used for the common mix-remote-with-local-services workflow (psql, redis, dev servers, etc.).

Adds a host-loopback forwarder that lives in the existing relay topology:

- An `iptables nat OUTPUT REDIRECT` rule in the agent netns captures every `127.0.0.0/8:*` connect from the wrapped command and routes it to a TCP forwarder bound on the agent's loopback.
- The forwarder reads `SO_ORIGINAL_DST` (both IP **and** port — the /8 match means the original destination can be `127.0.0.2`, `127.0.1.1`, etc., not just `127.0.0.1`) and ships `(ip, port, accepted_fd)` to the supervisor over a new SOCK_SEQPACKET socketpair via SCM_RIGHTS.
- The supervisor (host netns) dials the original `127.0.0.x:port` on its own loopback and bidi-copies. A defensive `net.IP.IsLoopback()` check before dial ensures a malformed frame can never steer the host-side dial off loopback.

A `SO_MARK` / `-m mark RETURN` exemption lets the worker's own dial to the agent loopback (the existing auto-expose reverse direction) skip the REDIRECT. The wrapped command runs without `CAP_NET_ADMIN` so it can't set `SO_MARK` to bypass.

A ready-pipe blocks the user-cmd `exec` until REDIRECT is installed, eliminating the race where a fast-starting child dials `127.0.0.1` before iptables lands.

## Built on top of #585's RawConn primitive

Both new directions — worker → supervisor sends on the lb sock, supervisor recv loop — use `SyscallConn().Read`/`Write` symmetrically with the inherited auto-expose paths. RawConn carries an internal `*os.File` reference for the goroutine's lifetime, so the GC-finalizer race that surfaces as EBADF in raw-int handoffs (root cause originally diagnosed in #585) can't happen here either. The runtime poller absorbs transient `EAGAIN`/`EWOULDBLOCK`/`EINTR` without us writing explicit retry logic.

Empirical validation under load: 100 concurrent requests across 5 runs through the host-loopback forwarder, **zero relay-recv errors**. Before #585 landed, an earlier version of this PR using the int-handoff anti-pattern reproduced EBADF at ~60% rate under the same workload.

## Threat model

- **Forwarder listener** is bound to `127.0.0.1` inside the **agent** netns. The agent netns has no inbound connectivity except via the TUN to the gateway; the gateway does not tunnel traffic to the agent's loopback. Only the wrapped command (and its same-netns children) can reach the forwarder.
- **Supervisor** only dials host `127.0.0.x:port` in response to a frame arriving on the SOCK_SEQPACKET lb sock. The worker end of that sock is held only by the worker process; the wrapped command does not inherit it (it's a child of the agent child, not of the worker).
- Therefore host-loopback services are reachable from the wrapped command and nowhere else — not from anything routed via the gateway/tailnet, not from off-box clients.
- IPv6 (`::1`) is not wired up in this change; the existing relay path doesn't cover it either, so it's left as a follow-up rather than expanding scope.

## Test plan

- [x] `go test ./...` — full suite green.
- [x] `gofmt -l` clean; `go vet` clean; `golangci-lint run ./cmd/clawpatrol/` — 0 issues.
- [x] Unit tests against real socketpairs (drop into the SyscallConn-based pattern from #585):
  - `TestWorkerPIDFrameRoundTrip` — PID frame survives `sendWorkerPID` → `recvWorkerPID` round-trip via RawConn.
  - `TestLoopbackRedirectRuleArgs` — pins exact iptables argv shape (mark-RETURN + /8 REDIRECT).
  - `TestLoopbackFrameRoundTrip` — `(ip, port)` wire frame encode/decode preserves all 127.0.0.0/8 addresses.
  - `TestGetOriginalDstPortByteOrder` — endian-correct port extraction from `RawSockaddrInet4`.
  - `TestHostLoopbackForwarder_EndToEnd` — re-execs into a fresh user+net namespace, installs the real REDIRECT, asserts a `127.0.0.2:port` dial round-trips with the original dst IP intact.
- [x] End-to-end with built binary against host services bound on multiple loopback addresses:
  - `127.0.0.1:18101` → reached ✓
  - `127.0.0.2:18102` → **reached the service bound specifically to 127.0.0.2** (IP preserved end-to-end) ✓
  - `localhost:18101` → NSS resolves to 127.0.0.1, caught by /8 rule ✓
  - `127.0.1.1:18101` → forwarded to host's `127.0.1.1` (connection refused there since no listener — proper RST propagation) ✓
  - `[::1]:port` → fast-fail (no IPv6 forwarder), as scoped ✓
  - Non-loopback IPs (`8.8.8.8`) → traffic still goes via TUN→gateway, not intercepted by the forwarder (verifies the `-d 127.0.0.0/8` clamp) ✓
- [x] 5 stress runs × 20 concurrent connections each: **all passed, zero relay-recv errors**.

Closes cl-1clt.


arnauorriols:
Can we make localhost requests still go through the gateway and come back if approved? Maybe mapping localhost ips to the device's tunnel ip, so that the device is upstream as well as downstream. This should be an optional behavior gated via config. Defaults enabled, but can be disabled to prevent localhost traffic from leaving the machine

arnauorriols:
[bot-reply]
Yes, this is feasible, and it's a clean fit for the existing relay topology — but it's a meaningfully larger change than what's on this branch, and it inverts one of this PR's threat-model guarantees, so I want to confirm the design before building it rather than guess.

**Where the branch is today.** The forwarder is deliberately gateway-bypassing: wrapped cmd dials `127.0.0.1:port` → iptables `REDIRECT` in the agent netns → relay-worker → SCM_RIGHTS to the host supervisor → supervisor dials the host's `127.0.0.1:port` directly. No tunnel hop, so no policy/HITL/logging. The commit's threat-model section leans on exactly this: *"Host-loopback services are reachable from the wrapped command and nowhere else — not from anything routed via the gateway or tailnet."*

**What your proposal would take.** Instead of the supervisor dialing host loopback directly, we'd route the connection through the tunnel so the gateway mediates it:

1. **Agent side:** swap the `REDIRECT`-to-local-forwarder for a `DNAT` `127.0.0.1:port → <device-tunnel-ip>:port`. That puts the packet on `wg0` → daemon gVisor → gateway, i.e. through the same policy path every other outbound connection already takes. (Source-port/conntrack rewrite so replies find their way back.)
2. **Gateway side:** the gateway has to be willing to route a connection whose destination is the *originating device's own* tunnel IP back down to that device. Today it doesn't — peers don't get tunneled back to themselves. This is the \"device is upstream as well as downstream\" piece, and it's the part that lives outside this repo / outside `clawpatrol run`.
3. **Daemon side:** the device needs a small reverse listener bound to its tunnel IP that accepts the gateway-forwarded conn and proxies it to the real `127.0.0.1:port`. The daemon is the natural owner (it already owns the underlay identity and the gVisor stack).

**Three design calls I'd want your read on before implementing:**

- **Threat-model expansion.** Making the device a reachable upstream means localhost services become reachable *via the gateway*, not just from the wrapped command. Even if we scope it tightly (only the device's own agent sessions, gated by the gateway's policy), it's a real broadening of what this PR currently guarantees. Are you OK trading that guarantee for the policy-visibility win? I'd want the reverse listener locked to gateway-originated, this-session conns only.
- **Policy keying / approval unit.** Policy today keys on hostname→VIP. Loopback has no hostname — the gateway would see `<device-tunnel-ip>:port`. Do we approve per (device, port)? Synthesize a stable pseudo-host like `localhost.<device>` so it shows up legibly in the dashboard/HITL prompt and rules can target it? That choice drives the UX of \"comes back if approved.\"
- **Default + the config flag.** You want default-on (localhost traverses the gateway) with an opt-out that keeps traffic on-box. Note that flips this PR's current behavior: as written, #589 never lets localhost leave the machine, and \"disabled\" would be exactly today's direct-forward path. So the flag is really a 3-state intent: *gateway-mediated* (new default) / *direct on-box forward* (today's behavior) / *off entirely*. I'd wire it as an HCL config knob (e.g. `loopback = \"gateway\" | \"direct\" | \"off\"`) honored by the daemon, with a matching `clawpatrol run` flag override. Does default `\"gateway\"` match your intent, with `\"direct\"` as the escape hatch?

**Suggested sequencing.** Land #589 as the *direct on-box forward* baseline (it's self-contained, tested, and is exactly the `\"direct\"` mode), then do the gateway-mediated path as a follow-up that adds the daemon reverse-listener + the gateway dial-back + the config knob — since the gateway change is cross-component and needs its own review. I'll file that follow-up bead with this design captured.

Happy to implement either way — just confirm (a) default `\"gateway\"` vs `\"direct\"`, (b) the policy-keying approach, and (c) follow-up vs. fold-into-this-PR, and I'll build it.

arnauorriols:
Support 127.0.0.0/8, not just 127.0.0.1. 

Regarding the optional proxying through the gateway, I imagine: 
User can set an endpoint that uses "localhost" as the host. If we do nothing, this would enable the existing VIP mechanics, and would map a VIP -> localhost, and when the agent dials the VIP, if the request is approved, the gateway tries to forward the request upstream, which would result in the gateway's `localhost`. What if we special-case localhost upstreams, so that they resolve to the source device's tunnel ip? When a device looks up localhost and an endpoint (used by the device's profile) claims localhost, the VIP -> upstream mapping becomes VIP -> device tunnel IP (instead of VIP -> localhost), and when the gateway forwards the request upstream, it sends it through the tunnel to the agent's netns. In there, the request should be redirected to localhost on the netns, and by means of the supervisor, routed to the host's loopback.

arnauorriols:
[bot-reply]
Pushed 156a4f0 addressing this comment.

**127.0.0.0/8 support** — done. The agent-netns REDIRECT now matches `-d 127.0.0.0/8` instead of `-d 127.0.0.1/32`, so a wrapped command dialing any loopback address (e.g. a host service bound to `127.0.0.2`) is captured. A range match alone would misroute (`127.0.0.2:p` → host `127.0.0.1:p`), so the original destination IP is preserved end to end: `getOriginalDst` now returns the `SO_ORIGINAL_DST` address as well as the port, the worker→supervisor lb-sock frame carries IP+port (6 bytes), and the supervisor dials the recovered address. It re-checks `IsLoopback` before dialing as a defensive guard. `TestHostLoopbackForwarder_EndToEnd` now dials `127.0.0.2` (not `.1`) and asserts the supervisor recovers `127.0.0.2`. IPv6 (`::1`) remains a separate follow-up, as before.

**Optional proxying through the gateway (VIP → localhost → device tunnel IP)** — this is a separate, larger feature on the gateway/VIP path, distinct from the `clawpatrol run` host-loopback forwarder this PR adds. I've filed `cl-ukn8` to track the design (localhost-upstream declaration in policy, per-device VIP→tunnel-IP mapping, netns-side redirect, approval/dashboard interaction) rather than expanding this PR's scope. Happy to fold it in here instead if you'd prefer it land together.

arnauorriols:
Verify local names are resolved by the device's /etc/host and not the gateway's

arnauorriols:
[bot-reply]
Verified — local names (`/etc/hosts` entries) are resolved by the **device's** `/etc/hosts`, never the gateway's. No code change needed; here's the reasoning grounded in the run path:

1. **The child only replaces `/etc/resolv.conf`, never `/etc/hosts`.** `runRunChild` unshares a fresh mount namespace (`CLONE_NEWNS`, `run_linux.go:163`), whose mount table is a copy of the device's. The single filesystem override inside it is `bindResolv(childResolvConf())` (`run_linux.go:332`), which bind-mounts a generated `/etc/resolv.conf` over the existing one. Nothing touches `/etc/hosts` anywhere in the codebase — `wireguard.go:6` even records the intent ("no /etc/hosts pinning on clients"). So the wrapped command sees the device's `/etc/hosts` verbatim.

2. **`/etc/hosts` lookups are a local `files` nsswitch source — they never traverse the network.** libc resolves a name against `/etc/hosts` in-process before any `connect()`. The gateway is only ever reachable as the *nameserver* in the bind-mounted `resolv.conf` (the gateway's tailnet IP in tsnet mode, `childResolvConf` at `run_linux.go:797`). That path is hit only for names **not** in `/etc/hosts`. The split is exactly what the comment asks for: `/etc/hosts` = device-local, DNS fallthrough = gateway.

3. **This PR's loopback forwarder is purely IP-layer, so it can't introduce gateway-side resolution.** The agent-netns iptables REDIRECT matches `127.0.0.0/8` (`relay_linux.go:900`) and the original destination is recovered numerically via `SO_ORIGINAL_DST` (`getOriginalDst`, `relay_linux.go:978`). The worker→supervisor frame ships a 4-byte IP + port, and the supervisor dials that recovered numeric IP (`handleLoopbackJob`, `relay_linux.go:339`). Name→IP resolution has already happened device-side before the `connect()`; neither the forwarder nor the gateway ever sees a name.

End to end: a device `/etc/hosts` entry like `127.0.0.1 mydb` resolves device-side to `127.0.0.1`, the connect is REDIRECTed, and the supervisor dials the device's host-loopback `127.0.0.1:port` — the device's name maps to the device's own loopback service. Consistent and correct.

PR #612, opened 2026-06-01
doc: design for localhost upstream → source-device tunnel IP (cl-ukn8)

arnauorriols:
## Summary

Design proposal for the gateway/VIP-path counterpart to PR #589's
run-time host-loopback forwarder (cl-ukn8). **Design only — no code
behavior change.** Adds `doc/localhost-upstream-tunnel-design.md`.

### The bug it addresses
An endpoint that declares a bare loopback upstream (`host = "localhost:5432"`
with no tunnel) gets a DNS-VIP (loopback hostnames are treated as
resolvable). On dispatch the plugin dials the upstream via the gateway's
**host network**, so `localhost` resolves to the *gateway's* own
loopback — not the agent's. Confirmed for postgres (`postgres.go:271`)
and ssh (`ssh.go:151`), both funneling through the single `DialUpstream`
closure (`main.go:1736`).

### What the doc settles
- **Fix shape:** reaching the source device's own loopback is **opt-in
  and explicit** via the built-in `tunnel = tunnel.clawpatrol.self`
  reference. When such an endpoint is dispatched, the gateway late-binds
  at the `DialUpstream` choke point, rewriting the target to `peerIP(c)`
  (the source device's tunnel IP, already captured at `main.go:1710`)
  and dialing back **through the netstack** (tsnet `ts.Dial` / WG gVisor
  netstack), never the host dialer. No persistent per-device VIP
  allocation.
- **Open question 1 (HCL) — resolved:** disambiguate the three meanings
  of a loopback `host` explicitly rather than guessing from the hostname:
  - `host = "localhost:..."` with **no tunnel** → **compile error**
    (silently dialed the *gateway's* loopback — never the intent).
  - `host = "localhost:..."` + a **real tunnel** (`local_command.*`,
    kubectl-portforward-ssh, …) → **allowed, no rewrite**; lands on the
    loopback at the tunnel's far end. Common pattern, left untouched.
  - `host = "localhost:..."` + `tunnel = tunnel.clawpatrol.self` →
    **allowed**, rewrites to the source device's tunnel IP (`peerIP(c)`).
- **Open question 2 (lifecycle) — resolved:** no per-device VIP; resolve
  to the tunnel IP per connection at dial time. Netns redirect reuses
  PR #589's reverse-relay, widened to also accept connections arriving
  on the tunnel IP.
- **Open question 3 (approval/dashboard) — resolved:** approval chain
  unchanged (the rewrite is post-decision); dashboard renders
  `localhost (→ device)`.
- **Threat model:** `self` is deliberately device-agnostic — it binds to
  the live VIP connection's authenticated WG peer (`pip`), so a device
  can only ever reach its *own* loopback, never another device's.
  Netstack-only dial constrains the target to the tailnet. Cross-device
  loopback sharing is a non-goal (use Tailscale).

### Dependency / staging
This is the gateway counterpart to PR #589's host-loopback forwarder,
which is **not yet merged to main**. The doc stages the work:
- **cl-ukn8a** — parse `tunnel.clawpatrol.self` + reject bare-loopback /
  reject self+real-tunnel combo [shippable now, #589-independent];
- **cl-ukn8b** — gateway netstack dial-back;
- **cl-ukn8c** — device-side reverse-relay tunnel-IP capture [needs #589];
- **cl-ukn8d** — dashboard rendering.

### DIFF
--- a/doc/localhost-upstream-tunnel-design.md
+++ b/doc/localhost-upstream-tunnel-design.md
@@ -6,13 +6,19 @@ at time of writing lives on a side branch and is **not merged to main**.
 
 ## Problem
 
-An endpoint may legitimately declare a loopback upstream:
+A user who wants to reach a service on their *own* host naively writes a
+loopback upstream:
 
 ```hcl
 endpoint "local_pg" "postgres" {
   host = "localhost:5432"   # or "127.0.0.1:5432"
 }
 ```
+
+This looks reasonable but silently dials the *gateway's* loopback, not
+the author's (traced below). The Proposal makes this bare form a compile
+error and offers `tunnel = tunnel.clawpatrol.self` as the explicit,
+correct spelling.
 
 `localhost` is a hostname, not an IP literal, so it is treated as a
 resolvable host and claims a DNS-VIP at policy build
@@ -44,16 +50,27 @@ same "reach the agent host's loopback" problem for the
 
 ## Proposal
 
-When a device's profile uses an endpoint whose chosen upstream host is
-loopback, rewrite the gateway's dial target from `localhost:port` to
-**the source device's tunnel IP**, and dial it **back through the
-netstack** (not the host dialer). On the device side, the existing
-PR #589 reverse-relay redirects that inbound `tunnelIP:port` connection
-to the device's host loopback.
+Reaching the source device's own loopback is **opt-in and explicit**:
+the endpoint declares the built-in `tunnel = tunnel.clawpatrol.self`
+reference. A bare `host = "localhost:5432"` with **no** tunnel is a
+compile error — the bare form silently dialed the *gateway's* loopback,
+which is never what the author meant (see Open question 1). When an
+endpoint bound to `tunnel.clawpatrol.self` is dispatched, the gateway
+rewrites its dial target to **the source device's tunnel IP** and dials
+it **back through the netstack** (not the host dialer). On the device
+side, the existing PR #589 reverse-relay redirects that inbound
+`tunnelIP:port` connection to the device's host loopback.
+
+```hcl
+endpoint "local_pg" "postgres" {
+  host   = "localhost:5432"
+  tunnel = tunnel.clawpatrol.self
+}
+```
 
 ```
 agent app ──dial localhost:5432──▶ DNS-VIP ──▶ gateway VIP conn
-gateway: upstream host is loopback
+gateway: endpoint tunnel is tunnel.clawpatrol.self
       └─ rewrite target → <sourceDeviceTunnelIP>:5432
       └─ dial back THROUGH the netstack (ts.Dial / wg netstack), not host net
 device netns: inbound on tunnelIP:5432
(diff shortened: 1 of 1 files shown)


arnauorriols:
looback reached via tunnel should be allowed. Reaching a loopback via tunnel is in fact a common pattern for accessing services not exposed to internet. In fact, maybe the solution that uses the device's tunnel should be explicit in the hcl? By that I mean, maybe the clawpatrol tunnel through which that specific "localhost" is reached should be an hcl tunnel block:
```
endpoint "local_pg" "postgres" {
  host = "localhost:5432"
  tunnel = tunnel.clawpatrol.self
}
```

Keep in mind that `tunnel.clawpatrol.self` is not explicit about the device deliverately. I don't want the device's localhost to become reachable by any other device in clawpatrol. Users can use tailscale for that.

And the thing that would actually NOT be allowed, would be localhost without a tunnel.

arnauorriols:
[bot-reply]
Good call — you're right that loopback-over-a-tunnel is the common case (a service bound to 127.0.0.1 on the far end), and the earlier draft was wrong to reject it. Pushed d56bdb9 reworking the design around your `tunnel.clawpatrol.self` proposal:

- **Inverted Open question 1.** The source-device loopback rewrite is now **opt-in and explicit** via the built-in `tunnel = tunnel.clawpatrol.self` reference. Three cases are now spelled out in a table:
  - `host = "localhost:..."` with **no tunnel** → **compile error** (silently dialed the *gateway's* loopback — never the intent).
  - `host = "localhost:..."` + a **real tunnel** (`local_command.*`, kubectl-portforward-ssh, …) → **allowed, no rewrite**; lands on the loopback at the tunnel's far end. This is the common pattern, untouched.
  - `host = "localhost:..."` + `tunnel = tunnel.clawpatrol.self` → **allowed**, rewrites to the source device's tunnel IP (`peerIP(c)`).
- **`self` is deliberately device-agnostic**, exactly as you noted: it binds to the live VIP connection's peer, so a device can only ever reach its *own* loopback — never another device's. Cross-device loopback sharing is called out as a non-goal (use Tailscale). Captured in the Proposal + Threat-model sections.
- **Implementation sketch** now triggers on a compile-time `ep.SelfTunnel` flag instead of sniffing whether the upstream string is loopback; real tunnels fall through to `g.dialThrough` unchanged.
- **Staging** updated: cl-ukn8a is now "parse `tunnel.clawpatrol.self` + reject bare-loopback / reject self+real-tunnel combo" instead of "reject loopback+tunnel".

One thing I kept as the literal spelling you wrote (`tunnel.clawpatrol.self`, three parts) and flagged as a **built-in/reserved** reference, since the existing grammar is two-part `<type>.<name>`. If you'd rather it be a real `tunnel "clawpatrol" "self" {}`-style block (or a different reserved spelling), say the word and I'll adjust.

arnauorriols:
[bot-reply]
Synced the PR description to the current design. The body still described the **superseded** draft (Open question 1 = "reject loopback + explicit tunnel"), which was already reworked in d56bdb9 per @arnauorriols's `tunnel.clawpatrol.self` feedback. The doc (`doc/localhost-upstream-tunnel-design.md`) and the body now agree:

- Open question 1 is now the explicit, opt-in `tunnel = tunnel.clawpatrol.self` model, with the three-case table (bare loopback → compile error; loopback + real tunnel → allowed, no rewrite; loopback + `tunnel.clawpatrol.self` → allowed, rewrite to `peerIP(c)`).
- Open questions 2 & 3 marked resolved (per-connection resolve, no per-device VIP; approval unchanged + `localhost (→ device)` dashboard rendering).
- Staging relabelled to cl-ukn8a..d.

No code/doc change in this push — body text only.

piscisaureus:
Closing during a backlog triage pass. Design-only documents should not sit as open PRs; the content remains on the branch for reference and can come back as an implementation PR if the work is picked up.

piscisaureus:
Reopening: closing this in the triage pass was premature. It was opened after the public launch and may well be a good addition; it will get a review on its merits rather than being closed for age.

PR #615, opened 2026-06-01
[codex] make generated host blocks explicit

ry:
Stacked on #614.\n\nThis follow-up keeps passthrough host blocking understandable from the HCL itself:\n\n- Adds endpoint-level `deny_profiles` for credentialless catch-all deny endpoints.\n- Generated passthrough host blocks now include `deny_profiles = [profile.default]` instead of relying on implicit compiler routing.\n- Adds a clearer post-apply success state in the modal with a link to the current config.\n- Documents `deny_profiles` in the generated config reference.\n\nValidation run on the stacked branch:\n- `go test ./cmd/clawpatrol -run 'TestGenerateRule|TestGeneratedSpliceHostBlock'`\n- `go test ./internal/config ./internal/tools/docgen`\n- `cd dashboard && deno task format:check && npm run build`

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -337,6 +337,7 @@ func newUpstreamDialer(resolver string) *net.Dialer {
 type Gateway struct {
 	cfg      *config.Gateway
 	cfgPath  string // path the HCL config was loaded from
+	configMu sync.Mutex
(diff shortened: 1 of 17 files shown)


ry:
Superseded by #614 after rebasing the action-driven rule generation work on top of #617. The useful success-modal behavior was folded into #614; the rejected deny_profiles config model is not being carried forward.

PR #618, opened 2026-06-01
Add ssh-family security rules

piscisaureus:
Adds an `ssh` rule family so SSH endpoints — which previously only
emitted an `allow` at session start — can be governed by the same
rule engine as http/sql/k8s. Three layers, smallest to deepest:

## Envelope facet

* New `ssh` facet exposes `ssh.verb` (`pty` / `exec` / `shell` /
  `subsystem` / `forward`), `ssh.command`, `ssh.subsystem`,
  `ssh.forward_host`, `ssh.forward_port`, `ssh.user`. The rule engine
  infers family from the endpoint, so rules targeting an ssh endpoint
  work once the facet is registered — no rules-engine changes
* The ssh endpoint evaluates each channel action through
  MatchRequest before forwarding it upstream, mirroring the postgres
  per-statement path (allow / deny / approve, default-deny when HITL
  is unconfigured). A denied exec / shell / subsystem request is
  refused and that channel torn down without dropping the rest of the
  SSH connection; a denied direct-tcpip forward is rejected at
  channel-open

## pty gating ("block interactive")

* Deny `ssh.verb == 'pty'` to block interactive terminals. The
  pty-req is the wire signal a session wants a terminal; denying it
  tears the session channel down before any shell or exec runs, so
  both `ssh host` and `ssh -t host bash` are refused. `ssh.verb ==
  'shell'` is NOT a robust interactive control (an exec'd shell
  bypasses it); `shell` stays as an auditable verb
* `ssh.command` is documented as best-effort / audit (the agent picks
  the invocation), not a hard boundary

## stdin pre-gating

* `ssh.stdin` exposes the buffered client→server stdin of a
  shell/exec session — the body of `ssh host < deploy.sh`. Rules can
  match it (`ssh.stdin.contains(...)`) or route it to an approver
  chain for an LLM judge
* Pre-gate by withholding: the gateway buffers the stdin and holds it
  from the upstream shell until the verdict, so a denied script never
  runs (the remote `read()` blocks until allow). The shell/exec is
  acked to the agent up front so clients that wait on the reply still
  send stdin; the gate is on the upstream forward, not the ack.
  Buffering is bounded by a cap, an idle timeout, and EOF; overflow
  fail-closes any ssh.stdin rule
* Lazy: a compile-time `CompiledEndpoint.InspectsTruncatable` flag
  keeps the channel splice byte-for-byte unchanged when no rule reads
  stdin. Only no-pty session channels on a stdin-inspecting endpoint
  take the buffered path; pty/subsystem bail to the envelope path.
  The shared splice helpers were extracted so `proxyChannel` is
  unchanged and the new `proxySessionStdinGated` reuses the
  graceful-close invariant

## Tooling, docs, tests

* "Download action" exports ssh events (incl. stdin) as fixtures and
  `clawpatrol test` replays them; adds an ssh endpoint + rules +
  fixtures to the testdata corpus
* Documented in `site/doc/rules.md` and `site/doc/clawpatrol-test.md`
* End-to-end tests drive both the `x/crypto/ssh` client and the real
  OpenSSH binary (skipped when absent) through the gateway into an
  in-process upstream: stdin allow forwards bytes, deny never reaches
  upstream, command rules still pre-gate on the stdin path, pty
  bails, and the no-stdin fast path is unchanged

## Scope / known limits

* The facet gates the channel envelope and bounded piped stdin, not
  the interactive byte stream of an open terminal. A no-`-t` `ssh
  host bash` (dumb shell over stdin, no pty) and streamed stdin past
  the cap are not pre-judged — documented inline


cmd/clawpatrol/testdata/example.hcl
```diff
@@ -38,4 +36,57 @@ rule "github-writes" {
...
+  reason    = "file transfer is not permitted on the build host"
+}
+
+// Block git pushes (receive-pack) while leaving fetches (upload-pack)
+// to the catch-all exec allow below. This blocks ALL pushes, not just
+// force pushes — force-vs-normal isn't visible at the command level
+// (see site/doc/rules.md, ssh family scope).
+rule "ssh-no-push" {
+  endpoint  = ssh.build-host
+  condition = "ssh.verb == 'exec' && ssh.command.startsWith('git-receive-pack')"
```
piscisaureus: Is this command really secure. Seems that it would be circumventable using `exec /path/to/git-receive-pack`.

cmd/clawpatrol/testdata/example.hcl
```diff
@@ -38,4 +36,57 @@ rule "github-writes" {
...
+rule "ssh-no-sftp" {
+  endpoint  = ssh.build-host
+  condition = "ssh.verb == 'subsystem' && ssh.subsystem == 'sftp'"
+  verdict   = "deny"
+  reason    = "file transfer is not permitted on the build host"
+}
+
+// Block git pushes (receive-pack) while leaving fetches (upload-pack)
+// to the catch-all exec allow below. This blocks ALL pushes, not just
+// force pushes — force-vs-normal isn't visible at the command level
```
piscisaureus: Don't mention force pushes in this PR or the code. That was something I hoped it'd be able to do but can't. Seems otherwise not so relevant for ssh rule enforcement.

cmd/clawpatrol/testdata/example.hcl
piscisaureus: Agreed — command-prefix matching is circumventable (full paths, wrappers), so shipping it as if it were a hard control was wrong. Removed the git-receive-pack rule entirely. `ssh.command` stays but is now documented as best-effort / audit only.

More importantly this is the same weakness the deny-shell rule had — an exec'd shell (`ssh host bash`) bypasses `ssh.verb == 'shell'`. So I added pty-req gating: denying `ssh.verb == 'pty'` tears the session channel down before any shell or exec runs, which blocks both `ssh host` and `ssh -t host bash`. That is now the headline "block interactive" example.

cmd/clawpatrol/testdata/example.hcl
piscisaureus: Done — removed all force-push mentions from the code comments, docs (rules.md, clawpatrol-test.md), example.hcl, and fixtures.

PR #627, opened 2026-06-03
fix: stop k8s tunnel jump pods leaking (cl-acxv)

arnauorriols:
## Problem

`kubernetes_port_forward` template jump pods (`clawpatrol-rds-jump-*`) accumulate in the cluster for days instead of being deleted when their tunnel closes.

## Root cause

**Primary — stale EKS bearer at cleanup.** `writeKubeconfig` mints the EKS bearer once at Open. That bearer embeds a presigned STS URL valid for only ~60s (`eksPresignMiddleware` sets `X-Amz-Expires=60`, `aws.go:142`). `kubectl port-forward` consumes it once at handshake and is fine, but `cleanupCreatedPod` makes a *fresh* `kubectl delete` at Close — minutes-to-days later — reusing the now-expired bearer. EKS returns 401, the error was logged and swallowed, and the pod leaks. This fires on essentially every tunnel close, explaining pods of all ages.

**Secondary — crash/SIGKILL gap.** `CloseAll` only runs on SIGINT/SIGTERM (`main.go:3340`); a SIGKILL/OOM/panic orphans created pods, and no startup sweep existed to catch them.

## Tunnel lifecycle

This is the full lifecycle of a `template`-mode jump pod and the EKS bearer that authenticates every kubectl call against it. The leak lived in the gap between **mint-once at Open** and **delete-much-later at Close**; the fix threads a re-mint through every late call and adds a boot-time sweep for the crash path.

**1. Open** (`Open`, first request that acquires the tunnel)
   - `newRuntime` → `writeKubeconfig` mints a fresh EKS bearer and writes a self-contained per-tunnel kubeconfig (server + CA + bearer) to a `0600` temp file. It also returns a `reauth(ctx)` closure that re-mints the bearer and rewrites that file in place.
   - `resolveTarget` → `applyAndWait` stamps the manifest with `clawpatrol.dev/managed-by=clawpatrol` + `clawpatrol.dev/tunnel=<name>`, runs `kubectl create`, records the resolved pod name in `createdPod` (only when `cleanup != "keep"`), and waits for `condition=Ready`.
   - `startPortForward` launches `kubectl port-forward` in its own process group and parses the bound local port from stdout. The bearer is consumed here, at handshake — within ~60s of the mint — so the long-lived forward never needs the token again.
   - If `resolveTarget` or `startPortForward` fails, Open unwinds immediately: `cleanupCreatedPod` then `removeKubeconfig`, so a half-open tunnel leaves nothing behind.

**2. Serving** (steady state)
   - `Dial` hands out connections to the local `127.0.0.1:<port>` listener. No kubectl calls, no token use — the `port-forward` child holds the session open. The bearer minted at Open has long since expired by now, and that is fine until something needs a *fresh* kubectl call.

**3. Close** (graceful, tunnel released — keepalive expiry, reload, or per-request teardown)
   - Guarded by `sync.Once`. `killPF` SIGTERMs the port-forward process group, then `cleanupCreatedPod` deletes the pod.
   - **This is where the fix lives:** `cleanupCreatedPod` calls `reauth(ctx)` to re-mint the bearer *before* the `kubectl delete`, because the Open-time token is long expired. The delete now authenticates with a fresh ~60s token.
   - The delete error is **returned, not swallowed** — `Close` propagates it, and `createdPod` is cleared **only on success**, so a failed delete stays observable in logs and remains a target for the next boot's sweep (step 5).
   - `removeKubeconfig` deletes the temp kubeconfig.

**4. CloseAll** (daemon graceful shutdown, SIGINT/SIGTERM — `main.go:3340`)
   - `installGatewayShutdown` drives `TunnelManager.CloseAll`, which calls `Close` (step 3) on every live tunnel. Every created pod gets the re-mint-then-delete treatment.
   - **Gap this does not cover:** SIGKILL / OOM / panic never reaches this handler, so those pods are orphaned with no live owner. That is what step 5 mops up.

**5. Startup reconcile** (`ReconcileOrphans`, once at boot — `main.go:2967`)
   - Wired only at boot, in the background (a slow/unreachable cluster can't stall startup), and **never on reload** — a reload runs with live tunnels that legitimately own pods a sweep would wrongly delete.
   - For each `template` tunnel with `cleanup != "keep"`: `newRuntime` mints a fresh bearer, then it lists pods by the managed-label selector and deletes each one **by name from that snapshot**. A pod created after the list (a fresh `generateName`) can't appear in the snapshot, so a concurrently-opening tunnel is never torn out from under itself.
   - Default-on, so it catches both the crash path **and** any pods that leaked before this fix shipped.

**Bearer-refresh invariant:** every kubectl call that runs *after* Open's ~60s mint window — the Close delete and the reconcile sweep — re-mints first. `reauth` re-fetches the credential secret on each mint, so credential rotation is picked up without re-Opening. Plugins using an external kubeconfig (no `server` set) have `reauth == nil` and authenticate on their own; the re-mint is a no-op for them.

## Fix

1. **Re-mint at cleanup.** `writeKubeconfig` now returns a `reauth(ctx)` closure that re-mints a fresh bearer and rewrites the kubeconfig in place. `cleanupCreatedPod` calls it before the delete.
2. **Surface failures.** `cleanupCreatedPod` returns the delete error (no longer swallowed); `Close` propagates it; `createdPod` is cleared only on success so a failed delete stays observable.
3. **Labels + startup reconciliation.** Template pods are stamped `clawpatrol.dev/managed-by=clawpatrol` + `clawpatrol.dev/tunnel=<name>`. New `TunnelReconciler` interface; `ReconcileOrphans` lists labeled pods and deletes each by name from the snapshot (safe vs concurrent Open). Wired at boot only — never on reload. Default-on unless `cleanup = "keep"`.

## Tests

Label injection, reauth-before-delete ordering, delete-error surfacing, `Close` propagation, reconcile-deletes-labeled-pods, reconcile-skips-non-template/keep. Doc regenerated via docgen.

Fixes cl-acxv.


arnauorriols:
Clarify in the PR description the tunnel lifecycle in detail

arnauorriols:
[bot-reply]
Expanded the PR description with a **Tunnel lifecycle** section walking the full lifecycle of a `template`-mode jump pod and its EKS bearer:

1. **Open** — mint bearer + write per-tunnel kubeconfig (+ `reauth` closure), label and create the pod, start `port-forward` (consumes the token at handshake within ~60s of the mint).
2. **Serving** — `Dial` only; no kubectl calls, token sits expired and unused.
3. **Close** — re-mint *before* `kubectl delete` (the fix), error surfaced not swallowed, `createdPod` cleared only on success.
4. **CloseAll** — graceful SIGINT/SIGTERM path; calls `Close` on every tunnel. Notes the SIGKILL/OOM/panic gap it can't cover.
5. **Startup reconcile** — boot-only labeled-pod sweep that mops up the crash path (and pre-fix leaks), delete-by-name from a snapshot so concurrent Opens are safe.

Plus a bearer-refresh invariant: every kubectl call after Open's ~60s window re-mints first, and `reauth` re-fetches the secret so credential rotation is picked up without re-Open.

Description-only change — no code touched.

PR #630, opened 2026-06-03
site: replace hero flow diagram with demo screencast

ry:
## Summary
- Swap the static `FlowDiagram` for an autoplaying, looping screencast of the `clawpatrol` CLI above the fold.
- Re-encode the recording (strip audio, H.264 CRF 28, faststart): 1.4MB → 316KB.
- Shift the hero grid from 50/50 to 5fr/7fr so the video column is wide enough that the terminal text is readable.

## Test plan
- [ ] Load `/` locally — video autoplays, loops, no audio.
- [ ] Text in the screencast is legible at desktop widths.
- [ ] Mobile: video stacks below the copy, no horizontal overflow.


ry:
closed in favor of #635

PR #640, opened 2026-06-08
Add Linux tsnet UDP relay for clawpatrol run

dhruvkelawala:
## Summary

Adds a first-class UDP relay for Linux `clawpatrol run` in tsnet mode. Embedded tsnet exposes a TCP fallback handler but not an equivalent arbitrary-UDP fallback, so child UDP flows now use a Claw Patrol-owned framed stream over tailnet TCP to the gateway instead of relying on unsupported tsnet exit-node UDP routing.

## Architecture

- The Linux run daemon still captures child UDP packets from the per-process gVisor TUN and preserves the child-facing UDP contract.
- For tsnet transports, each child UDP flow opens a bounded tailnet TCP stream to the gateway UDP relay listener.
- The relay hello carries:
  - original UDP destination IP/port,
  - daemon tsnet peer address,
  - the per-peer API token minted during onboarding when needed for first-contact placeholder promotion.
- The gateway accepts relay streams only from already-known onboard peers or peers presenting a valid per-peer onboarding token.
- Placeholder-token promotion mirrors `/api/peer/tsnet/register`, including hostname/profile preservation, while verifying any client-claimed peer address against Tailscale WhoIs before trusting it.
- Gateway UDP dispatch matches existing WG semantics:
  - UDP/53 goes to `dnsvip`.
  - Other UDP goes through existing `relayUDP`.
- TCP forwarding, WG-mode UDP behavior, dashboard/onboarding flows, and existing tsnet state/config are preserved.

## Security notes

- The relay is not an unauthenticated open UDP proxy: unregistered peers without a valid per-peer token are rejected, even when a default profile exists.
- Relay dials and handshakes are bounded to avoid blocking the per-session TUN reader on old gateways, missing listeners, or ACL issues.
- Non-DNS relay streams have server-side idle deadlines that are refreshed by traffic in either direction.
- Zero-length UDP datagrams are preserved.

## Tests / verification run locally

- `autoreview --mode local` → clean, no accepted/actionable findings.
- `git diff --check`
- `go test ./... -count=1`
- `make build`

Focused coverage added for:

- unauthorized relay peer rejection
- malformed relay hello rejection
- unregistered peer rejection even with default profile
- per-peer-token placeholder promotion
- unverified claimed-peer rejection/canonicalization
- registered peer default-profile fallback
- server-side relay idle timeout
- idle deadline refresh from downstream traffic
- zero-length UDP datagrams
- DNS over UDP through relay
- non-DNS UDP echo round-trip
- daemon UDP flow cleanup
- `/api/peer/tsnet/register` hostname-only row guard

## Manual gateway validation pending

This PR needs a matched gateway/client deployment for live validation because the relay is a client+gateway protocol change. The branch has been pushed so the gateway can be upgraded manually, preserving existing config/state. After the gateway is running this branch, validate:

- DNS canaries without `/etc/hosts` wrapper and without `RES_OPTIONS=use-vc`.
- Arbitrary UDP through `clawpatrol run` against a controlled UDP echo server.
- Codex/GitHub DNS canary.

## Rollback

The change is binary-only and preserves existing state/config. To roll back, restore the pre-upgrade gateway/client binaries and restart their existing supervisors. Do not rerun `clawpatrol join` and do not replace gateway state.

### DIFF
--- a/cmd/clawpatrol/daemon_session_linux.go
+++ b/cmd/clawpatrol/daemon_session_linux.go
@@ -35,6 +35,8 @@ import (
 // child-side TUN doesn't need to cap.
 const runStackTunMTU = 65535
 
+var runUDPFlowIdleTimeout = 30 * time.Second
+
 // newRunStack creates a gVisor TCP/IP stack bound to localIP, which
 // is the transport's underlay address (tsnet 100.x.x.x or wg /32).
 // Promiscuous + spoofing enabled so the stack accepts inbound
@@ -212,7 +214,7 @@ func (f *runUDPForwarder) handle(pkt []byte) {
 func (f *runUDPForwarder) readResponses(conn net.Conn, srcIP, dstIP [4]byte, srcPort, dstPort uint16) {
 	buf := make([]byte, 65535)
 	for {
-		_ = conn.SetReadDeadline(time.Now().Add(30 * time.Second))
+		_ = conn.SetReadDeadline(time.Now().Add(runUDPFlowIdleTimeout))
 		n, err := conn.Read(buf)
 		if err != nil {
 			return
--- a/cmd/clawpatrol/daemon_session_linux_test.go
+++ b/cmd/clawpatrol/daemon_session_linux_test.go
@@ -1,0 +1,156 @@
+//go:build linux
+
+package main
+
+import (
+	"context"
+	"net"
+	"net/netip"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestRunUDPForwarderCleansIdleFlowAfterReadTimeout(t *testing.T) {
+	oldTimeout := runUDPFlowIdleTimeout
+	runUDPFlowIdleTimeout = 10 * time.Millisecond
+	t.Cleanup(func() { runUDPFlowIdleTimeout = oldTimeout })
+
+	transport := &cleanupTestTransport{conn: newCleanupTestConn()}
+	f := &runUDPForwarder{
+		transport: transport,
+		flows:     map[udpFlowKey]net.Conn{},
+	}
+	pkt := buildUDPPacket([4]byte{100, 98, 20, 53}, [4]byte{192, 0, 2, 10}, 53000, 9999, []byte("ping"))
+	f.handle(pkt)
+
+	if got := transport.conn.writes(); got != 1 {
+		t.Fatalf("transport writes = %d, want 1", got)
+	}
+	deadline := time.Now().Add(time.Second)
+	for time.Now().Before(deadline) {
+		f.mu.Lock()
+		flows := len(f.flows)
+		f.mu.Unlock()
+		if flows == 0 {
+			if !transport.conn.closed() {
+				t.Fatal("flow was removed before conn closed")
+			}
+			return
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+	f.mu.Lock()
+	flows := len(f.flows)
+	f.mu.Unlock()
+	t.Fatalf("flow count = %d after timeout, want 0", flows)
+}
+
+type cleanupTestTransport struct{ conn *cleanupTestConn }
+
+func (t *cleanupTestTransport) Dial(context.Context, string, string) (net.Conn, error) {
+	return t.conn, nil
+}
+func (t *cleanupTestTransport) LocalAddr() netip.Addr {
+	return netip.MustParseAddr("100.98.20.53")
+}
+
+func (t *cleanupTestTransport) BootWarning() string {
+	return ""
+}
+
+func (t *cleanupTestTransport) Close() error {
+	return nil
+}
+
+type cleanupTestConn struct {
+	mu       sync.Mutex
+	deadline time.Time
+	writeN   int
+	closedV  bool
+}
+
+func newCleanupTestConn() *cleanupTestConn {
+	return &cleanupTestConn{}
+}
+
+func (c *cleanupTestConn) Read([]byte) (int, error) {
+	for {
+		c.mu.Lock()
+		deadline := c.deadline
+		closed := c.closedV
+		c.mu.Unlock()
+		if closed {
+			return 0, net.ErrClosed
+		}
+		if !deadline.IsZero() && time.Now().After(deadline) {
+			return 0, timeoutErr{}
+		}
+		time.Sleep(time.Millisecond)
+	}
+}
+
+func (c *cleanupTestConn) Write(b []byte) (int, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.closedV {
+		return 0, net.ErrClosed
+	}
+	c.writeN++
+	return len(b), nil
+}
+
+func (c *cleanupTestConn) Close() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.closedV = true
+	return nil
+}
+func (c *cleanupTestConn) LocalAddr() net.Addr {
+	return net.UDPAddrFromAddrPort(netip.MustParseAddrPort("100.98.20.53:53000"))
+}
+
+func (c *cleanupTestConn) RemoteAddr() net.Addr {
+	return net.UDPAddrFromAddrPort(netip.MustParseAddrPort("192.0.2.10:9999"))
+}
+
+func (c *cleanupTestConn) SetDeadline(t time.Time) error {
+	_ = c.SetReadDeadline(t)
+	return nil
+}
+
+func (c *cleanupTestConn) SetWriteDeadline(time.Time) error {
+	return nil
+}
+func (c *cleanupTestConn) SetReadDeadline(t time.Time) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.deadline = t
+	return nil
+}
+func (c *cleanupTestConn) writes() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.writeN
+}
+func (c *cleanupTestConn) closed() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.closedV
+}
+
+type timeoutErr struct{}
+
+func (timeoutErr) Error() string {
+	return "i/o timeout"
+}
+
+func (timeoutErr) Timeout() bool {
+	return true
+}
+
+func (timeoutErr) Temporary() bool {
+	return true
+}
+
+var _ net.Error = timeoutErr{}
--- a/cmd/clawpatrol/daemon_transport_tsnet_linux.go
+++ b/cmd/clawpatrol/daemon_transport_tsnet_linux.go
@@ -3,11 +3,11 @@
 package main
 
 // tsnetTransport: daemonTransport implementation backed by an embedded
-// tsnet.Server with the gateway pinned as its exit node. Every Dial
-// returns a tsnet.Server.Dial conn, which the tailnet routes through
-// the gateway — the gateway then picks it up in its tsnet
-// RegisterFallbackTCPHandler with the original dst intact (no
-// PROXY-v1 framing).
+// tsnet.Server with the gateway pinned as its exit node. TCP dials use
+// tsnet.Server.Dial and land in the gateway's tsnet
+// RegisterFallbackTCPHandler with the original dst intact. UDP dials use
+// Claw Patrol's framed tailnet relay because tsnet exposes no arbitrary-UDP
+// fallback equivalent to RegisterFallbackTCPHandler.
 
 import (
 	"context"
@@ -25,11 +25,16 @@ import (
 
 type tsnetTransport struct {
 	s           *tsnet.Server
+	gatewayAddr netip.Addr
+	apiToken    string
 	localAddr   netip.Addr
 	bootWarning string
 }
 
 func (t *tsnetTransport) Dial(ctx context.Context, network, addr string) (net.Conn, error) {
+	if strings.HasPrefix(network, "udp") {
+		return dialTsnetUDPRelay(ctx, t.s.Dial, t.gatewayAddr, t.localAddr, t.apiToken, addr)
+	}
 	return t.s.Dial(ctx, network, addr)
 }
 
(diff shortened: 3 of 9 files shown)


piscisaureus:
Reviewed the diff and compared the UDP path against the macOS transparent-proxy path, since that's the other place we deal with tsnet + UDP.

## Overall

Solid. The framing is clean, the auth model is right (the source is tsnet-authenticated via `RemoteAddr`/`peerIP`, and first-contact promotion is gated by the per-peer token with a WhoIs check of the client-claimed peer before it's trusted), and it reuses `relayUDP` / `dnsvip.ServeUDP`, which already own conn close + goroutine teardown — both `defer c.Close()`, so the gateway handler doesn't leak even though it doesn't close `raw` itself. Client side cleans up too (`readResponses` → `delete(flows)` + `conn.Close()`). Good coverage on the security/edge cases.

## macOS comparison (worth surfacing)

macOS does **not** use a relay for this. In the transparent-proxy / tsnet path, non-DNS UDP rides `s.Dial(ctx, "udp", dst)` over exit-node routing (`macos/netstack/wgnetstack.go`, the `ts_netstack_udp_connect` path — comment there is explicit: *"tsnet supports UDP natively (Dial 'udp'), so no framing or TCP relay needed"*), and only DNS/53 is special-cased: the extension rewrites the DNS destination to the gateway's tsnet IP so it lands on the real `:53` listener (`Provider.swift`, ~L41-48 spells out why — arbitrary UDP over the exit node hits the gateway's missing UDP fallback and *"silent black hole"*s).

Two takeaways:

- **The premise checks out.** The macOS code corroborates that the gateway has no arbitrary-UDP exit-node fallback — only `RegisterFallbackTCPHandler` plus the DNS-only `GetUDPHandlerForFlow` catch-all (`tailscale.go` ~L253-292). So framing arbitrary UDP over a gateway-owned TCP listener is a legitimate fix, not a workaround for something tsnet already offers.
- **It diverges the platforms.** After this lands, Linux tsnet gets working arbitrary UDP (and DNS through the relay), while macOS tsnet still black-holes non-DNS UDP and only handles DNS via the dst-rewrite hack. Since `:45353` is a gateway-side, client-agnostic protocol, the darwin path could point at the same relay in a follow-up — that would let us delete the macOS DNS-rewrite special-case and have one UDP path for both. Worth at least a tracking issue so they don't silently drift.

## Bigger discussion point — version skew / no fallback

The new client routes **all** UDP (including DNS) through the relay and drops the old `s.Dial("udp")` path in `tsnetTransport.Dial`. Against a gateway that doesn't yet have the `:45353` listener, `dialTsnetUDPRelay` just errors and the flow (incl. DNS) fails — there's no fallback to the old path. You flag this ("needs a matched gateway/client deployment"), but it's worth deciding explicitly: hard-require gateway-first rollout, or fall back to `s.Dial("udp")` on a relay-unreachable (connection refused) error? A new client landing on a not-yet-upgraded gateway and losing DNS is a sharp edge; a one-shot "relay unreachable → fall back once" would make rollout order-independent.

## Smaller notes

- **`udpRelayStreamConn.Read` is unguarded** (only `Write` has `writeMu`). It's correct because `relayUDP`/`ServeUDP` each read from a single goroutine, but that's an invariant worth a comment on the type so a future caller doesn't add a concurrent reader and shred the framing.
- **`Read` returns `(0, nil)` for a zero-length datagram** — the documented "discouraged" `io.Reader` case. It works with `relayUDP` (forwards the empty datagram) and you test it, but a one-line comment noting the deliberate deviation would save the next reader a double-take.
- **Deadline layering:** `ServeUDP` sets a 60s read deadline, but `refreshIdleDeadline` immediately overrides it with the 30s idle (`SetDeadline` covers read+write). Net effect is 30s, which is fine — just slightly surprising the inner layer wins.
- **No per-peer cap** on concurrent relay streams (one TCP stream + goroutine per UDP flow). Bounded by onboarding/token auth and the 2s handshake / 30s idle, so the blast radius is authorized peers only, but a chatty daemon can open a lot. Flagging in case you want a soft cap.
- **Per-flow TCP stream** means UDP inherits TCP head-of-line blocking. Fine for DNS; for QUIC (which the macOS path explicitly punts on) it could hurt — probably moot since MITM forces TCP anyway, but worth confirming nothing real relies on UDP QUIC through `run`.
- **`authorizeTsnetUDPRelayPeer` promotion** is a sequence of unguarded onboard/db mutations; two streams from the same daemon racing first contact could interleave. Narrow window and the DB `UPDATE` is idempotent, but worth confirming the onboard registry is internally synchronized.

CI: test + lint are green; `preview` is red but failed in ~5s, which reads as infra rather than this Go-only change — probably just needs a re-run.

piscisaureus:
Following up on my earlier comment — I want to walk back part of it after re-reading #558 ("gateway: catch-all UDP/53 interception in tsnet exit-node mode").

**My framing of "the gateway has no arbitrary-UDP exit fallback" was too strong.** The gateway is a userspace exit node, and tsnet's netstack already *receives* arbitrary forwarded UDP — #558's own message says an exit-node client's DNS "has its DNS queries silently forwarded by tsnet's netstack to that resolver," and it taps exactly the right hook to intercept that: `GetUDPHandlerForFlow` on `*netstack.Impl` (`installTsnetUDPDNSCatchAll` in `tailscale.go`). It just only claims port 53 today:

```go
ns.GetUDPHandlerForFlow = func(src, dst netip.AddrPort) (func(nettype.ConnPacketConn), bool) {
    if dst.Port() == 53 { return g.serveTsnetUDPDNSFlow, true }
    if orig != nil { return orig(src, dst) }   // <- everything else forwarded un-intercepted
    return nil, false
}
```

So arbitrary UDP isn't blocked — it's flowing through the exit node and falling through to `orig`. The natural fix is to **generalize that hook to all ports** rather than frame UDP over a TCP stream:

```go
if dst.Port() == 53 { return g.serveTsnetUDPDNSFlow, true }
return g.serveTsnetUDPFlow(src, dst), true   // resolve src->profile (like peerIP), hand to relayUDP
```

`src` is the peer's tsnet IP, so the gateway can do peer→profile the same way it does for TCP and reuse the existing `relayUDP`. Compared to this PR that keeps real UDP end-to-end and removes everything the relay adds: no `:45353` listener, no per-flow tailnet TCP stream, no framing, no datagram head-of-line blocking under TCP, no hello/token handshake. The client keeps a plain `s.Dial("udp")` (which it does today).

Two honest caveats so this isn't hand-waving:

1. **What the relay-hello buys that a raw-UDP handler can't:** token-based *first-contact* promotion of a placeholder peer. A handler keyed on `src` IP would fall to the default profile for the first UDP packets of a daemon whose `/api/peer/tsnet/register` hasn't landed yet. Narrow window (the daemon registers on startup), but worth a conscious call.
2. **One thing to confirm before treating the relay as redundant:** that the per-process `clawpatrol run` daemon's `s.Dial("udp")` actually traverses this gateway exit-node netstack, so `GetUDPHandlerForFlow` sees it (the way `s.Dial("tcp")` lands in `RegisterFallbackTCPHandler`). It should by symmetry — but the macOS extension still carries its DNS dst-rewrite workaround (`Provider.swift` ~L40-48), which predates #558. If #558 already covers `Dial("udp")` flows, that macOS hack is now dead code and can go; if it's still load-bearing, then `Dial("udp")` is bypassing the catch-all and *that's* the gap to close — on both platforms, still without UDP-over-TCP.

None of this is wasted work: the auth model, the placeholder-promotion logic, and the edge-case tests largely carry over to a `serveTsnetUDPFlow` handler. But I'd hold on the UDP-over-TCP transport and pursue generalizing the `GetUDPHandlerForFlow` catch-all first — it's less surface area, keeps real UDP semantics, and would let us delete the macOS DNS-rewrite so both platforms share one UDP path. Happy to pair on the handler version or help confirm the `Dial("udp")`-reaches-the-hook question.

piscisaureus:
Tested the open question empirically — does a tsnet client's `Dial("udp")` to a public IP actually reach the exit-node gateway's `GetUDPHandlerForFlow`, or does it bypass it. **It reaches it, for any port.**

Setup: a fresh Linux VM with a self-contained `headscale` control plane (so I could approve the exit-node route myself), and a small two-node tsnet probe — a "gateway" that advertises `0.0.0.0/0` and installs the exact hook clawpatrol uses (`installTsnetUDPDNSCatchAll`'s `GetUDPHandlerForFlow`, but logging every port), and a "client" that pins the gateway as its exit node and does `Dial("udp", ...)`. After approving the gateway's exit route in headscale:

```
client Dial("udp", 8.8.8.8:53)    [DNS]     -> GW HOOK FIRED: src=100.64.0.2 dst=8.8.8.8:53 (port 53);    handler read 13 bytes
client Dial("udp", 8.8.8.8:12345) [non-DNS] -> GW HOOK FIRED: src=100.64.0.2 dst=8.8.8.8:12345 (port 12345); handler read 13 bytes
```

The hook fires for the non-DNS port and the handler reads the actual datagram, with `src` = the client's tailnet IP and `dst` = the real target. So the gateway's userspace exit node already receives arbitrary UDP through the same hook #558 taps for `:53` — `Dial("udp")` is not bypassing it.

That means the goal of this PR (arbitrary UDP for tsnet-mode `clawpatrol run`, handled by clawpatrol) can be met by **generalizing `GetUDPHandlerForFlow`** — `:53 -> dnsvip` (today) and everything else -> `relayUDP`, keyed on `src` for the onboarded-peer check — instead of framing UDP over a TCP stream. That keeps real UDP end-to-end and drops the `:45353` listener, per-flow TCP streams, framing, and datagram head-of-line blocking.

I'm going to open a separate PR that does it that way. It'll reuse your peer-authorization logic and the relevant tests/docs (credited to you) — the auth model and edge-case coverage carry over; it's the transport that changes. Will link it here.

piscisaureus:
Opened #643 — the `GetUDPHandlerForFlow` version, with the empirical round-trip validation in the description. It reuses the onboarded-peer authorization model and doc framing from here (credited). If that direction looks good, this PR can close in its favor.

dhruvkelawala:
Thanks for the careful review, and for validating the `GetUDPHandlerForFlow` path empirically.

I agree that #643 is the better upstream direction now that we know tsnet exit-node UDP reaches the gateway hook for arbitrary ports.

Why I think #643 is preferable:

- keeps real UDP semantics end-to-end;
- avoids a new `:45353` listener/protocol;
- avoids per-flow TCP streams and UDP-over-TCP head-of-line blocking;
- avoids client/gateway protocol version skew;
- gives Linux and macOS a path toward one shared gateway-owned UDP implementation.

The token-based first-contact promotion in this PR was useful to explore, but I agree it does not justify the extra transport surface if the gateway hook can handle the main UDP problem directly.

I’m testing #643 from this environment and will report results there. Assuming the live canaries pass, this PR can close in favor of #643.

dhruvkelawala:
**Confirmed working live, for both DNS and arbitrary UDP, with no leak under soak.**

Tested on a real two-node exit-node tailnet — the same setup where the native-UDP approach (#643) times out because the client's packet filter drops inbound UDP that has no cached outbound flow. #640 sidesteps that entirely by framing UDP over a TCP stream (the client filter always accepts inbound non-SYN TCP), so the reply gets back.

**Functional**

- **UDP DNS** (`relay → dnsvip`): `dig @8.8.8.8 chatgpt.com` → `104.18.32.47` (`write 52B → read 83B`).
- **Arbitrary UDP** (`relay → relayUDP`, NTP/123, not DNS): `NTP_OK from 162.159.200.123` (`write 48B → read 48B`, ~2 ms).
- Handshake/auth and the per-peer token promotion path both work; the gateway ACL admits the client → `:45353`.

**Soak** (lean gateway, 2 bursts / 2,300 flows, up to 40 concurrent): **2,295/2,300 ok** (5 transient NTP timeouts), 105–155 flows/s.

| event | goroutines | rss |
| --- | --- | --- |
| baseline (idle) | 85 | 67 MB |
| burst-1 peak (800) | 1677 | 97 MB |
| burst-1 settled | 87 | 99 MB |
| burst-2 peak (1500) | 3067 | 135 MB |
| burst-2 settled | 88 | 138 MB |

Goroutines return to baseline after every burst (so every `handleTsnetUDPRelayConn` / `dnsvip.ServeUDP` / `relayUDP` flow is reclaimed), and a forced-GC heap snapshot at idle shows a normal live working set (`HeapAlloc ≈ 42 MB`, 40k objects); the elevated `HeapSys`/RSS is just Go arena retention (`HeapReleased` low), not a leak. **No goroutine or heap leak.**

Net: this is the approach that actually delivers UDP for tsnet `clawpatrol run` clients today. Only nits to weigh for merge: the extra `:45353` listener + framing/HOL vs native UDP, and that it relies on the client→gateway TCP grant.

dhruvkelawala:
Upstream root cause filed on Tailscale: **tailscale/tailscale#20064** (related: tailscale/tailscale#14229).

A userspace/netstack exit-node client drops inbound UDP *replies* because injected outbound UDP bypasses `RunOut`, so the packet filter never records the reverse-flow tuple and `runIn` drops the reply with "no matching rule" — while inbound non-SYN TCP is admitted unconditionally. That is exactly why this PR's TCP-framed relay works end-to-end where native UDP (#643) does not yet.

Sequencing implied by the upstream bug:
- **Now:** this PR (TCP relay) is the only path that works against current tsnet, since TCP is the transport the client filter admits.
- **Upstream:** tailscale/tailscale#20064 proposes recording flow state in `injectedRead`.
- **Later:** once that lands, #643's native UDP works end-to-end and the relay can be removed.

piscisaureus:
I think we can just vendor some of the tailscale code until the fix lands upstream

dhruvkelawala:
Follow-up after retesting latest #643 (`8d2de80`): **#643 now works end-to-end**, so this relay PR can close in its favor if maintainers agree.

The earlier #643 failure was real: the gateway emitted UDP replies correctly, but the tsnet client dropped them because injected outbound UDP bypassed Tailscale `RunOut`, so the reverse-flow tuple was never recorded. Latest #643 vendors a patched `tailscale.com` that fixes that client-side filter gap.

With latest #643 installed on both client and gateway, final smoke passes without this PR's `:45353` TCP relay:

| check | result |
| --- | --- |
| UDP DNS | `dig @8.8.8.8 chatgpt.com` resolves |
| Arbitrary UDP / NTP | `NTP_OK from 162.159.200.123 48B rtt=3ms` |
| TCP control | `dig +tcp @8.8.8.8 example.com` resolves |
| OpenClaw gateway through `clawpatrol run` | `/health` → `{"ok":true,"status":"live"}` |

So #640 remains a useful validated fallback/design record for current unpatched Tailscale, but with the vendored Tailscale fix in #643, the native UDP approach is better upstream: no extra listener/protocol, no UDP-over-TCP framing, no TCP HOL blocking, and real UDP semantics.

dhruvkelawala:
Closing in favour of #643

PR #643, opened 2026-06-09
gateway: handle tsnet exit-node UDP via GetUDPHandlerForFlow

piscisaureus:
Alternative to #640 that reaches the same goal — arbitrary UDP for tsnet-mode `clawpatrol run` — through the exit-node netstack instead of framing UDP over a TCP stream, plus making sure QUIC can't bypass the HTTPS MITM.

### The UDP path
* The gateway is a userspace exit node, so tsnet's netstack already receives the UDP its clients send. The UDP/53 catch-all from #558 proves this by tapping `GetUDPHandlerForFlow`. This generalizes that hook from DNS-only to all UDP: `:53` to any resolver IP → `dnsvip`; other UDP from an onboarded peer → `relayUDP` — the same dispatch WireGuard mode's promiscuous netstack already applies.
* No `:45353` listener, no per-flow tailnet TCP stream, no framing, no datagram head-of-line blocking — real UDP end-to-end. The client keeps a plain `s.Dial("udp")`.
* Non-onboarded sources fall through to tsnet's default handler, so the gateway doesn't become an open UDP proxy for arbitrary tailnet nodes that pin it as an exit node.
* Factored into a testable `tsnetUDPDisposition`, unit-tested.

### QUIC, without breaking pass-through
* **UDP/443 to an *intercepted* (VIP'd) host is dropped** — in both the tsnet catch-all and WG mode's dispatch — so HTTPS can't ride QUIC past the TCP/443 SNI-peek MITM; the client falls back to interceptable TCP. UDP/443 to a host the gateway *passes through* (no VIP) is relayed like any other UDP: we don't intercept that host's HTTPS either, so there's nothing to bypass and no reason to break its HTTP/3. (This also closes a pre-existing gap: WG mode relayed all UDP/443, latent because the CLI agents don't speak h3.)
* **`Alt-Svc` response header stripped** on MITM'd responses, so agents don't discover h3 for the hosts the gateway does intercept. Spliced/pass-through connections never reach the response-rewrite path, so their Alt-Svc is left alone. The client connection is forced to `http/1.1` by the MITM ALPN, so there's no HTTP/2 ALTSVC frame to handle. (Intercepted names already return no SVCB/HTTPS DNS record, so there's no DNS h3-discovery channel to close for them either.)

### Validation
Verified empirically on a fresh Linux VM with a self-contained headscale control plane (two tsnet nodes — a gateway advertising exit routes with this hook, and a client pinning it as exit node):

```
client Dial("udp", 216.239.35.0:123) [NTP, non-DNS]
  GW HOOK: UDP flow src=100.64.0.6 dst=216.239.35.0:123 (port 123) — relaying
  REPLY: 48 bytes  (round-trip OK)
client Dial("udp", 8.8.8.8:53) [DNS]
  GW HOOK: UDP flow src=100.64.0.6 dst=8.8.8.8:53 (port 53) — relaying
  REPLY: 61 bytes  (round-trip OK)
```

i.e. a non-DNS UDP datagram from the client transits the gateway hook, is relayed to the real upstream, and the reply comes back.

Credit to @dhruvkelawala / #640: the onboarded-peer authorization model and the doc framing are adapted from there. That PR's per-peer token first-contact promotion has no analogue on the raw-UDP path, so a daemon's very first UDP before its register lands falls through to the default handler (the register completes on startup). If we take this, #640 can close.


dhruvkelawala:
**Live result: the catch-all is correct, but it can't deliver UDP to a tsnet `clawpatrol run` client by itself — the reply is dropped on the _client_.**

I built this branch and instrumented both ends on a real two-node exit-node setup.

- **Gateway side is perfect.** `GetUDPHandlerForFlow` fires with the right disposition, `dnsvip` builds the answer, and the writeback leaves the netstack cleanly — counter delta across the write: `udp_packets_sent 0→1`, `ip_packets_sent +1`, every error/drop/forward counter `0`.
- **The client never receives the reply.** Daemon netstack: `udp_packets_received: 0`. Under the _identical_ client state, TCP DNS (`dig +tcp @8.8.8.8`) succeeds while UDP DNS times out.

Root cause is in Tailscale's packet filter on the client, not in this PR:

- `wgengine/filter.runIn` accepts inbound **non-SYN TCP unconditionally** (`"tcp non-syn"`), but inbound **UDP** only if the reverse flow tuple is cached in `f.state.lru` (recorded by `runOut`) or an ACL rule matches.
- On a tsnet client the agent's outbound UDP is emitted via `netstack.Impl → tundev.InjectOutboundPacketBuffer`, and the injected path **bypasses `RunOut`** (tstun `wrap.go`: _"injectedRead handles injected reads, which bypass filters"_). So the flow is never recorded, the node has `0 filters`, and the inbound reply (`src=8.8.8.8:53`) is dropped before it reaches the app. A kernel-TUN exit-node client doesn't hit this because its outbound UDP goes through `RunOut`.

Implication: a gateway-only native-UDP path can't work for tsnet `clawpatrol run` clients regardless of how clean the gateway code is. The reply has to arrive over a channel the client filter admits — i.e. TCP-framed (#640, confirmed live), or Tailscale has to change how tsnet netstack outbound is filtered. Happy to share the full packet-counter logs and filter trace.

dhruvkelawala:
Filed the upstream root cause: **tailscale/tailscale#20064** (related: tailscale/tailscale#14229).

This explains the live result where the gateway hook fires and writes the reply cleanly but the tsnet `clawpatrol run` client never receives it: on a userspace/netstack exit-node client, injected outbound UDP bypasses `RunOut`, so `wgengine/filter` has no recorded flow and `runIn` drops the inbound reply ("no matching rule"). Inbound non-SYN TCP is admitted unconditionally, which is why TCP works on the identical tunnel/state.

So this PR's native-UDP approach is the right end state but is **blocked until tailscale/tailscale#20064 is fixed** (record flow state in `injectedRead`). Until then #640's TCP-framed relay is the only path that survives the client filter.

piscisaureus:
@dhruvkelawala Can you try this version again? It should work now.

dhruvkelawala:
**Retested latest #643 (`8d2de80`) — confirmed working end-to-end now.**

This is a different result from my earlier test of `f41191a`: the native gateway path was already emitting replies correctly, but the tsnet client dropped them. The new vendored Tailscale patch fixes that client-side filter gap by recording `RunOut` state for injected netstack packets, so inbound UDP replies are now admitted.

Setup:

- Client: latest #643 binary (`dcb465770428ead9...` locally), joined to gateway `100.66.146.96`.
- Gateway: latest #643 on `clawpatrol-gateway-006` (`100.66.146.96`).
- Gateway log confirms:
  ```text
  tsnet: UDP catch-all installed (:53 → dnsvip, :443 QUIC dropped for VIPs, other → relay for onboarded peers)
  ```
- No #640 `:45353` UDP-over-TCP relay in this path; this is native UDP via `GetUDPHandlerForFlow` + the patched Tailscale netstack.

Final smoke:

| check | result |
| --- | --- |
| UDP DNS: `dig @8.8.8.8 chatgpt.com A +short` | `104.18.32.47`, `172.64.155.209` |
| Arbitrary UDP / NTP | `NTP_OK from 162.159.200.123 48B rtt=3ms` |
| TCP control: `dig +tcp @8.8.8.8 example.com A +short` | `172.66.147.243`, `104.20.23.154` |
| OpenClaw gateway through `clawpatrol run` | `/health` → `{"ok":true,"status":"live"}` |

Conclusion: **#643 now fixes the real issue and works for Linux tsnet `clawpatrol run` UDP.** The earlier blocker is resolved by the vendored Tailscale patch; native UDP is the right upstream direction over #640's TCP relay.

PR #644, opened 2026-06-09
aws_credential: re-sign non-EKS proxied requests + inject placeholder AWS env vars

arnauorriols:
Implements **cl-pqgf**: extend the `aws_credential` plugin so plain aws-cli calls to `*.amazonaws.com` work through an `("https" "aws")` endpoint on a host with no real AWS credentials.

## Part 1 — Re-sign non-EKS proxied requests
`SignHTTPRequest` now branches on the endpoint type:
- **EKS endpoint** (`awsEKSParams` satisfied): unchanged — mints the `k8s-aws-v1.` bearer.
- **Non-EKS endpoint**: new `reSignProxiedRequest` path. It strips the client's `Authorization` + `X-Amz-Security-Token`, derives service/region from the incoming SigV4 credential scope (`Credential=AKID/<date>/<region>/<service>/aws4_request`, falling back to the hostname), reuses the existing `X-Amz-Content-Sha256` as the payload hash, and re-signs with the gateway's stored creds via `aws-sdk-go-v2/aws/signer/v4`.

## Part 2 — Auto-inject placeholder AWS env vars
`AWSCredential` now implements `EnvPushdownProvider`, exporting placeholder `AWS_ACCESS_KEY_ID` (20-char AKIA shape), `AWS_SECRET_ACCESS_KEY` (40-char), and `AWS_DEFAULT_REGION` — mirroring how `anthropic_oauth` injects `ANTHROPIC_AUTH_TOKEN`. aws-cli signs locally with placeholders; the gateway re-signs with real creds per Part 1.

## Tests
- `TestAWSCredentialReSignsNonEKSRequest` — scope wins over host, client signature/session-token replaced, content hash reused.
- `TestAWSCredentialReSignFallsBackToHost` — hostname fallback + STS session-token stamping.
- `TestParseSigV4CredentialScope` / `TestServiceRegionFromHost` — unit coverage incl. global-endpoint defaults.
- `TestAWSCredentialEnvVarsMirrorsAnthropic` — env-var shapes.
- EKS path regression guard (`...EmitsEKSBearer`) unchanged and passing.

### DIFF
--- a/cmd/clawpatrol/integrations.go
+++ b/cmd/clawpatrol/integrations.go
@@ -127,6 +127,7 @@ func caPathPushdownVars(caPath string) []pushdownEnvVar {
 		"GIT_SSL_CAINFO",
 		"DENO_CERT",
 		"PIP_CERT",
+		"AWS_CA_BUNDLE",
 	}
 	out := make([]pushdownEnvVar, 0, len(keys))
 	for _, k := range keys {
--- a/internal/config/plugins/credentials/aws.go
+++ b/internal/config/plugins/credentials/aws.go
@@ -17,10 +17,14 @@ package credentials
 // Either way the agent never sees real credentials.
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -39,9 +43,14 @@ import (
 )
 
 // emptyPayloadHash is the SHA-256 of the empty string — the SigV4
-// payload hash the signer requires when an incoming request carried no
-// X-Amz-Content-Sha256 header to reuse.
+// payload hash for a request with no body.
 const emptyPayloadHash = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+
+// unsignedPayload is the SigV4 sentinel that omits the body from the
+// canonical request's payload hash. S3 emits it as a literal
+// X-Amz-Content-Sha256 header for streaming / chunked uploads; we also
+// fall back to it for non-S3 bodies too large to buffer for hashing.
+const unsignedPayload = "UNSIGNED-PAYLOAD"
 
 // AWSCredential is part of the clawpatrol plugin API.
 //
@@ -111,11 +120,24 @@ func (c *AWSCredential) reSignProxiedRequest(ctx context.Context, req *http.Requ
 		return fmt.Errorf("aws_credential: cannot derive service/region for %q (no SigV4 credential scope and unrecognized host)", req.URL.Host)
 	}
 
-	// Reuse the hash the agent already computed and signed over rather
-	// than recomputing from the (possibly already-consumed) body.
+	// Determine the SigV4 payload hash the agent signed over.
+	//
+	// botocore only puts X-Amz-Content-Sha256 on the wire for S3
+	// (S3SigV4Auth); every other service still signs over SHA256(body)
+	// in its canonical request but sends no such header. So:
+	//   header present → reuse it verbatim. It's the value the client
+	//     signed over and may be one we cannot recompute from the body
+	//     (S3 UNSIGNED-PAYLOAD, a streaming-chunked literal, …).
+	//   header absent  → the client signed over SHA256(body); recompute
+	//     it from the actual body so the canonical requests agree. The
+	//     old empty-string fallback broke every non-S3 service with a
+	//     body (STS, IAM, DynamoDB, …) with SignatureDoesNotMatch.
 	payloadHash := req.Header.Get("X-Amz-Content-Sha256")
 	if payloadHash == "" {
-		payloadHash = emptyPayloadHash
+		payloadHash, err = hashRequestBody(req)
+		if err != nil {
+			return fmt.Errorf("aws_credential: hash %s request body: %w", service, err)
+		}
 	}
 
 	// Drop the agent's placeholder signature + any session token it
@@ -129,6 +151,42 @@ func (c *AWSCredential) reSignProxiedRequest(ctx context.Context, req *http.Requ
 		return fmt.Errorf("aws_credential: re-sign %s request: %w", service, err)
 	}
 	return nil
+}
+
+// hashRequestBody computes the SigV4 payload hash over req.Body — what
+// botocore's base SigV4Auth signs over for non-S3 services — and
+// restores the body so the upstream send is unchanged. A nil/empty body
+// hashes to SHA256(""). Bodies over the buffer cap return
+// UNSIGNED-PAYLOAD rather than buffering unbounded memory; the gateway
+// terminates TLS, so transport integrity still covers the streamed body.
+func hashRequestBody(req *http.Request) (string, error) {
+	if req.Body == nil || req.Body == http.NoBody {
+		return emptyPayloadHash, nil
+	}
+	// The gateway already buffers POST/PUT/PATCH bodies up to this cap
+	// for the rules engine before we sign, so for the path that needs a
+	// recomputed hash (non-S3 bodies) the bytes are typically in memory
+	// already — reading here is cheap.
+	const limit = config.DefaultBodyBufferLimit
+	orig := req.Body
+	buf, err := io.ReadAll(io.LimitReader(orig, limit+1))
+	if err != nil {
+		return "", err
+	}
+	if int64(len(buf)) > limit {
+		// Over the cap: don't buffer the rest. Stitch the bytes we
+		// peeked back ahead of the untouched remainder and sign
+		// UNSIGNED-PAYLOAD.
+		req.Body = struct {
+			io.Reader
+			io.Closer
+		}{io.MultiReader(bytes.NewReader(buf), orig), orig}
+		return unsignedPayload, nil
+	}
+	sum := sha256.Sum256(buf)
+	req.Body = io.NopCloser(bytes.NewReader(buf))
+	req.ContentLength = int64(len(buf))
+	return hex.EncodeToString(sum[:]), nil
 }
 
 // awsServiceRegion derives the SigV4 service + region for re-signing. It
--- a/internal/config/plugins/credentials/aws.go
+++ b/internal/config/plugins/credentials/aws.go
@@ -1,24 +1,33 @@
 package credentials
 
 // aws_credential: static AWS API credentials (access key + secret +
-// optional session token) used to mint an EKS-style bearer token for
-// the kubernetes endpoint. At inject time the gateway calls STS
-// GetCallerIdentity via aws-sdk-go-v2's presigner with the
-// `x-k8s-aws-id` header carrying the cluster name; the presigned URL
-// is base64url-encoded as `k8s-aws-v1.<url>` and stamped as the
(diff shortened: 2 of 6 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR description against the branch (`020f2c6`) — it accurately describes the implementation, and no change is needed. Verified:

- **Part 1 (re-sign):** `SignHTTPRequest` branches EKS vs. non-EKS; `reSignProxiedRequest` strips `Authorization` + `X-Amz-Security-Token`, derives service/region from the SigV4 credential scope with a hostname fallback (`awsServiceRegion` → `parseSigV4CredentialScope` / `serviceRegionFromHost`, incl. global-endpoint → us-east-1), reuses `X-Amz-Content-Sha256` as the payload hash (empty-string SHA-256 fallback), and re-signs via `aws-sdk-go-v2/aws/signer/v4`.
- **Part 2 (env pushdown):** `AWSCredential` implements `EnvPushdownProvider`; placeholders are AKIA-shaped 20-char key + 40-char secret + `AWS_DEFAULT_REGION`, mirroring `anthropic_oauth`.
- **Gates:** `go test ./internal/config/plugins/credentials/` passes (incl. all listed tests + the EKS regression guard); `go vet` and `gofmt -l` clean.

No code changes from this watcher pass.

arnauorriols:
### `CLAWPATROL_SECRET_<NAME>_ACCESS_KEY_ID` env path doesn't actually populate the AWS slots

Hit this while wiring up a local gateway against this branch. Both the doc comment and the error message advertise an env-var path for the AWS credential material:

- `aws.go:50` — *"filled via the dashboard or `CLAWPATROL_SECRET_<NAME>_<SLOT>`"*
- `aws.go:198` — *"set `CLAWPATROL_SECRET_<NAME>_ACCESS_KEY_ID` and `_SECRET_ACCESS_KEY`, or fill the slots in the dashboard"*

But `EnvSecretStore` only folds a fixed allowlist of suffixes into `Secret.Extras`:

```go
// internal/config/runtime/secrets.go:13
var envParts = []string{"CA", "CERT", "KEY"}
```

`awsCredentialMaterial` reads `sec.Extras["access_key_id"] / ["secret_access_key"] / ["session_token"]` (`aws.go:193-195`), and none of those suffixes are in `envParts` — so `CLAWPATROL_SECRET_AWS_ACCESS_KEY_ID` etc. are read but never land in `Extras`. The env path silently no-ops; only the dashboard (which writes per-slot rows to `credential_secrets`) works today.

**Fix options:**
1. Extend `envParts` to include `ACCESS_KEY_ID`, `SECRET_ACCESS_KEY`, `SESSION_TOKEN` (makes the advertised path real — nice for headless/CI gateways with no dashboard), or
2. Drop the env-var mention from the comment + error message and point only at the dashboard.

I'd lean toward (1) since the env path is genuinely useful for unattended gateways, but either resolves the mismatch.

arnauorriols:
[bot-reply]
Went with fix option 1. Pushed 8c39d54:
- Extended `envParts` in `internal/config/runtime/secrets.go` to include `ACCESS_KEY_ID`, `SECRET_ACCESS_KEY`, `SESSION_TOKEN`. Lowercased, these map exactly to the `access_key_id` / `secret_access_key` / `session_token` keys `awsCredentialMaterial` reads (`aws.go:193-195`), so the advertised `CLAWPATROL_SECRET_<NAME>_ACCESS_KEY_ID` env path now actually populates the AWS slots for headless/CI gateways.
- Added `secrets_test.go` covering both the new AWS suffixes and the original CA/CERT/KEY mtls path (regression guard).

arnauorriols:
### `clawpatrol run -- aws …` fails TLS: `AWS_CA_BUNDLE` missing from the CA-bundle env pushdown

Hit this taking the AWS path end-to-end (`clawpatrol run -- aws sts get-caller-identity` against a gateway built from this branch). First call:

```
aws: [ERROR]: SSL validation failed for https://sts.us-east-1.amazonaws.com/
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain
```

**Root cause.** `clawpatrol run` already injects the MITM CA into the wrapped process for curl / node / python-requests / deno / git via `caPathPushdownVars` (`cmd/clawpatrol/integrations.go:121-136`), all pointing at `~/.clawpatrol/ca.crt`:

```
SSL_CERT_FILE  NODE_EXTRA_CA_CERTS  REQUESTS_CA_BUNDLE
CURL_CA_BUNDLE  GIT_SSL_CAINFO  DENO_CERT  PIP_CERT
```

But aws-cli / botocore honors **none** of those. The only CA override it reads is `AWS_CA_BUNDLE` (or `--ca-bundle` / the `ca_bundle` config) — botocore otherwise uses its own vendored certifi bundle and never consults the OS trust store. Since `AWS_CA_BUNDLE` isn't in that list, aws-cli never sees the gateway CA and rejects the MITM cert.

**Confirmed empirically.** With all seven vars above auto-injected by `run`, aws-cli still failed `CERTIFICATE_VERIFY_FAILED`. Re-running with `AWS_CA_BUNDLE` pointed at the same `ca.crt` made the handshake succeed (it then reached AWS and signing took over):

```
clawpatrol run -- env AWS_CA_BUNDLE=~/.clawpatrol/ca.crt aws sts get-caller-identity
```

Worth noting: installing the CA in the system trust store (`update-ca-certificates`, as the `join` success message suggests) does **not** fix aws-cli either, because botocore ignores the OS store. So for the AWS use case the system-trust path is a red herring — `AWS_CA_BUNDLE` is the real lever.

**Fix.** One line — add `"AWS_CA_BUNDLE"` to the key list in `caPathPushdownVars` (`integrations.go:122`). That helper is the single source for both `clawpatrol env` and the `run` pushdown, so it covers every path at once. After that, `clawpatrol run -- aws …` works with zero manual env and no trust-store changes.

This is technically a clawpatrol-core gap rather than something this PR introduced, but the AWS path this PR adds isn't usable through `clawpatrol run` without it — so flagging it here. Happy to send the one-liner as a separate PR.

(Separately, still chasing a `SignatureDoesNotMatch` from STS once TLS is past — will report separately if it turns out to be in `reSignProxiedRequest`.)

arnauorriols:
[bot-reply]
Pushed 100b49b addressing this comment:
- Added `AWS_CA_BUNDLE` to the `caPathPushdownVars` key list in `cmd/clawpatrol/integrations.go` so it points at the same `~/.clawpatrol/ca.crt` as the other CA-bundle vars. Since that helper is the single source for both `clawpatrol env` and the `run` pushdown, aws-cli/botocore now sees the gateway CA on every path and `clawpatrol run -- aws …` works with zero manual env and no trust-store changes.

Re the `SignatureDoesNotMatch` from STS: not touched here — please report separately and we'll chase `reSignProxiedRequest` if it turns out to be on our side.

arnauorriols:
### Bug: `reSignProxiedRequest` produces `SignatureDoesNotMatch` for every non-S3 service with a request body

Following up on the earlier note — I chased the `SignatureDoesNotMatch` and it's a real bug in this PR's re-signing, not a credential issue. **Please fix.**

#### Repro / evidence

Through a gateway built from this branch, with a correctly-configured `aws_credential`:

| Call | Body | Client sends `X-Amz-Content-Sha256`? | Result |
|------|------|------|--------|
| `aws sts get-caller-identity` | non-empty | ❌ no (botocore base `SigV4Auth`) | **SignatureDoesNotMatch** |
| `aws s3api list-buckets` | empty | ✅ yes (`S3SigV4Auth`) | ✅ success — returned real buckets |

S3 succeeding proves the stored credential is correct and re-signing works *when the client puts the payload hash on the wire*. The only differing variable is that header.

#### Root cause

`reSignProxiedRequest` (`internal/config/plugins/credentials/aws.go:116-119`):

```go
payloadHash := req.Header.Get("X-Amz-Content-Sha256")
if payloadHash == "" {
    payloadHash = emptyPayloadHash   // SHA256("")
}
```

botocore only emits the `X-Amz-Content-Sha256` **header** for S3 (`S3SigV4Auth`). Every other service still signs over `SHA256(body)` in its canonical request but doesn't send it as a header. So for STS the gateway sees no header, falls back to `emptyPayloadHash`, and signs over the empty-string hash — while the actual POST body (`Action=GetCallerIdentity&Version=2011-06-15`) is non-empty. AWS recomputes `SHA256(real body)`, the canonical requests diverge, signature rejected.

**Scope is broad:** this breaks essentially every non-S3 service that sends a body — STS, IAM/EC2 (query-protocol POSTs), DynamoDB/SQS/SNS (JSON POSTs), etc. Only S3 and empty-body requests work today. The canonical first smoke test (`sts get-caller-identity`) is in the broken set.

#### Suggested fix

Keep the reuse-when-present branch (it's required for S3 `UNSIGNED-PAYLOAD` / streaming-chunked, where the header is a literal, not a body hash). Only the empty fallback is wrong — when the header is absent, hash the actual body:

```
header present            → reuse it                (S3, UNSIGNED-PAYLOAD, streaming)   no read
header absent, body ≤ cap → payloadHash = SHA256(body)
header absent, body > cap → payloadHash = "UNSIGNED-PAYLOAD" (stream untouched)
```

**On the buffering worry:** reading the body here is not a new footgun, and the `HTTPRequestSigner` contract already sanctions it — *"SignHTTPRequest may consume and replace req.Body (typically reading it in full to hash, then restoring with io.NopCloser)"* (`internal/config/runtime/runtime.go:42-44`). More importantly, the gateway **already** buffers POST/PUT/PATCH bodies up to `BodyBufferLimit` (1 MiB) for the rules engine at `cmd/clawpatrol/main.go:2073-2078`, which runs *before* the sign call at `main.go:2401`. So for the path that needs it (non-S3 POSTs like STS), the bytes are already in memory — recomputing the hash is free. Large bodies (>1 MiB) are S3, which hits the reuse path; the `UNSIGNED-PAYLOAD` fallback covers the rare large non-S3 case with no buffering (TLS, which the gateway terminates, provides body integrity).

A test mirroring `TestAWSCredentialReSignsNonEKSRequest` but with **no** `X-Amz-Content-Sha256` header on a non-empty body would have caught this and would guard the fix.

arnauorriols:
[bot-reply]
Pushed 551d0ee5 addressing this comment:
- `reSignProxiedRequest` no longer falls back to `SHA256("")` when the client omits `X-Amz-Content-Sha256`. The reuse-when-present branch is kept (required for S3 `UNSIGNED-PAYLOAD` / streaming-chunked literals); only the empty fallback changed.
- New `hashRequestBody` helper: header absent + body ≤ cap → `payloadHash = SHA256(body)`; header absent + body > `DefaultBodyBufferLimit` (1 MiB) → `UNSIGNED-PAYLOAD` (stream not buffered; TLS, which the gateway terminates, covers integrity). nil/empty body → `SHA256("")`. The body is restored in every case so the upstream send is unchanged.
- Tests: added `TestAWSCredentialReSignsNonEKSRequestNoContentHashHeader` (the requested mirror of `TestAWSCredentialReSignsNonEKSRequest` with no content-hash header on a non-empty STS body) plus a `TestHashRequestBody` table covering the body-hash, empty, and over-cap paths.

This fixes `SignatureDoesNotMatch` for non-S3 services with a body (STS, IAM, DynamoDB, SQS/SNS, …). `gofmt`, `go vet`, and the credentials package tests are green.

PR #645, opened 2026-06-09
feat: profile-scoped endpoint/credential discovery endpoint

arnauorriols:
## What

Adds a reserved, gateway-intercepted address — `https://clawpatrol/` (plus a fixed-IP fallback) — that an agent running behind the tunnel can curl to learn, **scoped to its own device profile**, exactly which endpoints and credentials it has access to and *how* to reach each one. Replaces the manual "here's how you reach Postgres, here's the tunnel to run" paragraph users paste into their agent each session.

## How

- **Identity from the connection, not a token.** The manifest is computed live from the calling device's current profile, resolved from the tunnel IP via the existing `profileFor(peerIP(c))` path. It lists only what *this* profile grants — never the whole config.
- **Per endpoint**: protocol/type, host(s)/port(s), database/path, the credential placeholder to present, whether a tunnel must be active (and which), and a concrete client-invocation hint (`psql` / `kubectl` / `clickhouse-client` / `ssh` / `curl`).
- **Credentials**: what the profile has, with placeholders and reachable endpoints; endpoints with no bound credential are flagged so the agent knows the boundary.
- **Two formats, one representation**: agent-readable markdown (default) and JSON (`?format=json` or `Accept: application/json`), both rendered from a single `DiscoveryManifest`.
- **Interception**: reserved SNI `clawpatrol` (and the fixed discovery VIP for the no-SNI IP-literal case) are caught in `handle()` and served locally, reusing the proven `serveTSNetDirect` TLS-terminate pattern. DNS answers the reserved name from a fixed VIP slot (one past `MaxID`, never allocated to a real host) and never forwards it upstream.

## Files

- `cmd/clawpatrol/discovery.go` — manifest builder, markdown/JSON renderers, TLS serve, host/VIP predicates
- `cmd/clawpatrol/main.go` — hook reserved SNI / fixed VIP in `handle()`
- `cmd/clawpatrol/dnsvip/dnsvip.go` — local answer for the reserved name

## Tests

`go test ./cmd/clawpatrol/ ./cmd/clawpatrol/dnsvip/` — all green; `gofmt -l .` and `go vet` clean.

- profile A sees only A's endpoints/credentials; profile B a different set (no cross-leak)
- tunneled endpoint reports its tunnel; non-tunneled reports none
- host/port/sslmode/placeholder/disambiguator extraction (https, postgres, clickhouse_native)
- markdown + JSON render from one representation, same scoping
- unknown profile → empty manifest + note, not a config dump
- reserved name served locally from its fixed VIP, authoritative, never forwarded upstream; no collision with real allocations

**E2E note:** a full WireGuard two-device curl wasn't run (no WG harness in the worktree); the HTTP render path is exercised end-to-end via `httptest`, and the TLS/serve glue reuses the existing `serveTSNetDirect` path.


arnauorriols:
[bot-reply]
Reviewed the PR description — it's an accurate summary of the work already implemented on this branch (commit 906d2ec): the manifest builder + markdown/JSON renderers and TLS serve in `discovery.go`, reserved-SNI / fixed-VIP interception in `handle()` (`main.go`), and the local authoritative answer for the reserved name in `dnsvip.go`. It contains no change request or open question, and CI (lint / test / preview) is green. No code changes made.

arnauorriols:
Write some functional tests that assert the exact output given different hcl configurations (empty profile, profile with simple endpoints and credentials, profiles with multiple creds for the same endpoint, etc)

arnauorriols:
[bot-reply]
Pushed bb6601d addressing this comment — golden functional tests asserting the exact rendered output (markdown + JSON) per HCL config shape:

- **empty profile** (`testdata/discovery/empty.hcl`): a declared profile that grants no credentials, with other endpoints/creds present in the config — locks down that scoping returns an empty manifest (0 endpoints, 0 credentials), not a leak or error.
- **simple** (`simple.hcl`): one `https` endpoint + one bearer credential — host, placeholder, curl hint, and the Credentials section.
- **multiple creds, same endpoint** (`multicred.hcl`): two `postgres_credential`s bound to the same endpoint, disambiguated by user/database — both list under the endpoint (sorted), both back-reference the one endpoint, and the `psql` hint pins to the first by sort order.

Each case renders one profile and compares byte-for-byte against committed goldens. The JSON golden is captured through the real `writeDiscoveryResponse` path, so it matches exactly what an agent receives. Goldens regenerate with `go test ./cmd/clawpatrol -run TestDiscoveryGolden -update`. `gofmt`, `go vet`, and the full package test suite are green.

arnauorriols:
Add also a test that uses tunnels

arnauorriols:
[bot-reply]
Pushed 5308d4e addressing this comment:
- Added a `tunnel` golden case to `TestDiscoveryGolden` (fixture `testdata/discovery/tunnel.hcl`, golden `.md`/`.json`). It renders the `tunneled` profile, whose `prod-pg` endpoint sits behind a `local_command` tunnel (`csql`).
- The golden locks down the tunnel-reporting path end-to-end: JSON carries `"tunnel": {"name":"csql","type":"local_command"}`, markdown spells out `Tunnel: REQUIRED — csql (local_command) ...`, and the same profile's directly-reachable `github` endpoint renders `Tunnel: none` so the contrast is captured.

cmd/clawpatrol/testdata/discovery/tunnel.md
```diff
@@ -0,0 +1,31 @@
...
+- Tunnel: none (reachable directly through the gateway)
+- Credential: bearer_token `gh` — send placeholder `PH_GH`
+- Example: `curl https://api.github.com/ -H "Authorization: Bearer PH_GH"`
+
+### prod-pg  (postgres)
+
+- Host(s): main-pg.example
+- Port: 5432
+- SSL mode: require
+- Tunnel: REQUIRED — `csql` (local_command) must be active to reach this endpoint
```
arnauorriols: is this information really relevant for the agent? Add the other tunnel types in this test, and think twice what is the relevant information from the point of view of the agents regarding the tunnel. I think they don't need a specific tunnel info, we only need to tell the agent the proper host they need to hit

cmd/clawpatrol/testdata/discovery/simple.md
```diff
@@ -0,0 +1,21 @@
...
+## Endpoints (1)
+
+### github  (https)
+
+- Host(s): api.github.com
+- Tunnel: none (reachable directly through the gateway)
+- Credential: bearer_token `gh` — send placeholder `PH_GH`
+- Example: `curl https://api.github.com/ -H "Authorization: Bearer PH_GH"`
+
+## Credentials (1)
```
arnauorriols: is this info really relevant for the agent? maybe we shouldn't structure the info as endpoints/credentials. Instead, focus exactly on what can be useful to know by agents to leverage the endpoints and credentials that are available in their gateway profile

cmd/clawpatrol/testdata/discovery/empty.md
arnauorriols, line 1: This is not useful for agents. We should tell the agent that their gateway profile is empty, and give some hints suggesting to configure the profile in the gateway so that they can get all the value from clawpatrol. We should tell the agent also where the user can reach the dashboard to change the profile of the device, for example

cmd/clawpatrol/testdata/discovery/tunnel.md
arnauorriols: [bot-reply]
Agreed — pushed c0a9cbe. You're right that the tunnel name/type is noise for the agent: the gateway intercepts transparently and brings the tunnel up itself, so the agent dials the same host whether or not a tunnel sits behind it, and it can't act on the tunnel's identity anyway. The host is the only thing it needs, and that was already in the manifest.

What the commit does:
- Drops the `DiscoveryTunnel` field from the manifest (JSON `tunnel` key) and the `Tunnel: …` markdown line — both the "REQUIRED" and the "none" variants. Tunneled endpoints now render identically to directly-reachable ones (host/port/credential/example only).
- Extends the `tunnel` golden fixture to exercise all four tunnel plugin types (local_command, tailscale, ssh_port_forward, kubernetes_port_forward), each behind its own endpoint. The test now locks down that *no* tunnel type leaks into the manifest, rather than asserting one gets reported.

Note: there's a sibling thread on `simple.md` asking whether the endpoints/credentials *structure* itself serves the agent well — that's a broader manifest-shape question I've left to that thread.

cmd/clawpatrol/testdata/discovery/simple.md
arnauorriols: [bot-reply]
Agreed — that section wasn't pulling its weight. Pushed 32e2c52.

The standalone `## Credentials` section was a credential→endpoint reverse index that just re-stated what each endpoint already lists inline (type, placeholder, connect disambiguators, example invocation). Agents work endpoint-first — they dial a host and read that endpoint's how-to — so the reverse mapping was noise, not leverage.

Change:
- Dropped the `## Credentials` section from the agent-facing markdown manifest (`Markdown()`); regenerated the `.md` goldens (simple/empty/multicred/tunnel).
- Left the JSON output and the `DiscoveryManifest.Credentials` struct field untouched — programmatic consumers and the data model keep the credential-centric view; only the LLM-facing doc is trimmed.

If you'd also like `credentials` gone from the JSON payload, say the word and I'll drop it from the struct too — held off since the comment was on the markdown and JSON consumers may index by credential.

cmd/clawpatrol/testdata/discovery/empty.md
arnauorriols: [bot-reply]
Pushed dc15e45 addressing this. The empty-profile manifest no longer renders a bare "none reachable / none granted" — it now leads with a `## This profile is empty` guidance block that:

- tells the agent its device is mapped to an empty profile, and that this is a configuration state, not an error/broken gateway;
- explains that the profile needs endpoints + credentials bound to it to get value from Claw Patrol;
- points at the dashboard (`gateway.public_url`) where an operator assigns the device a profile or adds grants, telling the agent to ask its human to do so;
- tells the agent to re-fetch `GET https://clawpatrol/` once configured.

`public_url` is mirrored onto `CompiledPolicy` and surfaced as a `dashboard` field on the JSON manifest, set only when the profile grants nothing (non-empty manifests are already actionable, so it'd be noise there). Golden `empty.md`/`empty.json` regenerated; added unit tests for the empty-profile guidance, the no-`public_url` fallback, and that populated profiles don't surface the pointer.

cmd/clawpatrol/testdata/discovery/tunnel.md
```diff
@@ -0,0 +1,48 @@
+# Claw Patrol access manifest — profile: tunneled
+
+You are connected through the Claw Patrol gateway. It intercepts your
+connections transparently: dial the hosts below as you normally would and
+the gateway injects credentials and enforces policy. A credential
+`placeholder` is a literal string you send where the secret would go — the
+gateway swaps it for the real secret. This manifest is scoped to YOUR
+device profile; it lists only what this profile grants.
```
arnauorriols, line 8: we should tell the agent information about the MITM that clawpatrol performs, and the nature of the certs that the agent should expect

cmd/clawpatrol/testdata/discovery/tunnel.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Host(s): k8s-pg.example
+- Port: 5432
+- SSL mode: require
+- Credential: postgres_credential `k8s-rw` — connect with database=prod user=app
+- Example: `psql "host=k8s-pg.example port=5432 user=app dbname=prod sslmode=require"`
+
+### metrics  (clickhouse_native)
+
+- Host(s): ch.example
```
arnauorriols, line 50: Double-check this is the right host that the agent has to dial, considering this connection goes through the tunnel. Verify also the other cases

arnauorriols: [bot-reply]
Verified — the host listed (`ch.example` for `metrics`, and the others) is the correct dial target in every case, including through the tunnel. No change needed.

**Why:** clawpatrol is a transparent intercepting gateway. The agent always dials the *declared endpoint host*; the tunnel is the gateway's own egress concern and is invisible to the agent.

Confirmed end-to-end:

1. **Renderer** (`discovery.go` `describeEndpoint`) pulls hosts only from the endpoint's declared config (`body.Host`/`body.Hosts`) — never from a tunnel's `listen` address. The comment at `discovery.go:120-126` spells out the intent: the tunnel line is deliberately omitted because "the agent dials the host below either way and can't act on the tunnel's name or type."

2. **Data plane** (`main.go:1526-1610`) confirms it's real, not just doc: tunneled endpoints reach upstream via a `dnsvip` VIP allocated for the *declared host* — the agent dials the host, it resolves to a VIP, `handle()` intercepts, and `dialThrough` routes through the `TunnelManager`. Non-tunneled endpoints are intercepted by DNS-resolved dst IP. Either way the dial target is the declared host.

Walking the four tunnel cases in `tunnel.hcl`, all correct:

| endpoint | tunnel type | agent dials | gateway egress (hidden) |
|---|---|---|---|
| `metrics` | tailscale | `ch.example:9440` | tailscale `clawpatrol-tunnel-o11y` → ch.example |
| `prod-pg` | local_command | `main-pg.example:5432` | cloud_sql_proxy on gw `127.0.0.1:5432` |
| `rds-pg` | ssh_port_forward | `rds.example:5432` | ssh via `bastion.example:22` |
| `k8s-pg` | kubernetes_port_forward | `k8s-pg.example:5432` | kube port-forward svc/postgres |

The tempting-but-wrong alternative would be to surface the tunnel's local listen address (e.g. `127.0.0.1:5432` for the `local_command` case). That's the *gateway's* loopback, unreachable from the agent's side of the WireGuard tunnel — so the declared host is exactly right, and it's the same as the direct `github` endpoint by design (that's what the fixture is asserting).

arnauorriols:
[bot-reply]
Pushed e5d2942 addressing this comment:
- Added a "TLS is intercepted" paragraph to the discovery manifest preamble (`DiscoveryManifest.Markdown()`), shown for every profile. It tells the agent that the gateway is a transparent man-in-the-middle terminating TLS for every host, that the cert it sees per upstream is minted on the fly by Claw Patrol's own CA (hostname matches, issuer is the gateway CA — not the host's real public cert), to trust that CA via `https://clawpatrol/ca.crt` and verify the fingerprint at `https://clawpatrol/info`, and that a CA mismatch against the public web PKI is expected here rather than an attack.
- Regenerated the `.md` golden fixtures (empty/simple/multicred/tunnel) via `go test -run TestDiscoveryGolden -update`. JSON goldens are unchanged (the preamble is markdown-only).

cmd/clawpatrol/testdata/discovery/empty.md
```diff
@@ -0,0 +1,40 @@
+# Claw Patrol access manifest — profile: empty
+
+You are connected through the Claw Patrol gateway. It intercepts your
+connections transparently: dial the hosts below as you normally would and
+the gateway injects credentials and enforces policy. A credential
+`placeholder` is a literal string you send where the secret would go — the
+gateway swaps it for the real secret. This manifest is scoped to YOUR
+device profile; it lists only what this profile grants.
+
+TLS is intercepted. The gateway terminates TLS for every host you dial —
```
arnauorriols: This is not completely true. It terminats TLS for every host declared in the profile of the device. Other hosts are passed through without interception. This is relevant because non-intercepted host certificates must still be verified against the public web PKI

cmd/clawpatrol/testdata/discovery/empty.md
```diff
@@ -0,0 +1,40 @@
...
+it is a transparent man-in-the-middle. The certificate you see for an
+upstream is minted on the fly by Claw Patrol's own certificate authority,
+not the host's real public certificate: the hostname matches but the
+issuer is the gateway CA. Trust that CA to verify these connections —
+fetch it from https://clawpatrol/ca.crt and check its fingerprint against
```
arnauorriols: clawpatrol already installs the clawpatrol CA in the device (in the system store, and using envvar pushdown). Explain this

arnauorriols:
[bot-reply]
Pushed d0317657 addressing this comment:
- Reworded the manifest TLS preamble: interception is now scoped to "the hosts this profile grants — the endpoints listed below", matching `g.handle()` which only MITMs profile-bound hosts and splices everything else through untouched.
- Added an explicit passthrough paragraph: every other host is passed through untouched, presents its real upstream certificate, and must still be verified against the public web PKI as usual.
- Regenerated the four discovery golden fixtures (empty/simple/multicred/tunnel .md) via `-update`.

cmd/clawpatrol/testdata/discovery/empty.md
arnauorriols: [bot-reply]
Pushed 706a8d10 addressing this comment.

You're right — the manifest read as if the agent had to set up trust manually. It doesn't: clawpatrol already installs its CA on join, both in the system trust store (`update-ca-certificates` / `security add-trusted-cert`) and via env-var pushdown that `clawpatrol run` sets for wrapped processes (`SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, and similar).

Reworded the TLS section to say so: most clients validate these connections out of the box, and the `https://clawpatrol/ca.crt` fetch + fingerprint check is now framed as a fallback only for clients that ignore both the system store and those env vars. Regenerated the discovery golden fixtures (`empty/simple/multicred/tunnel.md`).

arnauorriols:
We should also list the env vars that are pushed down, so that the agent is aware of them, and what is being pushed to them

arnauorriols:
[bot-reply]
Pushed c4251469 addressing this comment:
- The discovery manifest now lists the **environment variables clawpatrol pushes into the agent's process** for the calling profile, so the agent knows which of its env vars the gateway controls and what each carries.
- Built by mirroring the env-pushdown API (`apiEnvPushdown`): walk the profile's reachable endpoints, take each bound credential's `EnvVars()` first then each endpoint plugin's, dedup by name, sorted for byte-stable output.
- Each entry shows the var name, the value clawpatrol exports (a placeholder/synthetic token, **never** the real secret — swapped at MITM time), a description, and the pushing plugin type. Rendered in both markdown (`## Environment variables (N)`) and JSON (`env_vars`).
- CA-bundle vars (`SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, ...) stay excluded — they're client-side paths, the env-pushdown API omits them too, and the manifest intro already explains the CA install.
- Added an `envvars` golden case (gemini credential → `GOOGLE_API_KEY`/`GEMINI_API_KEY`) plus a unit test asserting listing, scoping, and that a profile with no env-pushdown credential reports none. `go test ./cmd/clawpatrol/`, `gofmt -l .`, and `go vet` all clean.

arnauorriols:
Let's add an optional `description` field in the endpoint and credential blocks, so that users can add there a human/llm readable description of each block. These descriptions should be added into the descovery .md and .json outputs.

arnauorriols:
[bot-reply]
Pushed ef9b50ce addressing this comment:
- Added an optional `description` attribute to both `endpoint` and `credential` blocks (a human/LLM-readable note). It's implemented as a framework-level attr, so it's available on every endpoint and credential type and round-trips through config emit/dump.
- Surfaced descriptions in both discovery outputs:
  - **JSON**: `description` field on each endpoint, on each top-level credential, and on the per-endpoint credential ref.
  - **Markdown**: rendered under each endpoint's heading and inline on each credential line.
- Unified an existing one-off: the `kubernetes` endpoint plugin already had a plugin-local `description` for dashboard display — that's now subsumed by the cross-cutting framework attr (duplicate removed).
- Tests: new `TestDiscoveryDescriptions` plus updated discovery golden fixtures (`simple.{hcl,md,json}`) capturing the rendered output end-to-end.

cmd/clawpatrol/testdata/discovery/multicred.hcl
```diff
@@ -0,0 +1,30 @@
...
+credential "postgres_credential" "pg-rw" {
+  endpoint = postgres.pg
+  user     = "writer"
+  database = "app"
+}
+credential "postgres_credential" "pg-ro" {
+  endpoint = postgres.pg
+  user     = "reader"
+  database = "app"
+}
```
arnauorriols, line 30: these 2 credentials sharing the same endpoint could use a description, this is the use case that gets the most value out of the description field

arnauorriols:
[bot-reply]
Pushed 2969c5c2 addressing this comment:
- Added `description` to both `pg-rw` and `pg-ro` credentials in `multicred.hcl` ("read-write: schema migrations and data fixes" / "read-only: reporting and ad-hoc queries") — exactly the two-creds-on-one-endpoint case where the description field earns its keep.
- Regenerated the `multicred.md` / `multicred.json` goldens so the fixture now exercises per-credential descriptions in both render paths.

cmd/clawpatrol/dnsvip/dnsvip.go
```diff
@@ -85,6 +85,33 @@ var (
 // pathological reload loops.
 const MaxID uint32 = 0xFFFE
 
+// DiscoveryHostname is the reserved name the gateway answers locally
+// with a manifest of the caller's profile (see cmd/clawpatrol's
+// discovery endpoint). The allocator hands back a fixed VIP for it so
+// `curl https://clawpatrol/` resolves inside the tunnel; the WG
+// forwarder then routes the :443 SYN to the local discovery handler
+// rather than any upstream.
+const DiscoveryHostname = "clawpatrol"
```
arnauorriols: This should be clawpatrol.internal, and it should be the canonical entrypoint for the devices to the API of clawpatrol. Then, to get the manifest, they should do https://clawpatrol.internal/manifest. To get the ca, they should do https://clawpatrol.internal/ca.crt. etc

arnauorriols: [bot-reply]
Pushed 9c110e87 addressing this comment:
- Renamed the reserved discovery name from the bare label `clawpatrol` to the fully-qualified `clawpatrol.internal` (`.internal` private-use TLD — no search-domain/dotless-name surprises), in both `dnsvip.DiscoveryHostname` and `discovery.go`'s `discoveryHostname`.
- Made the in-tunnel discovery endpoint a path-routed API surface so `clawpatrol.internal` is the canonical device entrypoint:
  - `GET /` and `GET /manifest` → the caller's profile manifest
  - `GET /ca.crt` → the gateway CA in PEM (mirrors the web server's `serveCA`)
  - `GET /info` → liveness + CA fingerprint (mirrors `serveInfo`)
  - anything else → 404
- Previously the handler returned the manifest at *every* path, so the manifest's own `/ca.crt` and `/info` links resolved to the manifest itself; updated the manifest copy and the discovery golden fixtures to the new host/paths.

`go test ./cmd/clawpatrol/ ./cmd/clawpatrol/dnsvip/`, `go vet`, and `gofmt -l` all clean.

cmd/clawpatrol/discovery.go
```diff
@@ -68,11 +69,67 @@ func (g *Gateway) serveDiscovery(c net.Conn, certHost string) {
...
 
+// routeDiscovery dispatches a request to the in-tunnel discovery
+// entrypoint by path. clawpatrol.internal is the canonical device-facing
+// API surface, so it exposes the profile manifest at / and /manifest,
+// the gateway CA at /ca.crt, and a liveness + CA-fingerprint blob at
+// /info — the same public endpoints the gateway's tailnet web server
+// serves, mirrored here so a device with only tunnel reachability can
+// fetch them by name. Unknown paths 404 rather than falling through to
+// the manifest, so the canonical paths stay unambiguous.
+func (g *Gateway) routeDiscovery(w http.ResponseWriter, r *http.Request, policy *config.CompiledPolicy, profile string) {
```
arnauorriols: This no longer is only about "discovery". find a better name. Also change `discoveryHostname`

arnauorriols:
Fix the tests

cmd/clawpatrol/discovery.go
arnauorriols: [bot-reply]
Pushed 45c92905 addressing this comment.

You're right — the `clawpatrol.internal` endpoint is no longer just discovery; it's the canonical device-facing API surface (it routes `/manifest`, `/ca.crt`, and `/info`). I renamed the host + entrypoint/serving/VIP layer from `discovery*` to `internal*`, tying the names to the literal `.internal` hostname:

- `discovery.go`: `discoveryHostname`→`internalHostname`, `isDiscoveryHost`→`isInternalHost`, `serveDiscovery`→`serveInternal`, `routeDiscovery`→`routeInternal`, `serveDiscoveryCA`→`serveInternalCA`, `serveDiscoveryInfo`→`serveInternalInfo`, `isDiscoveryVIP`→`isInternalVIP` (plus the otel/log labels).
- `dnsvip`: `DiscoveryHostname`→`InternalHostname`, `DiscoveryVIPs`→`InternalVIPs`, `discoveryID`→`internalID`, `isDiscoveryName`→`isInternalName`.
- `main.go` call sites and the affected tests updated.

I deliberately left the manifest *data model* (`DiscoveryManifest`, `DiscoveryEndpoint`, `DiscoveryCredential`, `buildDiscoveryManifest`, …) named as-is: that payload is still genuinely a discovery manifest — it's now just one of the things the internal API serves. If you'd rather rename those too (e.g. to a neutral `Manifest`/`ManifestEndpoint`), say the word and I'll follow up.

`gofmt -l`, `go vet`, and `go test ./cmd/clawpatrol/ ./cmd/clawpatrol/dnsvip/` all clean.

arnauorriols:
[bot-reply]
Pushed 311a209b addressing this comment:
- Regenerated `site/doc/config-reference.md` so `TestGeneratedDocIsFresh` passes. The k8s endpoint's plugin-local `description` field was subsumed by the new framework-level `description` attr (ef9b50ce), but the committed doc still listed the stale plugin row. Removing it makes the committed doc match `go run ./internal/tools/docgen`.

Full `go test ./...` is green locally (the only other "failures" were `dashboard/dist` embed errors, which CI resolves via `make dashboard` before running tests).

cmd/clawpatrol/testdata/discovery/multicred.hcl
arnauorriols: [bot-reply]
Pushed 2969c5c2 addressing this comment:
- Added per-credential `description` fields to both postgres credentials sharing the `postgres.pg` endpoint in `multicred.hcl` (`pg-rw` → "read-write: schema migrations and data fixes", `pg-ro` → "read-only: reporting and ad-hoc queries"), since user/database is the only disambiguator — exactly the use case that gets the most value from the description field.
- Regenerated the golden `multicred.json` and `multicred.md` so the discovery fixture exercises per-credential descriptions in both render paths.

PR #646, opened 2026-06-09
fix: honor wireguard.listen_port instead of silently ignoring it (cl-94cf)

arnauorriols:
## Problem

`wireguard.listen_port` was silently ignored — setting it had no effect on the port WireGuard actually bound, with no error or warning (cl-94cf).

## Root cause

The field was parsed and plumbed all the way into `JoinConfig.WGListenPort`, then dropped: `StartWGServer` (`cmd/clawpatrol/wireguard.go`) derived the bind port from the `endpoint` string (or the 51820 default) and never read `WGListenPort`.

## Fix

- Bind the configured `listen_port` (default 51820) via a small `wgBindPort` helper.
- The `endpoint`'s port is purely **client-advertised** and may legitimately differ from the bind port under NAT/split-host, so it no longer drives the server bind.
- A non-honorable port (in use/privileged) now fails loudly: `IpcSet` returns `IpcErrorPortInUse` and the caller `log.Fatalf`s it — no more silent fallback.
- `wgClientEndpoint` also defaults the advertised port to `listen_port`, so a single-host config that sets only `listen_port` advertises the matching port.

## Tests

- `TestStartWGServerHonorsListenPort` — boots the device and asserts the bound port via `IpcGet` (the end-to-end acceptance guard).
- `TestWGBindPort` — default/override resolution.
- Expanded `TestWGClientEndpoint` — `listen_port` → advertised port, and endpoint-port-overrides-listen_port (split-host NAT).

Full `cmd/clawpatrol` suite, `gofmt -l`, and `go vet` all pass.

### DIFF
--- a/cmd/clawpatrol/tailscale.go
+++ b/cmd/clawpatrol/tailscale.go
@@ -13,7 +13,8 @@
 // through the WG netstack's promiscuous forwarder; main.go's
 // tcpDispatch routes dst port 443 to g.handle. Alongside the
 // netstack we also open a loopback TCP listener on 127.0.0.1:8443
-// for host-local clients — single-host deployments (the gateway
+// (configurable via wireguard.host_loopback_port) for host-local
+// clients — single-host deployments (the gateway
 // running under one user account, clawpatrol-run invoked from
 // another on the same machine, loopback WG between them) are a
 // first-class pattern, not a debug mode.
@@ -64,12 +65,28 @@ func gatewayTsnetDir(stateDir string) (string, error) {
 	return dir, nil
 }
 
+// defaultHostLoopbackPort is the TCP port the gateway binds on
+// 127.0.0.1 for host-local clients when wireguard.host_loopback_port
+// is unset.
+const defaultHostLoopbackPort = 8443
+
+// hostLoopbackPort resolves the loopback TCP port the gateway binds
+// for host-local clients, honoring wireguard.host_loopback_port and
+// falling back to defaultHostLoopbackPort when unset (0).
+func hostLoopbackPort(port int) int {
+	if port != 0 {
+		return port
+	}
+	return defaultHostLoopbackPort
+}
+
 // openListener brings up the gateway's transport listeners. Either
 // or both of the returned values may be non-nil depending on which
 // transport blocks the operator declared:
 //
 //   - WireGuard enabled → returns a loopback TCP listener on
-//     127.0.0.1:8443. Host-local agents (e.g. clawpatrol-run from a
+//     127.0.0.1:<host_loopback_port> (default 8443). Host-local
+//     agents (e.g. clawpatrol-run from a
 //     different UID on the same box) connect directly. Off-host
 //     agents reach g.handle via the WG netstack's promiscuous
 //     forwarder; that path doesn't touch this socket.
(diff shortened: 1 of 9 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR body. It is a descriptive summary of the `wireguard.listen_port` fix (problem / root cause / fix / tests) authored by the PR submitter — it contains no question or change request directed at the implementer, so no code change is warranted.

No action taken. If a reviewer wants something changed, please leave a specific review comment and the watcher will dispatch a follow-up.

arnauorriols:
I don't see any doc changes. Double check no doc changes are needed or update the docs.

arnauorriols:
[bot-reply]
Double-checked — no doc changes are needed, and adding any would make the docs *wrong*. The relevant docs already describe exactly the behavior this PR implements; the bug was that the code silently contradicted them.

What the docs on `main` already say:

- `site/doc/config-reference.md` — `listen_port`: *"The UDP port the gateway binds for WG peers. Default 51820."* and `endpoint`: *"port defaults to listen_port."*
- `site/doc/configure-gateway.md` — *"The default WireGuard listener is `:51820` (set `wireguard.listen_port` to override)"*, with the example `listen_port = 41820  # server binds this UDP port`.
- `examples/gateway.example.hcl` — *"listen_port defaults to 51820. Clients dial host(public_url): port(endpoint||listen_port)."*

That is precisely the contract this PR enforces:
- `StartWGServer` now binds `wgBindPort(WGListenPort)` (configured port, default 51820) instead of deriving the bind port from the endpoint string.
- `wgClientEndpoint` now defaults the advertised port to `listen_port`, matching the documented `port(endpoint||listen_port)` precedence.

So the docs were already correct; this is a pure code fix bringing the implementation in line with them. Nothing to update.

arnauorriols:
Follow-up suggestion (not blocking): consider extending the same configurability to the host-local TCP listener.

While verifying this PR I had two clawpatrol gateways on the box (one running, one under test) — `listen_port` from this PR meant their WG UDP ports could now be disjoint, but the hardcoded `127.0.0.1:8443` in `cmd/clawpatrol/tailscale.go:87` still collided, so the second gateway exited with `bind: address already in use` even with distinct `dashboard_listen`, `state_dir`, `subnet_cidr`, and `listen_port`. Ended up running the verification under `unshare -Urn` to dodge it.

It's the symmetric gap: the WG bind port is now operator-controlled, but the host-local TLS landing pad that single-host deployments use to reach `g.handle` / dashboard mux isn't. Same shape of fix — a `host_loopback_port` (or `host_loopback_listen`) field on `WireGuardBlock`, default `8443`, plumbed into `openListener`. No client code dials 8443 from a literal (checked: only references are the listener itself and comments), so it's a single-file change plus config plumbing.

Filing as a suggestion here since the operator story ("run two gateways on one host" — dev/test, blue-green, multi-tenant) only really works once *both* hardcoded ports are configurable. Happy to open a bead if useful.

arnauorriols:
[bot-reply]
Pushed 17975e8 addressing this follow-up — the host-local TCP landing pad is now operator-controlled, symmetric with `listen_port`:

- Added `host_loopback_port` (default 8443) to `WireGuardBlock`, plumbed into `openListener` via a `wgBindPort`-style `hostLoopbackPort` helper (replaces the hardcoded `127.0.0.1:8443`).
- Round-tripped through config emit/dump, regenerated `site/doc/config-reference.md`, and documented it in `examples/gateway.example.hcl`.
- Tests: `TestHostLoopbackPort` (default/override) and `TestOpenListener_WireGuardHonorsHostLoopbackPort` (boots `openListener` with an explicit port and asserts the listener bound it).

Two gateways on one host can now set disjoint `host_loopback_port` (plus `listen_port`) and coexist without the `bind: address already in use` collision — no more `unshare -Urn` workaround.

PR #647, opened 2026-06-10
macos: idle network extension black-holes internet after sleep/wake (cl-jpya)

arnauorriols:
## Problem

On macOS, the clawpatrol network extension black-holes **all** internet connectivity after a lid-close/sleep/wake cycle — **even when no agent (`clawpatrol run`) is active**. Disabling the extension immediately restores connectivity. The extension is supposed to be a transparent no-op when idle.

## Root cause

The extension is a `NETransparentProxyProvider` that claims **all** outbound TCP+UDP (`includedNetworkRules`). The idle pass-through relies on:
- TCP → `handleNewFlow` returns `false` → kernel passthrough
- UDP → `bypassUDP()` relays via a fresh host socket per flow

There are **no** routes/DNS/utun installed (the transparent proxy is flow-based), so nothing dangles.

The defect is the sleep/wake state machine added in #187. `sleep()` **unconditionally** set `reasserting = true`. A transparent proxy in the reasserting state blocks every flow it has claimed — including `bypassUDP` DNS (the same mechanism #187's own commit message documents: *"a reasserting loop that blocks ALL traffic — including bypassUDP DNS"*). The flag is cleared only by `wake()`, and macOS does **not** reliably deliver `wake()` after a lid-close / full-sleep cycle. A missed `wake()` strands the proxy in `reasserting` forever → all UDP (incl. DNS) blocked → total black-hole until the user disables the extension.

## Fix

Only enter the reasserting state when there is something **actually tunneled** to protect: whole-machine mode, or per-process mode with a live session registered. A pure pass-through (per-process, no agent) has no in-memory tunnel state worth preserving across a restart, so it never reasserts and survives sleep/wake as a transparent no-op — whether macOS keeps the extension running (bypassUDP/passthrough recover on their own) or restarts it (`startProxy` rebuilds clean state; it no longer blocks, per #187).

`wake()` still clears `reasserting` defensively.

## Acceptance criteria

- [x] Idle (no agent) + sleep/wake → never enters reasserting → connectivity intact.
- [x] Agent running → restart still suppressed while a session is registered → interception unchanged across sleep/wake.
- [x] No dangling routes/DNS/utun (transparent proxy installs none).

⚠️ Swift-only change; builds in macOS CI (no Swift toolchain on the Linux runner). Cannot be exercised on the dev box — needs a real macOS sleep/wake verification.

### DIFF
--- a/macos/ClawpatrolExtension/Provider.swift
+++ b/macos/ClawpatrolExtension/Provider.swift
@@ -240,24 +240,47 @@ class TransparentProxyProvider: NETransparentProxyProvider {
         completionHandler()
     }
 
-    // Sleep/wake — without these overrides macOS may restart the
-    // extension when the network interface comes back up after sleep,
-    // causing startProxy to be called again. Setting reasserting=true
-    // during sleep tells the runtime the tunnel is temporarily down;
-    // clearing it on wake avoids a full stop/start cycle.
+    // Sleep/wake. `reasserting = true` tells macOS the tunnel is
+    // temporarily down so it won't tear the extension process down (and
+    // lose the in-memory session registry) when the primary interface
+    // bounces on wake. macOS only restarts the proxy on interface change
+    // when we're NOT reasserting, so this suppresses the restart.
+    //
+    // But it is a trap for the idle case. A NETransparentProxy in the
+    // reasserting state blocks every flow it has claimed, and we claim
+    // ALL outbound UDP (see applyNetworkSettings) — so while reasserting
+    // is set, bypassUDP DNS is blocked too (this is the same mechanism
+    // #187 documents). The flag is cleared only by wake(), and macOS does
+    // NOT reliably deliver wake() after a lid-close / full-sleep cycle.
+    // A missed wake() then leaves the proxy reasserting forever, black-
+    // holing ALL internet — even with no agent running, when the
+    // extension is supposed to be a transparent no-op. The only escape is
+    // disabling the extension (cl-jpya).
+    //
+    // So only enter that fragile state when there is something actually
+    // tunneled to protect: whole-machine mode, or per-process mode with a
+    // live session registered. A pure pass-through (per-process, no agent)
+    // has no in-memory tunnel state worth preserving across a restart, so
+    // it must never reassert — an idle extension survives sleep/wake as a
+    // transparent no-op no matter whether macOS keeps it running
+    // (bypassUDP/passthrough recover on their own) or restarts it
+    // (startProxy rebuilds a clean state; it no longer blocks, per #187).
     override func sleep(completionHandler: @escaping () -> Void) {
-        reasserting = true
+        if wholeMachine || !sessionPids().isEmpty {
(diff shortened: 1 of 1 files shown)


arnauorriols:
[bot-reply]
Reviewed the change against the code on `polecat/guzzle-mq7q67l7` — no changes requested. The root-cause analysis and fix check out:

- **Root cause confirmed.** `applyNetworkSettings` claims all outbound TCP+UDP, and a `NETransparentProxy` in `reasserting` blocks every claimed flow — so `reasserting = true` does black-hole `bypassUDP` DNS, matching the mechanism #187's commit message documents. `wake()` is the only clearer, and a missed `wake()` after lid-close strands it forever.
- **Fix is correct and minimal.** Gating on `wholeMachine || !sessionPids().isEmpty` matches exactly the cases `shouldTunnel` would actually tunnel: whole-machine always, per-process only when a session pid is registered. Idle per-process (no agent) tunnels nothing, so reasserting protects nothing — correct to skip it.
- **Signal is reliable.** `sessionPids()` is lock-guarded and kept honest by `sessionRegister`/`sessionUnregister` plus the 5s reaper that drops SIGKILLed CLIs, so the idle vs. active distinction won't go stale.
- The defensive `wake()` clear (always reset `reasserting`) is a sensible belt-and-suspenders even though the idle path no longer sets it.

One caveat, already flagged in the body: this is Swift-only, builds in macOS CI but can't be exercised on the Linux dev box — still needs a real macOS sleep/wake verification before it can be called done.

piscisaureus:
Reviewed this against current main. The three-line change itself is safe for the active-session path and rebases cleanly, but I don't think it can land on its current basis:

- The explanation (a `reasserting` idle extension black-holes claimed flows, and `wake()` is not reliably delivered) is a hypothesis; nothing in the PR or the cl-jpya report shows extension logs for a missed `wake()`. Meanwhile #762 landed after this PR with measurements of a different cause for the same symptom: one leaked host UDP socket per BypassUDP flow until the ephemeral range is exhausted, after which all UDP dies until the extension process is recycled. Long uptime is the trigger, and sleep/wake correlates with long uptime.
- "startProxy no longer blocks, per #187" only holds for WG mode. In tsnet mode `startProxy` runs `ts_netstack_init` before applying network settings, so if the premise is right, an idle tsnet-mode extension would now restart after every wake and sit in the connecting state for as long as the tailnet join takes.
- `sleep()` logs nothing, so a maintainer can't tell from `log show` whether the flag was set, whether `wake()` arrived, or whether the extension restarted. A one-line `os_log` there is the most useful change in this area regardless.

Suggested path: re-reproduce on a build that includes #762 (lid-close for 10+ minutes, then `log show --predicate 'subsystem == "dev.clawpatrol.app.extension"'` and `pmset -g log`). If it still reproduces and the logs show the missed `wake()`, add the `sleep()` log line, fix the tsnet-mode claim in the comment, and this is good to go. If it doesn't reproduce, we close it as fixed by #762.

PR #651, opened 2026-06-10
feat(credentials): add Amplitude OAuth credential

dhruvkelawala:
## Summary

Adds a built-in `amplitude_oauth` credential for Amplitude's hosted MCP server and the `amp` CLI.

The credential:

- injects `Authorization: Bearer <gateway OAuth token>` for Amplitude MCP requests
- pushes `AMPLITUDE_ACCESS_TOKEN`, `AMPLITUDE_OAUTH_TOKEN`, and `AMPLITUDE_REGION` placeholders for `amp`
- supports `region = "us" | "eu"`
- uses Amplitude MCP dynamic client registration (`/register`) with PKCE
- refreshes tokens through the gateway OAuth registry with the persisted dynamic `client_id`

## OAuth callback model

Amplitude is stricter than Notion MCP about redirect URIs: it rejects `http://<tailnet-ip>:8080/oauth/callback` and only allows insecure HTTP for localhost. To keep this working without requiring operators to expose the dashboard over HTTPS, this credential pins the redirect URI to the same loopback shape used by `amplitude-cli`:

```text
http://localhost:8900/callback
```

The browser may land on a connection-failed localhost page, but the dashboard's existing copy/paste exchange flow can complete the connection. I also made the exchange handler accept a pasted full callback URL, not only the raw code, to make that fallback less fragile.

I would appreciate maintainer guidance on whether this should stay credential-specific, or whether Claw Patrol should grow a first-class callback model for dynamic OAuth providers, e.g. provider-declared callback strategy, `gateway.public_url` preference, Funnel HTTPS fallback, or dashboard-assisted loopback handling.

## Validation

- `XDG_RUNTIME_DIR=/proc go test ./cmd/clawpatrol ./internal/config/plugins/credentials ./internal/tools/docgen`
- `XDG_RUNTIME_DIR=/proc go test ./...`
- `go build ./cmd/clawpatrol`
- Live Amplitude EU dynamic registration probe accepted `http://localhost:8900/callback`
- Tested on a gateway: dashboard connect succeeded, then `clawpatrol run -- amp auth status` and `clawpatrol run -- amp events list` succeeded with placeholder env vars; direct `amp` with the placeholder failed as expected.

### DIFF
--- a/cmd/clawpatrol/oauth.go
+++ b/cmd/clawpatrol/oauth.go
@@ -347,11 +347,11 @@ func (s *oauthState) setToken(tok *oauth2.Token) {
 		// (returns "Invalid request format" otherwise). Stdlib oauth2
 		// only sends form-urlencoded.
 		base = &anthropicRefreshSource{cfg: s.cfg, current: tok}
-	case isNotionMCPTokenURL(s.cfg.Endpoint.TokenURL):
-		// Notion's MCP token endpoint refreshes via form-urlencoded body
-		// and expects the dynamically registered client_id (no static
+	case isDynamicMCPTokenURL(s.cfg.Endpoint.TokenURL):
+		// Hosted MCP token endpoints refresh via form-urlencoded body and
+		// expect the dynamically registered client_id (no static
 		// ClientSecret — PKCE-only public client).
-		base = &notionMCPRefreshSource{cfg: s.cfg, current: tok}
+		base = &dynamicMCPRefreshSource{cfg: s.cfg, current: tok}
 	default:
 		base = s.cfg.TokenSource(context.Background(), tok)
 	}
@@ -535,7 +535,7 @@ func (r *OAuthRegistry) loadFromDB() error {
 		}
 		s := newState(it, r.db)
 		// Restore the dynamically-registered client_id BEFORE setToken
-		// so the refresh source (notion_mcp) picks it up via s.cfg.
+		// so dynamic MCP refresh sources pick it up via s.cfg.
 		if clientID.Valid && clientID.String != "" {
 			s.clientID = clientID.String
 			s.cfg.ClientID = clientID.String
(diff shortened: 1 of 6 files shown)


piscisaureus (commented):
Thanks for this — it's a well-executed contribution, and the PR description's validation notes made it easy to review. Generalizing the `notion_mcp` flow into `dynamic_mcp` is the right structure, and the back-compat handling (old flow string still routed, empty `RedirectURI` falling back to the dashboard callback) is done correctly. I checked out the branch locally and confirmed gofmt, `go build ./...`, docgen, and the test suites all pass. The pinned localhost redirect is also safer than it first looks since PKCE keeps the verifier on the gateway, so a stray local listener on 8900 can't do anything with the code.

We'll land this once the following is addressed:

**Blocking:**

* Add `amplitude_oauth` to `dashboard/src/lib/credentialLabels.ts` (label plus a `third_party` category entry), and ideally a logo case in `Logos.tsx`. Without it the settings card shows the raw type name under "other" — every other built-in credential has an entry.

**Cheap to do now, but fine as follow-ups:**

* `normalizeOAuthExchangeInput`: a pasted callback URL without a scheme (`localhost:8900/callback?code=...&state=...`) misses the `http(s)://` branch, falls into the plain-code path, and gets truncated at `?` — the user then sees an opaque upstream exchange error. Treating any input containing `?code=` as a URL would close this.
* The `httptest` handlers in the new tests call `t.Fatalf` from the server goroutine, where it doesn't actually stop the test; use `t.Errorf` (plus an early `http.Error`) inside handlers.
* `isDynamicMCPTokenURL` is a hardcoded host list that every future dynamic-MCP provider has to remember to update, far from the plugin that defines the provider. The registry already persists and restores the dynamic `client_id`, so dispatching the refresh source on that (or on the stored flow string) would make the next provider zero-touch. A TODO is fine for now.

On the callback-model question you raised: keeping the pinned redirect URI as plugin data is right for now — Amplitude is the first provider that needs it, and the field already existed. A first-class strategy (`gateway.public_url` preference, Funnel HTTPS) is worth designing when a second provider forces it. One ask in the meantime: mention the copy/paste-the-callback-URL step in the plugin's doc comment so docgen surfaces it in the config reference — it's surprising on first run and the `region` attribute is currently the only operator-facing documentation.

dhruvkelawala:
This PR can be closed in favour of #656

PR #656, opened 2026-06-10
feat(pluginsdk): support external HTTP credential plugins

dhruvkelawala:
## Summary

This PR adds generic support for **external HTTP credential plugins**.

External plugins can now expose the same credential capabilities that previously required built-in Go code:

- dashboard secret slots
- agent environment pushdown placeholders
- OAuth flow metadata
- request-time HTTP header injection through the built-in HTTPS endpoint
- plugin-returned redactions for derived/injected secrets

The goal is to let provider-specific integrations live out of tree while keeping Claw Patrol's upstream runtime generic.

## Motivation

Some credentials need more than static bearer/header injection, but they should not all become built-ins.

Two motivating examples:

1. **Amplitude MCP OAuth**
   - declares a dynamic MCP OAuth flow
   - pushes placeholders expected by the `amp` CLI
   - lets the gateway OAuth registry own token storage/refresh
   - injects the current access token into Amplitude MCP HTTPS requests

2. **Customer.io service accounts**
   - stores a durable service-account token only on the gateway
   - exchanges it gateway-side for short-lived JWTs
   - injects the JWT into Customer.io HTTPS requests
   - keeps raw service-account material out of agent env/files

This PR adds the generic SDK/runtime surface needed for both without adding Amplitude or Customer.io provider logic to core.

## Architecture walkthrough

Visual explainer:

https://clawpatrol-external-credentials.vercel.app

It covers the config-load path, request-time injection path, OAuth path, security model, and the two reference POCs.

### Plugin SDK

`pluginsdk.CredentialDef` can now declare HTTP injection support and return instance-specific metadata from `Build` via `pluginsdk.CredentialBuildResult`:

- `SecretSlots`
- `EnvVars`
- `OAuth`
- `Disambiguators`
- `HTTPInject`

External credentials can also implement:

```go
InjectHTTP(ctx context.Context, req pluginsdk.HTTPInjectRequest) (*pluginsdk.HTTPInjectResponse, error)
```

The request includes credential identity, canonical config, gateway-held secret bytes/extras, and HTTP request metadata. The response returns header mutations plus redaction strings.

### External plugin adapter

The external-plugin adapter maps plugin metadata onto the existing internal credential interfaces:

- secret slot provider
- env pushdown provider
- OAuth flow provider
- HTTP credential runtime
- HTTP redaction provider

This lets the rest of the gateway continue to use the same abstractions as built-in credentials.

### HTTPS credential dispatch

The built-in HTTPS endpoint can now route placeholder-matched requests to external HTTP credentials.

At request time the gateway:

1. detects the credential placeholder
2. fetches the gateway-held secret
3. calls the external plugin's `InjectHTTP` RPC
4. applies returned header mutations
5. consumes returned redactions for audit/sample redaction

### OAuth improvements

Dynamic MCP OAuth is now selected by the credential flow (`dynamic_mcp`) rather than provider-hostname-specific checks. This makes external MCP OAuth providers possible.

This branch also fixes a few dynamic-MCP edge cases discovered during closeout review:

- authorization-code exchange uses public-client form params instead of first attempting HTTP Basic auth
- manual callback parsing accepts raw query strings even when `state` appears before `code`
- loopback callback URLs can be pasted directly into the dashboard exchange field

### Documentation

`site/doc/plugins.md` now documents external credential metadata, HTTPS injection, runtime privilege split, dynamic MCP flow selection, and provider-specific exchange guidance.

## Security model

External credential plugins are **trusted gateway-side code**.

The security boundary remains:

- wrapped agents receive placeholders only
- real secrets stay on the gateway
- the gateway sends credential secret bytes only to trusted local plugin processes
- provider-specific exchange/injection logic happens on the gateway side
- returned derived secrets can be redacted from request samples/audit views

`InjectHTTP` is intentionally header-only. External credential plugins cannot rewrite the destination URL or request body through this hook.

## Reference POCs

Two external plugins were built and exercised against this branch:

- Customer.io service-account credential plugin  
  https://github.com/dhruvkelawala/clawpatrol-customerio-plugin

- Amplitude MCP OAuth credential plugin  
  https://github.com/dhruvkelawala/clawpatrol-amplitude-plugin

They demonstrate two different external credential patterns:

1. dashboard secret slot + plugin-owned token exchange + HTTP injection
2. plugin-provided OAuth metadata + gateway OAuth registry + HTTP injection

Both were tested end-to-end through Claw Patrol gateway mode with real CLIs while keeping provider credentials gateway-held.

## Validation

Core branch:

```bash
XDG_RUNTIME_DIR=/proc go test -count=1 ./...
cd dashboard
deno task build
deno task lint
deno task format:check
```

External plugin POCs:

```bash
cd clawpatrol-customerio-plugin
go test -count=1 ./...
go build -o clawpatrol-customerio-plugin .

cd clawpatrol-amplitude-plugin
go test -count=1 ./...
go build -o clawpatrol-amplitude-plugin .
```

Structured autoreview was run on:

- the Customer.io plugin
- the Amplitude plugin
- this implementation branch

Accepted findings from review were fixed before opening this PR.

## Notes for reviewers

- This intentionally avoids provider-specific upstream code.
- The POC repositories are referenced as validation/examples, not as dependencies of Claw Patrol.
- External plugins should be treated like gateway extensions: operators should load only trusted binaries and protect plugin install paths like the gateway binary itself.

### DIFF
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -10,6 +10,10 @@ concurrency:
 
 jobs:
   preview:
+    # Demo previews need DEMO_DEPLOY_TOKEN to read/write denoland/clawpatrol-demo.
+    # GitHub does not expose repository secrets to pull_request runs from forks,
+    # so skip this optional preview job there instead of failing fork PR CI.
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -10,10 +10,6 @@ concurrency:
 
 jobs:
   preview:
-    # Demo previews need DEMO_DEPLOY_TOKEN to read/write denoland/clawpatrol-demo.
-    # GitHub does not expose repository secrets to pull_request runs from forks,
-    # so skip this optional preview job there instead of failing fork PR CI.
-    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/cmd/clawpatrol/extplugin_http_credential_test.go
+++ b/cmd/clawpatrol/extplugin_http_credential_test.go
@@ -157,6 +157,9 @@ rule "allow-api" {
 	if strings.Contains(fmt.Sprint(end.ReqHeaders), "real-external-token") {
 		t.Fatalf("request headers audit leaked injected external credential: %#v", end.ReqHeaders)
 	}
+	if strings.Contains(fmt.Sprint(end.ReqHeaders), "derived-real-external-token") {
+		t.Fatalf("request headers audit leaked derived external credential: %#v", end.ReqHeaders)
+	}
 	if got := end.ReqHeaders["X-Magic"]; got != credentialSampleRedaction {
 		t.Fatalf("X-Magic audit header = %q, want credential redaction marker", got)
 	}
@@ -351,12 +354,13 @@ func main() {
 				if got := req.Headers.Get("Authorization"); !strings.HasPrefix(got, "Bearer PH_") {
 					return nil, fmt.Errorf("authorization = %%q", got)
 				}
+				derived := "derived-" + string(req.CredentialSecret)
 				return &pluginsdk.HTTPInjectResponse{
 					Headers: []pluginsdk.HeaderMutation{
 						{Op: pluginsdk.HeaderSet, Name: "Authorization", Values: []string{"Bearer " + string(req.CredentialSecret)}},
-						{Op: pluginsdk.HeaderSet, Name: "X-Magic", Values: []string{string(req.CredentialSecret)}},
+						{Op: pluginsdk.HeaderSet, Name: "X-Magic", Values: []string{derived}},
 					},
-					Redactions: []string{string(req.CredentialSecret)},
+					Redactions: []string{string(req.CredentialSecret), derived},
 				}, nil
 			},
 		}},
--- a/cmd/clawpatrol/extplugin_http_credential_test.go
+++ b/cmd/clawpatrol/extplugin_http_credential_test.go
@@ -1,0 +1,396 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/denoland/clawpatrol/internal/config"
+	"github.com/denoland/clawpatrol/internal/config/extplugin"
+	_ "github.com/denoland/clawpatrol/internal/config/plugins/all" // register built-in plugins
+	"github.com/denoland/clawpatrol/internal/config/runtime"
+)
+
+type externalHTTPTestSecretStore map[string]string
+
+func (s externalHTTPTestSecretStore) Get(name string) (runtime.Secret, error) {
+	v, ok := s[name]
+	if !ok {
+		return runtime.Secret{}, fmt.Errorf("unexpected secret lookup %q", name)
+	}
+	return runtime.Secret{Bytes: []byte(v)}, nil
+}
+
+func TestExternalCredentialInjectsAuthorizationThroughBuiltInHTTPS(t *testing.T) {
+	typeName := fmt.Sprintf("ext_https_cred_%d", time.Now().UnixNano())
+	pluginPath := buildExternalHTTPTestPlugin(t, typeName)
+
+	mgr := extplugin.New(nil)
+	config.SetPluginLoader(mgr)
+	t.Cleanup(func() {
+		mgr.Stop()
+		config.SetPluginLoader(nil)
+	})
+
+	gw, diags := config.LoadBytes([]byte(fmt.Sprintf(`
+plugin "extcredtest" { source = %q }
+
+gateway {
+  state_dir  = "/tmp/clawpatrol-test"
+  public_url = "https://gw.example.test"
+  wireguard { subnet_cidr = "10.55.0.0/24" }
+}
+endpoint "https" "api" {
+  hosts = ["api.example.test"]
+}
+credential %q "external" {
+  endpoint    = https.api
+  placeholder = "PH_external"
+}
+profile "default" { credentials = [%s.external] }
+rule "allow-api" {
+  endpoint = https.api
+  verdict  = "allow"
+}
+`, pluginPath, typeName, typeName)), "external-http-credential-test.hcl")
+	if diags.HasErrors() {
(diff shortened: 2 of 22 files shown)


piscisaureus (commented):
Solid, well-scoped design. The endpoint/tunnel/facet model is genuinely generic, the "header-only" injection contract is enforced by the wire shape (not just docs), CSRF/state and PKCE in the OAuth path are intact, backward compat is handled (optional func fields + additive proto3), and gofmt is clean. No agent-facing secret-exfil bypass found. The items below are what I'd want before merge, then a follow-up for the extensibility gaps.

## Part 1 — Address before merge

**Correctness / security**

- **Derived-secret redaction is opt-in and untested** (`internal/config/extplugin/adapter.go`). The gateway auto-redacts the *raw* secret bytes, but a plugin that injects a **derived** value (an exchanged JWT, an HMAC signature) must return that exact string in `Redactions` or it lands verbatim in the audit sample — `ev.ReqHeaders` only auto-masks header *names* matching `auth|token|secret|key|...`, so a derived secret in `X-Signature` leaks. This is the core promise of the feature, and the integration test only injects the *raw* secret, so the derived path has zero coverage. Add a derived-value plugin test and document the redaction contract prominently.
- **Injection fails open, and the nil-client case is silent** (`main.go:2419-2421`, `adapter.go:856-859`). On `InjectHTTP` error the request is forwarded with the placeholder header (logged); on a nil/dead plugin client `injectHTTPWithExternalCredential` returns `nil` (success) and no-ops with *no log at all*. No real secret leaks (upstream just 401s), and it's consistent with the existing sign/secret-fetch behavior — but at minimum log the nil-client path, and make a deliberate call on fail-open vs fail-closed for a credential-bound request.
- **No panic recovery in plugin RPC handlers** (`pluginsdk/server.go`). A nil deref in one plugin's `InjectHTTP`/`Build`/`HandleConn` callback aborts the whole subprocess, dropping every concurrent connection it serves — not just the offending RPC. Wrap each dispatch in `defer/recover`.

**Nits**

- **No header name/value validation in `applyHeaderMutations`** (`adapter.go:952`). `m.Name`/`m.Values` go straight to `h.Del`/`h.Add`; a `"X-Foo\r\nInjected: 1"` relies entirely on `net/http` rejecting it at write time (it does). Cheap defense-in-depth to reject invalid field names at the proxy layer.
- **`ADD` leaves the agent's placeholder header** (`adapter.go:963`). `SET` does `Del` then `Add`; `ADD` only appends, so an `ADD` on `Authorization` forwards both the placeholder and the real value upstream. Document that auth injection should use `SET`.
- **`normalizeOAuthExchangeInput` edge cases** (`oauth.go:806-831`). A denied-auth callback (`?error=access_denied`, no `code`) falls through to prefix-truncation and POSTs a garbage "code"; a bare opaque code containing `&`/`=` can be misparsed by `url.ParseQuery`. Surface the OAuth error instead of an opaque exchange failure. (Not security — state is always from the trusted session, never pasted input.)
- **No shape validation of plugin metadata** (`register.go`). Duplicate `SecretSlot`/`EnvVar` names or empty `EnvVar.Name` pass through to the dashboard/agent env with no diagnostic pointing at the plugin.
- **Asymmetric capability gating** (`register.go`). Build output setting `http_inject` is rejected unless the manifest declared it, but Build-declared `oauth` is accepted with no manifest flag — so two instances of one type can diverge into OAuth-capable vs not. Gate consistently or document as intentionally instance-scoped.
- **`redactions map[*http.Request][]string` has no TTL/cap** (`adapter.go`). Correct on the current path (consume runs unconditionally), but any future caller that forgets the paired consume leaks one entry per request for the credential instance's lifetime.
- **Docgen**: confirm any new documented HCL fields were regenerated via `go run ./internal/tools/docgen` — CI's test job fails otherwise.

## Part 2 — Future improvements (follow-up)

The injection abstraction is narrower than the rest of the SDK. Fine for the two shipped POCs, but a plausible third credential type hits a seam. Worth a tracked follow-up rather than blocking this PR:

- **Expose the request body to injection.** `HTTPInjectRequest.BodyPrefix`/`BodyTruncated` are declared but unwired, so body-signing schemes (AWS SigV4, HMAC-over-payload, webhook signers) can't be external plugins. Note the asymmetry: the gateway's internal `HTTPRequestSigner` (whole-request, used by built-in SigV4) was *not* exposed to plugins, so the external surface is strictly weaker than the internal one. Consider exposing body access / the signer interface behind a capability flag.
- **Make OAuth config open-ended.** `OAuthConfig` enumerates a fixed set of knobs; a provider needing `audience`, RFC 8707 `resource`, or a JWT-bearer/assertion grant can't express it without a core change, and each new `Flow` value still needs gateway-side code. An `Extra map[string]string` for unmodeled grant params would close most of this.
- **Nested schema blocks.** `Schema` is flat go-cty type strings; it's telling that `OptionalScopeGroup` is returned as Build metadata rather than declared in schema. Structured/repeated config will keep needing escape hatches until the schema supports nesting.

Net: the endpoint/protocol SDK is sufficiently generic; the HTTP-header credential-injection hook is deliberately narrow. Land once Part 1 is handled, and track Part 2 so the next provider integration doesn't discover the injection ceiling the hard way.

dhruvkelawala:
Addressed the review feedback in two commits:

- `b4b6ccc fix(extplugin): harden external credential injection`
  - Added coverage for derived-secret redaction: the HTTPS integration test now injects a derived header value and asserts it is redacted from audit samples.
  - Documented the `HTTPInjectResponse.Redactions` contract in the SDK/docs, including the derived-token/JWT/HMAC case.
  - Made the nil/unavailable external credential client path return an inject error instead of silently no-oping. The gateway still intentionally follows the existing fail-open behavior, but now logs that it is forwarding without injection.
  - Wrapped plugin callback dispatch with panic recovery for Build, InjectHTTP, endpoint HandleConn, and tunnel callbacks so one plugin panic returns an RPC error instead of taking down the whole subprocess.
  - Redacts longer derived strings before raw secrets so values like `derived-<secret>` do not leave an unredacted prefix in audit output.

- `e1dfd91 fix(extplugin): address review nits`
  - Validates plugin-provided HTTP header mutation names and values before applying them.
  - Documents that auth-style injection should use `HeaderSet`; `HeaderAdd` appends and can leave the placeholder header in place.
  - Tightens OAuth pasted-input parsing: denied callbacks now surface the OAuth error, while opaque bare codes containing `&` / `=` are preserved.
  - Adds plugin credential metadata shape validation for duplicate secret slot/env var names and empty env var names.
  - Documents OAuth metadata as intentionally Build-time/instance-scoped.
  - Caps the per-credential external HTTP redaction map as a guardrail for future callers that might forget to consume entries.
  - Ran `go run ./internal/tools/docgen`; there was no generated diff.

Verification run locally:

- `XDG_RUNTIME_DIR=/proc go test -count=1 ./...`
- `PATH="$(go env GOPATH)/bin:$PATH" make lint`

Current PR checks: `lint` and `test` pass. `preview` is still failing before build/deploy at `actions/checkout` because fork PRs do not receive the required demo deploy token (`Input required and not supplied: token`).

PR #657, opened 2026-06-10
onboarding: surface profile assignment at approval time

### LINKED ISSUE #655: onboarding: profile assignment is invisible at approval time — no picker on the approval page, no device row until first traffic
## Summary

A device's profile is decided at approval time, but the operator never sees or controls it:

- The dashboard approval page has no profile picker. `OnboardPage.tsx` posts `/api/onboard/approve?code=...` only, even though `apiOnboardApprove` already accepts `?profile=` (`cmd/clawpatrol/onboard.go:731-761`).
- `clawpatrol join --profile` (#85) exists and is honored — it's stashed as a suggestion at `/start` and passed explicitly on the `--login` auto-approve POST — but the approval page doesn't display the suggestion, so the approving operator can't see where the device is about to land, let alone override it.
- The dashboard's "add device" dialog (`AddDeviceModal.tsx`) renders a bare copy-pastable `clawpatrol join <url>` — it never suggests `--login` or `--profile`, so the flags that would make the assignment correct up front go unused.
- A freshly approved device doesn't appear in the dashboard devices list until it has produced traffic, so the post-hoc workaround (switch the profile from the devices list) requires generating dummy traffic first.

When nothing is specified anywhere, the device falls to `defaultProfileName` (the profile literally named `default`, else the first in source order).

## Observed (live join, 2026-06-10)

Joining a fresh exe.dev machine (`mandolin-kraken`) to the deno gateway:

1. `clawpatrol join http://clawpatrol-gateway:8080` → approve the user_code on the dashboard (from a phone via the QR code). No profile choice offered or shown; the device silently lands in the fallback profile.
2. The new device is invisible on the dashboard until it sends traffic. The operator had to run `clawpatrol run -- curl google.com` purely to make the device row appear, refresh, and then switch its profile manually.

## Why it matters

Profiles are the scoping unit: the discovery manifest (#645) is profile-scoped and VIP dispatch drops connections whose profile doesn't grant the endpoint (`handleVIPConn` "no endpoint matches profile", #652). A device that lands in the wrong profile works for passthrough traffic but fails on exactly the endpoints the gateway exists for, and the failure (silent drop) reads as a network bug, not a misassignment.

## Desired behavior

- **Approval page: profile picker.** Populated from the policy's profiles, preset to the `--profile` suggestion when the client passed one, otherwise a reasonable default (if the policy has exactly one profile, that one; else the `default`/first-in-source-order fallback). Selection is sent as `?profile=`. The "approved" confirmation should state the resulting assignment.
- **Same default for auto-approved joins.** `clawpatrol join --login` (auto-approve, no dashboard step) with no `--profile` should land on the same reasonable default as the picker preset.
- **"Add device" dialog: smarter copy-pastable command.** The suggested command should include `--login` when appropriate — especially when the current dashboard session is a tailnet identity on the operators allowlist (so the join will auto-approve) — and should suggest a `--profile` flag. When the right profile can't be inferred from dashboard context, the dialog shows a profile picker and selecting one updates the copy-pastable command in place.
- **Devices list shows just-approved devices.** A device must appear in the dashboard immediately after approval — before it connects or produces any traffic — so its profile can be inspected and changed right away.

## Related

- #85 — added `join --profile` (the CLI side of this).
- #645 — profile-scoped discovery manifest (what made the assignment visible).
- #652 — device-identity/profile collapses; failure mode 3 is what a wrong-profile device experiences on VIP endpoints.
- #604 — duplicate device names; same onboarding surface.

piscisaureus:
Fixes #655, plus a ground-up cleanup of the `clawpatrol join` output that surfaced while testing this end to end on a real client.

## Onboarding: profile assignment (the #655 work)

* `/api/onboard/lookup` returns the policy's profile names plus a `suggested_profile` (the CLI's `join --profile` value when it names a real profile, else the same default `apiOnboardApprove` would pick), so the approval page renders an accurate picker.
* Approval page shows a profile picker preset to the suggestion (static text when there's exactly one profile); the choice is sent as `?profile=`.
* `apiOnboardApprove` validates the profile: an explicit query param naming a nonexistent profile is a 400 (a typo'd `--login` auto-approve then falls back to dashboard approval instead of landing in a ghost profile); a bogus stashed suggestion degrades to the default. Zero-profile policies keep the old accept-anything behavior. Responses carry the resolved profile.
* "Add device" dialog builds a smarter copy-pastable command: `--login` when the dashboard session is a tailnet operator, a `--profile` picker (shown when there's more than one profile, preset to the gateway default), and a Mode selector — per-process `clawpatrol run` (default) or `--whole-machine` ("route all of this machine's traffic through clawpatrol").
* Just-approved devices appear in the devices list immediately, before any traffic: tsnet daemon joins seed the agents registry under a `tsnet-<host>` placeholder at approve time (mem-only; `upsertLocked` refuses placeholder rows), and the register promotion carries profile/owner/hostname across to the real tailnet IP. The claim path seeds on claim.

## `--login` auto-approve

* Fixed the auto-approve 404: it was POSTing to the public Funnel join URL (whose allowlist excludes `/api/onboard/approve`). It now targets the dashboard origin from `verify_url` (the operator-gated tailnet `:8080` API), where the bootstrap node's whois identity is actually checked.
* Time-boxed that POST (15s) and added a clear fallback — a wrong-tailnet login no longer hangs the join forever; it falls through to dashboard approval with a hint.
* `awaitTailnetAuth` fails fast on `NeedsMachineAuth` ("tailnet requires admin approval for this device") and reports the last backend state on timeout, instead of waiting out the 10-minute deadline silently.
* The ephemeral `--login` bootstrap node (`clawpatrol-bootstrap-*`) no longer shows up as a managed device — it's dropped on whois resolution and filtered from the devices list.

## `clawpatrol join` output, reworked

Every step is now a live `*` pending block that, on a TTY, collapses to a single `✓`/`!` line when it completes — no spinners. A finished join is a clean column:

```
✓ logged in to the tailnet
✓ auto-approved as tailnet operator
✓ CA installed in system trust
✓ joined gateway

Installed! Try: clawpatrol run claude
```

* `beginStep(headline, detail, erasable)` prints the pending block and returns a finish func that erases it (cursor-up + clear) and reprints one status line; off a TTY it appends instead, so piped logs stay clean. `linkDetail(url, qr)` builds the shared URL+QR detail used by the login and approve steps.
* Login and approve show the URL with an indented QR beneath it (scan from a phone — the joining box is usually headless). CA trust install runs behind a step (erases only when sudo is non-interactive). The CA fingerprint check (WireGuard mode) explains *why* it matters.
* Consistent wording: `✓ joined gateway` across all paths, `clawpatrol run claude` everywhere.

## Review / correctness

Ran a multi-agent review over the branch; fixed the real findings: a data race on `AgentRegistry.lc` (now triggered at boot by `Seed`→`fillIdentity`), the approver being dropped at register promotion, an unsynchronized `s.profile`/`s.wholeMachine` read, and a placeholder ghost-row race (seed the agent before minting the token register authenticates with).

New tests cover the lookup profile/suggestion shapes, approve validation + resolution priority, claim-time seeding, the placeholder→register promotion (profile/owner/hostname carry-over + cleanup), and `approveBaseFromVerifyURL`.

A walkthrough of the expected join output across OS / transport / `--login` / approval / `--whole-machine` / sudo is in [this comment](https://github.com/denoland/clawpatrol/pull/657#issuecomment-4674817018).

## Known limitations (pre-existing, not addressed here)

* The tsnet placeholder's profile/owner are in-memory only, so a gateway restart in the approve→register gap loses them.
* An abandoned approval leaves a "pending" device row and a non-expiring peer api-token with no GC.
* The default-profile rule lives in three places (Go `defaultProfileName`, the lookup suggestion, and the dashboard); a gateway-served default would unify them.

### DIFF
--- a/cmd/clawpatrol/agents.go
+++ b/cmd/clawpatrol/agents.go
@@ -195,6 +195,14 @@ func (r *AgentRegistry) Seed(ip string) {
 		}
 	}
 	r.agents[ip] = a
+	// Kick off the whois fill the same way trackUA does on first sight.
+	// Without this, pre-seeding the entry here means the agent-miss
+	// branch in trackUA never fires for this IP, so OS / tailnet login
+	// would never populate. Skip placeholder IDs (tsnet-<host>) — they
+	// aren't real addresses; promotion re-seeds under the tailnet IP.
+	if !strings.HasPrefix(ip, tsnetPlaceholderPrefix) {
+		go r.fillIdentity(ip)
+	}
 }
 
 func (r *AgentRegistry) track(remoteAddr, host string, in, out int64) {
--- a/cmd/clawpatrol/agents.go
+++ b/cmd/clawpatrol/agents.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/netip"
+	"slices"
 	"sort"
 	"strings"
 	"sync"
@@ -299,6 +300,14 @@ func (r *AgentRegistry) fillIdentity(ip string) {
 		return
 	}
 	if who.Node != nil {
+		// An ephemeral `--login` join-bootstrap node is not a managed
+		// device — its API calls during onboarding get tracked like
+		// any tailnet peer, but it's discarded the instant join
+		// completes. Drop it once whois reveals what it is.
+		if strings.HasPrefix(who.Node.Hostinfo.Hostname(), bootstrapHostnamePrefix) {
+			delete(r.agents, ip)
+			return
+		}
 		// Onboard-supplied hostname (via /api/onboard/claim or
 		// /api/peer/tsnet/register) takes priority. The whois response
 		// reflects whatever the tsnet node registered with (the daemon's
@@ -945,6 +954,12 @@ func (w *webMux) agentsList() []*Agent {
 	if w.g.agents != nil {
 		snap = w.g.agents.snapshot()
 	}
+	// Hide ephemeral `--login` join-bootstrap nodes — fillIdentity drops
+	// them once whois resolves, but filter here too so they never flash
+	// into the list in the window before that lands.
+	snap = slices.DeleteFunc(snap, func(a *Agent) bool {
+		return strings.HasPrefix(a.Hostname, bootstrapHostnamePrefix)
+	})
 	// External IPs: the underlay v4/v6 each WG peer is dialing in from.
 	// Show these in place of the server-side /32 (routing artefact).
 	// Live endpoint observed via wg-go IpcGet — persist into the devices
--- a/cmd/clawpatrol/agents.go
+++ b/cmd/clawpatrol/agents.go
@@ -289,7 +289,16 @@ func (r *AgentRegistry) fillIdentity(ip string) {
 	addrPort := netip.AddrPortFrom(addr, 0)
 	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
-	who, err := r.lc.WhoIs(ctx, addrPort.String())
+	// Snapshot the LocalClient under the lock — SetLocalClient swaps it
+	// in once tsnet is up, and Seed now spawns fillIdentity at boot
+	// (via seedAgentsFromDevices) which can race that write.
+	r.mu.Lock()
+	lc := r.lc
+	r.mu.Unlock()
+	if lc == nil {
+		return
+	}
+	who, err := lc.WhoIs(ctx, addrPort.String())
 	if err != nil || who == nil {
 		return
 	}
--- a/cmd/clawpatrol/onboard.go
+++ b/cmd/clawpatrol/onboard.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"net/netip"
 	"net/url"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -185,6 +186,23 @@ func (r *onboardRegistry) assignProfileMemOnly(ip, profile string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	r.profileByIP[ip] = profile
+}
+
+// seedPlaceholder records hostname/owner for a tsnet placeholder ID,
+// in memory only (upsertLocked refuses placeholder IDs), so the
+// agents-registry Seed at approve time renders a recognizable device
+// row before the daemon's first register call promotes the
+// placeholder to a real devices row. ForgetIP at promotion clears
+// these entries.
+func (r *onboardRegistry) seedPlaceholder(id, hostname, owner string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if hostname != "" {
+		r.hostnameByIP[id] = hostname
+	}
+	if owner != "" {
+		r.ownerByIP[id] = owner
+	}
 }
 
 // AgentIPFor returns the device IP that traffic from ip should be
@@ -716,11 +734,25 @@ func (w *webMux) apiOnboardLookup(rw http.ResponseWriter, r *http.Request) {
 		http.Error(rw, "unknown or expired code", 404)
 		return
 	}
+	// Profile names + the preselection for the approval page's picker.
+	// The suggestion is the CLI's `join --profile` value when it names
+	// a real profile, else the same fallback apiOnboardApprove uses —
+	// so what the picker shows is exactly what an unmodified approve
+	// would assign. After approval s.profile holds the final choice,
+	// so the page can render the assignment.
+	policy := w.g.cfg.Load().Policy
+	profiles := orderedProfileNames(policy)
+	suggested := s.profile
+	if !slices.Contains(profiles, suggested) {
+		suggested = defaultProfileName(policy)
+	}
 	writeJSON(rw, map[string]any{
-		"user_code":      s.userCode,
-		"approved":       s.approved,
-		"created_at":     s.created.Unix(),
-		"ca_fingerprint": w.caFingerprint(),
+		"user_code":         s.userCode,
+		"approved":          s.approved,
+		"created_at":        s.created.Unix(),
+		"ca_fingerprint":    w.caFingerprint(),
+		"profiles":          profiles,
+		"suggested_profile": suggested,
 	})
 }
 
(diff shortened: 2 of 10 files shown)


piscisaureus:
Ran a multi-agent review against this branch. Fixed the confirmed correctness issues in 4d353cac:

* **Whois identity fill was suppressed for every seeded device.** `trackUA` only kicks off `fillIdentity` in its agent-miss branch, so pre-seeding the agents entry at approve/claim/register meant OS and tailnet login never populated on first traffic. `AgentRegistry.Seed` now launches `fillIdentity` for real (non-placeholder) IPs.
* **Owner was dropped at the tsnet register promotion.** `ForgetIP` cleared `ownerByIP[placeholder]` before promotion, and only profile/hostname were re-applied — so a device's approver vanished the moment it first connected. Now owner (and hostname, if the daemon didn't supply a fresher one) are captured before `ForgetIP` and carried across via a new `SetOwner`. Test extended to assert it.
* **`apiOnboardLookup` read `s.profile` unsynchronized and could misreport the assignment.** It now reads session fields under `onboard.mu` and, once approved, reports the stored profile verbatim (so a profile removed from the policy post-approval doesn't get rewritten to the default in the UI).

Cleanups also applied: folded profile into `seedPlaceholder` (one critical section, dropped the single-use `assignProfileMemOnly`), extracted `resolveProfileFromList` shared by lookup+approve (collapsed the approve ladder and its dead trailing branch), and made placeholder rows non-navigable in the devices table with the `tsnet-` prefix check hoisted to one const.

Known limitations left as-is (pre-existing, not introduced here; worth a separate issue): the placeholder's profile/owner are in-memory only, so a gateway restart in the approve→register gap loses them; an abandoned approval leaves a "pending" row and a live api-token with no GC; and deleting a placeholder device doesn't revoke its token. All race/unit tests and the dashboard build pass.

piscisaureus:
Live `--login` join testing surfaced more issues; fixed in b63c899e and redeployed:

* **Auto-approve 404 (the main bug).** The CLI POSTed `/api/onboard/approve` to the join URL, which for a funnel-enabled gateway is the public `:443` Funnel — whose allowlist (`funnelPublicHandler`) only permits start/poll/claim, so approve 404'd. It now derives the approve origin from the start response's `verify_url` (the operator-gated dashboard/API — tailnet `:8080` in tsnet mode), the same origin the dashboard Approve button posts to and where the bootstrap node's tailnet whois identity is actually checked. Covered by a new `TestApproveBaseFromVerifyURL`.
* **Ephemeral bootstrap node showed up as a managed device.** The `--login` bootstrap tsnet node's onboarding API calls get tracked like any tailnet peer. Agents whose whois hostname carries the `clawpatrol-bootstrap-` prefix (now a shared const) are dropped on identity resolution and filtered from the devices list.
* **QR for the tailnet login URL.** The joining box is usually headless, so `clawpatrol join` now prints a QR for the `login.tailscale.com` URL too — scan from a phone instead of waiting on a browser that can't open.
* **Add-device dialog.** The copy button moved into the label row so it no longer overlaps the (long) join command, and a **Mode** selector was added — per-process `clawpatrol run` stays the default, with an explicit `--whole-machine` option that appends the flag.

All Go + dashboard checks pass. Deployed to the gateway; `bert@deno.com` is in the operators allowlist, so re-running the `--login` join should now auto-approve cleanly.

piscisaureus:
## Expected `clawpatrol join` output across combinations

Derived from the join flow on this branch (`setup.go` / `setup_tailnet_login.go`), updated for the live-step output.

**How the output works now:** every step prints as a `*` pending block (its headline plus any URL / QR / fingerprint detail) and, on a TTY, collapses to a single `✓` (done) or `!` (needs-attention) line once it completes — no spinners. Off a TTY (piped/non-interactive) the status line is appended instead. So a finished join is a clean column of `✓` lines; the samples below show that collapsed end-state, with the pending block shown where it's distinctive.

Six axes: **OS** (linux/macOS), **transport** (Tailscale/WireGuard — set by the *gateway*, the client adapts), **`--login`**, **approval** (auto/manual), **`--whole-machine`**, **sudo**.

### How the axes interact (and collapse)

- **`--login`** stands up a throwaway tailnet node to reach the gateway and is what *enables* auto-approval; it only applies to a **Tailscale-mode** gateway. With it you get the login step at the top. Even without it, a tailnet-shaped gateway URL that's unreachable from this network auto-falls-back to the same bootstrap (`gateway <url> unreachable from this network; falling back to tailnet login.`).
- **Approval:** **auto** requires `--login` (or the fallback) **and** the joining tailnet identity being a dashboard operator; otherwise it degrades to **manual**. **WireGuard is always manual** → *WireGuard + auto / + `--login` = N/A*.
- **CA fingerprint** is shown only in **WireGuard** manual (CA fetched over plain HTTP up front → verify by eye). Tailscale fetches the CA over the tailnet, so no fingerprint.
- **QR on the approve step** appears when the verify URL is tailnet-only (Tailscale → tailnet `:8080`); WireGuard's is the public dashboard → no QR.
- **`--whole-machine`** adds a `✓ shell rc updated` line, changes the final `Installed!` line, and brings up host-wide routing.
- **sudo:** the CA install needs it → `✓ CA installed in system trust`; unavailable (or `--no-trust`) → `! CA not trusted yet — install manually: …`. Linux `--whole-machine` also needs sudo for system Tailscale / `wg-quick` or it fails outright.
- **OS** changes only the manual-trust hint command and the `--whole-machine` final line.

### Validity grid

| Transport | `--login` | Approval | Valid? |
|-----------|-----------|----------|--------|
| Tailscale | yes | auto (operator) | ✅ Sample **A** |
| Tailscale | yes | manual (non-operator) | ✅ **B**, with login step |
| Tailscale | no | manual | ✅ **B**, no login step (or auto-fallback → login step) |
| WireGuard | — | manual | ✅ Sample **C** |
| WireGuard | yes / auto | — | ❌ N/A |

Each ✅ row then × {linux, macOS} × {`--whole-machine` on/off} × {sudo on/off} — handled by the deltas below.

---

### Sample A — Tailscale, `--login`, auto-approve (per-process)

While running, each step shows then collapses, e.g.:
```
* Log in to the tailnet to reach the gateway
  https://login.tailscale.com/a/94b1a5d0166ab

  ▄▄▄▄▄▄▄ ▄  ▄▄▄▄▄▄▄
  █ ▄▄▄ █ ▀█ █ ▄▄▄ █      (QR, indented under the URL)
  █▀▀▀▀▀▀▀ ▀ ▀▀▀▀▀▀▀
```
Final (every step collapsed):
```
✓ logged in to the tailnet
✓ auto-approved as tailnet operator
✓ CA installed in system trust
✓ joined gateway

Installed! Try: clawpatrol run claude
```

### Sample B — Tailscale, manual approval (per-process)

Drop `--login` (or use it but not as an operator; the login step then appears first and collapses to `✓ logged in to the tailnet`). The approve step stays up until the dashboard approve click lands, then collapses to `✓ approved`:
```
* Approve this device on the dashboard — code NDVL-4343
  http://100.79.206.14:8080/#/onboard/NDVL-4343

  ▄▄▄▄▄▄▄ ▄  ▄▄▄▄▄▄▄
  █ ▄▄▄ █ ▀█ █ ▄▄▄ █      (QR — scan from a tailnet device)
  █▀▀▀▀▀▀▀ ▀ ▀▀▀▀▀▀▀
```
Final:
```
✓ approved
✓ CA installed in system trust
✓ joined gateway

Installed! Try: clawpatrol run claude
```

### Sample C — WireGuard, manual approval (per-process)

`--login` / auto don't apply. The approve step carries the CA fingerprint (no QR — the dashboard URL is public):
```
* Approve this device on the dashboard — code NDVL-4343
  https://clawpatrol.example.com/#/onboard/NDVL-4343

  before approving, confirm this CA fingerprint matches the dashboard:
  SHA256:ab:cd:ef:12:…
```
Final:
```
✓ approved
✓ CA installed in system trust
✓ joined gateway

Installed! Try: clawpatrol run claude
```

---

### Delta: `--whole-machine`

Adds `✓ shell rc updated (eval "$(clawpatrol env)")` and replaces the final line:

| OS | Transport | Final line | Extra |
|----|-----------|-----------|-------|
| Linux | WireGuard | `Installed! All host traffic routes via the gateway (clawpatrol0).` | runs `wg-quick up` |
| Linux | Tailscale | `Installed! Try: clawpatrol run claude` | installs system Tailscale, `tailscale up`, sets `--exit-node` (all **need sudo**) |
| macOS | either | `Installed! All host traffic routes via the system extension.` | NE extension owns routing |

### Delta: `--login` (and auto-fallback)

Prepends the login step — `* Log in to the tailnet to reach the gateway` + URL + QR, collapsing to `✓ logged in to the tailnet`. With operator identity it continues as **A**; otherwise as **B**. If the tailnet gates new devices on admin approval the node never reaches Running, and the step now fails fast: `! tailnet requires admin approval for this device`. Auto-fallback (no `--login`, unreachable tailnet URL) is the same but prefixed with `gateway <url> unreachable from this network; falling back to tailnet login.`

### Delta: sudo unavailable (any combo)

The CA step's `* Installing CA in system trust` resolves to a `!` line instead of `✓`:
```
! CA not trusted yet — install manually: sudo cp /home/you/.config/clawpatrol/ca.crt /usr/local/share/ca-certificates/clawpatrol.crt && sudo update-ca-certificates
```
- macOS hint: `! CA not trusted yet — install manually: sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /Users/you/.config/clawpatrol/ca.crt`
- `--no-trust` shows `! CA fetched but not trusted (--no-trust) — install manually: …` (no install attempt).
- Linux `--whole-machine` additionally fails outright without sudo.

*(Interactive password sudo: the `* Installing CA…` line is kept rather than erased, since the sudo prompt prints inside the step — the `✓`/`!` line is appended below it.)*

PR #665, opened 2026-06-11
run: apply Claude Code OAuth shim on the passwordless-sudo path (cl-ulck)

arnauorriols:
## Problem

`CLAWPATROL_CLAUDE_OAUTH_SHIM=1` silently broke credential injection on the **passwordless-sudo run path** (now the default after #663): enabling it left Claude Code unable to use the injected OAuth credential — it behaved as API-key/bearer auth, so OAuth-only features like `/remote-control` stayed gated.

## Root cause

`runViaSudo` called `installClaudeCodeOAuthShim` **before** the gateway env-pushdown had injected `ANTHROPIC_AUTH_TOKEN`. The pushdown is fetched and merged *root-side* in `runRunPrivileged` (`buildPrivilegedEnv`), long after the unprivileged parent captured its environment. So the shim saw no bearer and returned an early no-op — never writing `.credentials.json`, never setting `CLAUDE_CONFIG_DIR`, never stripping the token. The pushdown then injected `ANTHROPIC_AUTH_TOKEN` into the child anyway, pinning Claude Code to bearer mode (precedence #2) instead of subscription OAuth (#6).

The in-code comment at `run_sudo_linux.go` ("capture … after the OAuth shim, so `CLAUDE_CONFIG_DIR` is included") documented the intended-but-broken behavior.

The Linux userns path and both darwin paths already apply the pushdown **before** the shim, so they were unaffected.

## Fix

- Extract the shim core into `planClaudeCodeOAuthShim`, shared by the process-env path (`installClaudeCodeOAuthShim`) and a new sudo slice-env path.
- `applyClaudeCodeOAuthShimSudo` runs in `runRunPrivileged` against the fully-built child env (where the pushdown has injected the bearer), derives the managed config dir from the **child's** `HOME`, and `chown`s the synthesized `.credentials.json` to the target uid/gid so the dropped-to-user `claude` can read it.
- Remove the dead call in `runViaSudo`; fix the stale comment.
- Document the per-run-path ordering requirement in `doc/claude-code-oauth.md`.

## Tests

New `run_sudo_linux_test.go` covers the sudo env transform: opt-in (token stripped, `CLAUDE_CONFIG_DIR` set, creds written 0600 with `user:sessions:claude_code`), not-opted-in (env untouched), and operator-set `CLAUDE_CONFIG_DIR`. Existing `installClaudeCodeOAuthShim` tests unchanged and green. `gofmt`/`go vet` clean.

Note: `TestEnvPushdownVars{ServerDriven,ErrorReturnsCAOnly}` fail on the base commit too (env-sensitive httptest fetch, untouched by this PR) — filed separately.

### DIFF
--- a/cmd/clawpatrol/integrations.go
+++ b/cmd/clawpatrol/integrations.go
@@ -377,31 +377,81 @@ func applyEnvPushdownVars(vars []pushdownEnvVar) {
 // `.credentials.json` — that login wins once we drop the bearer, so we
 // leave it untouched.
 func installClaudeCodeOAuthShim(cmd []string) {
-	if len(cmd) == 0 || filepath.Base(cmd[0]) != "claude" {
+	res := planClaudeCodeOAuthShim(cmd, os.Getenv, defaultClawpatrolDir(), -1, -1)
+	if res.warn {
+		warnClaudeCodeRemoteControlDisabled()
 		return
 	}
-	bearer := os.Getenv("ANTHROPIC_AUTH_TOKEN")
-	if bearer == "" {
+	if !res.applied {
 		return
+	}
+	// Redirect Claude Code's credential read onto our synthesized file
+	// (only when we own the dir; an operator-set one is already exported).
+	// Unsetting ANTHROPIC_AUTH_TOKEN drops Claude Code out of precedence #2
+	// so the synthesized credentials.json (precedence #6) wins.
+	if res.configDir != "" {
+		_ = os.Setenv("CLAUDE_CONFIG_DIR", res.configDir)
+	}
+	if res.unsetAuthToken {
+		_ = os.Unsetenv("ANTHROPIC_AUTH_TOKEN")
+	}
+}
+
+// claudeShimResult is the decision + side effects of one shim
+// evaluation. The file write (credentials.json) has already happened
+// when applied is true; the caller is responsible only for reflecting
+// the env changes (configDir / unsetAuthToken) onto whatever env
+// representation it uses — os.Setenv for the live process, or a
+// []string slice for the privileged sudo child.
+type claudeShimResult struct {
+	applied        bool   // shim fired: credentials.json was (re)written
+	warn           bool   // not opted in — caller should print the notice
+	configDir      string // non-empty → set CLAUDE_CONFIG_DIR to this
+	unsetAuthToken bool   // drop ANTHROPIC_AUTH_TOKEN from the child env
+}
+
+// planClaudeCodeOAuthShim is the shared core of the `clawpatrol run
+// claude` OAuth shim. It reads env via get, decides whether the shim
+// applies, and (when it does) writes the synthesized credentials.json —
+// chowning it to chownUID/chownGID when chownUID >= 0, so the privileged
+// sudo path can materialize a file the dropped-to-user child can read.
+// It returns the env mutations for the caller to apply; it deliberately
+// does NOT touch process env, so it works equally for os.Getenv and for
+// a captured []string env.
+//
+// clawDir is the `.clawpatrol` directory to carve the managed config dir
+// out of (the user's, not root's, on the sudo path).
+func planClaudeCodeOAuthShim(cmd []string, get func(string) string, clawDir string, chownUID, chownGID int) claudeShimResult {
+	if len(cmd) == 0 || filepath.Base(cmd[0]) != "claude" {
+		return claudeShimResult{}
+	}
+	bearer := get("ANTHROPIC_AUTH_TOKEN")
+	if bearer == "" {
+		return claudeShimResult{}
 	}
 	// Opt-in only. Writing credentials + repointing CLAUDE_CONFIG_DIR can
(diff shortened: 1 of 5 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR description. It's a self-contained summary of the change already on `polecat/rust-mq99k6ph` (problem → root cause → fix → tests), not a change request or question — so there's nothing to action in code.

Confirmed the branch matches the described scope: `integrations.go` (extracted `planClaudeCodeOAuthShim`), `run_sudo_linux.go` (sudo slice-env apply + dead-call/comment cleanup), new `run_sudo_linux_test.go`, and `doc/claude-code-oauth.md`. lint + preview checks are green; `test` is still running. The pre-existing `TestEnvPushdownVars{ServerDriven,ErrorReturnsCAOnly}` failures are unrelated to this PR (noted in the body, filed separately).

No further action from this watcher cycle.

arnauorriols:
I am not running clawpatrol run in a passwordless sudo user, and I am experiencing the bug

arnauorriols:
[bot-reply]
Good catch — you're right, and the original PR missed your path. The fix only covered the **passwordless-sudo** path; it left a third Linux run path untouched.

**Root cause of your case:** on a `--whole-machine` device, `clawpatrol run` takes neither the userns nor the sudo path — it goes through `runWholeMachineDirect` (added in #662), which exec's the command directly with no sandbox. That function injected the gateway env-pushdown (`ANTHROPIC_AUTH_TOKEN`) but **never called `installClaudeCodeOAuthShim`**. So with `CLAWPATROL_CLAUDE_OAUTH_SHIM=1` the bearer was injected and left set: Claude Code stayed in API-key mode and `/remote-control` (and other OAuth-only features) stayed gated. Same bug, different path — which is why "not passwordless sudo" still reproduced it.

Pushed f80b3cbd:
- Call `installClaudeCodeOAuthShim(cmd)` in `runWholeMachineDirect`, right after the pushdown injects the token and before exec. The command runs as the invoking user (no privilege drop), so the process-env shim is sufficient — no `chown` dance like the sudo path needs.
- Document the whole-machine path in the per-run-path ordering section of `doc/claude-code-oauth.md` (now three paths, not two).

The shim is now wired on every run path: userns (`run_linux.go`), sudo (`run_sudo_linux.go`), whole-machine (`runWholeMachineDirect`), and both darwin paths. `gofmt`/`go vet`/build clean; shim tests green. (The two `TestEnvPushdownVars*` failures remain pre-existing and unrelated, as noted earlier.)

Can you confirm `clawpatrol run claude` now reaches OAuth mode on your whole-machine device with the shim opted in?

arnauorriols:
I am not joining with --whole-machine either

arnauorriols:
[bot-reply]
I've stopped guessing paths and audited every `clawpatrol run` path end-to-end. After f80b3cbd, all five apply the env-pushdown *then* the OAuth shim, in the correct order:

- Linux userns — `run_linux.go:164-165` (pushdown → shim), child inherits via `os.Environ()`
- Linux passwordless-sudo — fixed in 07a4b4c2
- Linux whole-machine — fixed in f80b3cbd
- Darwin normal — `run_darwin.go:105-106`
- Darwin tsnet — `run_tsnet_darwin.go:121-122`

So if you're on neither sudo nor whole-machine, you're on the **userns (Linux)** or **darwin** path — and both already strip `ANTHROPIC_AUTH_TOKEN` and write the synthesized `.credentials.json` before the wrapped command execs. I verified the daemon-mediated userns path too: the daemon forwards `ANTHROPIC_AUTH_TOKEN` in its cached pushdown (covered by `daemon_linux_test.go`), the parent applies it at :164, the shim drops it at :165, and `strippedRunEnv()` preserves `CLAUDE_CONFIG_DIR`. There's no remaining path where the shim is skipped after the pushdown.

That means your symptom is most likely **not** an ordering/missing-path bug — it's a firing-condition issue. The shim **silently no-ops (leaving the bearer set — exactly your symptom, no warning printed)** when the wrapped command's basename isn't literally `claude` (`integrations.go:425`: `filepath.Base(cmd[0]) != "claude"`). A wrapper script, `npx`, an absolute path to a differently-named shim, etc. would all bypass it.

Could you run this one diagnostic so I can pin the real gap instead of guessing again?

```
CLAWPATROL_CLAUDE_OAUTH_SHIM=1 clawpatrol run -- env | grep -E 'ANTHROPIC_AUTH_TOKEN|CLAUDE_CONFIG_DIR'
```

and tell me:
1. OS (Linux userns or macOS)?
2. The exact `clawpatrol run …` command you invoke (specifically, what's the first token after `run --`)?
3. Is `CLAWPATROL_CLAUDE_OAUTH_SHIM=1` exported in that same shell?

If `ANTHROPIC_AUTH_TOKEN` is still present in that output, the shim isn't firing — and (2) will tell us why. If it's absent but OAuth features are still gated, then the synthesized-credential mechanism is the suspect and I'll dig there. Happy to push the fix as soon as I know which it is.

PR #666, opened 2026-06-11
discovery: document HITL and expose approval-status polling on the internal API

arnauorriols:
Extends the profile-scoped discovery manifest (#645) to make agents HITL-aware.

## What changed
- **Flag HITL endpoints per profile.** An endpoint is flagged when at least one enabled rule routes matching requests through a human approver (`human_approver` or the built-in `dashboard`). Rules decided purely by an automated `llm_approver` are not flagged — they don't wait on a human. Surfaced as a per-endpoint `hitl` field and a top-level `gated_endpoints` list.
- **Explain HITL in the manifest.** A new `## Human-in-the-loop approval` section (markdown) and `hitl` object (JSON), rendered from one representation, explain that a matching rule parks the request pending human approval, that it may be parked indefinitely, and that the agent should poll for status rather than treat a slow request as a failure. The 202 → poll → retry flow is documented.
- **Expose the poll endpoint on the internal API.** `GET https://clawpatrol.internal/api/hitl/operations/{id}/status` now answers on the reserved internal host, resolving the caller from the connection-derived profile/principal (the same identity that parked the request) or a capability token. It mirrors the existing tailnet web surface and its state machine.

## Tests
- `hitl.hcl` golden fixture pairs all four cases (human / dashboard / llm / no-rule) in one profile.
- Asserts the human and dashboard endpoints are flagged, the llm and no-rule endpoints are not, and the llm approver's private credential never leaks into the manifest.
- The internal poll endpoint returns the parked state to the owning device, 404s for a foreign principal or wrong/unknown id, and resolves by capability token.
- `gofmt -l .` clean; full `cmd/clawpatrol` package green.


arnauorriols:
Do not mention the 202 responses, those are only for the async hitl. I want this to work for sync HITL too. make sure the polling endpoint works for these too

arnauorriols:
[bot-reply]
PR-body watcher reply — no code change.

The body is an accurate self-description of the diff, not a reviewer question or change request, so there's nothing to implement. I checked the body's claims against the diff and they hold:

- **Flag HITL endpoints per profile** — `endpointHasHITL`/`isHumanApprover` flag an endpoint only when an enabled rule's approve chain hits `human_approver` or the built-in `dashboard`; `llm_approver`-only rules are excluded. Surfaced as `DiscoveryEndpoint.HITL` and the top-level `gated_endpoints`, with `buildDiscoveryHITL` deriving the gated set from the per-endpoint flags so they can't disagree.
- **Explain HITL in the manifest** — single `DiscoveryHITL` representation rendered into both the JSON `hitl` object and the `## Human-in-the-loop approval` markdown section; explanation built from live routing constants (internal host, poll path, retry header) so docs can't drift from the served flow. 202 → poll → retry documented.
- **Expose the poll endpoint** — `serveInternalHITLStatus` answers `GET /api/hitl/operations/{id}/status` on `clawpatrol.internal`, resolving the caller from the connection-derived profile/principal (`hitlPeerPrincipalID(agentIPFor(c))`) or a capability `token`.
- **Tests** — `hitl.hcl` golden fixture covers human/dashboard/llm/no-rule; `discovery_hitl_status_test.go` covers owner-by-principal, foreign-principal 404, and token resolution.

If the intent was to request specific changes, drop a comment describing them and I'll pick it up next cycle.

arnauorriols:
Done. Dropped the 202-centric framing — the manifest now describes the default sync park (the gateway holds the connection until a human decides and returns the real response or a denial on it) and presents the approval-status poll as the universal status check that works for a parked request whether it is still held (sync) or handed back to follow later (async). The internal poll endpoint already serves a `sync_waiting` operation, not just an async-pending one; added a test that polls a sync-parked request and sees it pending with upstream not yet called, and regenerated the golden fixtures. `gofmt` clean, package green.

arnauorriols:
[bot-reply]
No new code change — this comment is a self-summary of work already pushed in `59df1c64` ("discovery: document HITL polling for sync and async parking"), it just missed the `[bot-reply]` prefix, so the watcher re-triggered on it. Confirmed the commit on `polecat/rust-mq9dayhi` matches the summary:

- Drops the 202-centric framing; leads with the synchronous park (held connection) as the default and presents the `operation_id` / `status_url` poll as the universal status check for any parked request (sync-held or async-handed-back).
- Internal poll handler + `DiscoveryHITL` doc note the endpoint serves a sync-waiting (held) operation, not only an async-pending one.
- Adds a poll test for a `sync_waiting` operation: the owning device polls it and sees it pending with upstream not yet called.
- Regenerates the discovery golden fixtures.

Nothing further to implement for this comment.

arnauorriols:
Now, let's make the sync hitl timeout optional (if not already).

arnauorriols:
[bot-reply]
Pushed d58ab020 making the sync HITL timeout optional:

- `sync_wait_timeout` was already `,optional` at the HCL level, but `ValidateHITLAsyncGrant` emitted a `sync_wait_timeout is required` error whenever `async_grant.enabled = true`. Dropped that requirement — it's now genuinely optional even with an async grant.
- Behavior when unset: the gateway parks the request synchronously for the full approval window (holds the connection until a human decides) and never hands back an early 202; the async retry-grant path stays dormant. This is already what the runtime does — `maybeStartAsyncHITLOperation` returns `false` when the sync wait is `<= 0`, so no code change was needed there.
- When `sync_wait_timeout` *is* set, it's still validated as a positive duration.
- Updated the field doc and flipped the old `RequiresSyncWaitTimeoutWhenEnabled` test to assert the omitted case now loads and reads back a zero hold budget. `gofmt` clean; `internal/config/...` and the `cmd/clawpatrol` HITL/Discovery tests green.

arnauorriols:
1. do not mention status_url, operation_id, or any other machinery of the async HITL. We are currently adding info about the sync HITL
2. Implement an endpoint to list all the parked actions of a device. This endpoint should be available at https://clawpatrol.internal/pending. 
3. Remove the clawpatrol.internal/api/hitl/... endpoint, this is not useful in sync hitl

arnauorriols:
[bot-reply]
Pushed 5f33367 addressing all three points:

1. **Dropped async machinery from the manifest.** The HITL explanation now describes synchronous parking only — the gateway holds your connection open until a human decides and answers on that same connection. All `status_url` / `operation_id` / 202 / retry-header / poll-by-id prose is gone (JSON `hitl.poll_path` → `hitl.pending_path`).
2. **Added `GET https://clawpatrol.internal/pending`.** Lists the device's parked actions (`sync_waiting` + `pending_approval`), scoped to the connection-derived profile/principal, each with redacted method/endpoint/target and no async-poll handles. Backed by a new `HITLOperationStore.ListParkedForPrincipal`.
3. **Removed `clawpatrol.internal/api/hitl/operations/{id}/status`** (the internal poll endpoint) and its helper. The pre-existing public tailnet poll surface is untouched.

Goldens regenerated; `cmd/clawpatrol` + `internal/config/...` tests, `go vet`, and `gofmt` all clean.

arnauorriols:
### `/pending` is empty for plain sync HITL — it should also list the in-memory sync operations

I tested this branch end-to-end against a local gateway (passthrough credential + `httpbin.org` endpoint + a no-condition `approve` rule), driving real parked requests through `clawpatrol run`. The manifest copy and the synchronous park/approve/relay all work great. But the new `/pending` endpoint returns an empty list for the **plain synchronous HITL** case — which is exactly the case the manifest tells the agent to use it for.

#### Repro
Rule using a sync-only approver (no async grant), e.g. `approve = [builtin.dashboard]` or a `human_approver` without `async_grant`:

- Request parks (connection held) ✅
- Dashboard `GET /api/hitl/pending` shows it ✅
- `hitl_operations` table has **0 rows**
- `GET https://clawpatrol.internal/pending` → `{"pending":[]}` ❌

Flip the same setup to `hitl_async_grants = true` + `async_grant { enabled = true }` + `sync_wait_timeout` and everything lines up — a `sync_waiting` row lands in `hitl_operations` and `/pending` lists the action. So `/pending` only works when async grants happen to be enabled.

| Config | parks? | dashboard shows? | `hitl_operations` | `/pending` |
|---|---|---|---|---|
| sync-only (`builtin.dashboard`) | yes | yes | **0 rows** | **`[]`** |
| async grants enabled | yes | yes | 1 row | lists it ✅ |

#### Root cause
`serveInternalPending` (discovery.go:125) reads only the DB via `ListParkedForPrincipal` (hitl_operation_store.go:203). But `hitl_operations` rows are created exclusively inside `startAsyncHITLOperation` (hitl_async_runtime.go:87), which early-returns unless the profile has `HITLAsyncGrants` **and** the approver has `async_grant.enabled` **and** `sync_wait_timeout > 0`. Pure-sync parks never touch the DB — they live only in the in-memory `HITLRegistry` pool that the dashboard's `g.hitl.List()` reads. So the manifest advertises `/pending` unconditionally, but it's empty on any sync-only deployment.

Worth noting this reproduces against the PR's own `testdata/discovery/hitl.hcl` shape (`human_approver`/`builtin.dashboard`, no async grants): that manifest advertises `/pending`, yet `/pending` would be empty when those requests park. `discovery_pending_test.go` doesn't catch it because it seeds rows with `store.Create` directly rather than parking a real request.

#### Suggestion: also list the in-memory sync entries
Have `serveInternalPending` merge the in-memory pool into its response, scoped to the caller. `HITLPending` (runtime.go:602) already carries `AgentIP`, `Host`, `Method`, `Path`, `Endpoint`, `CreatedAt`, `ExpiresAt` — enough to render the same redacted `pendingActionView`-style item.

A clean de-dup falls out of `buildPending` (approvers/util.go), which sets `OperationID: req.AsyncOperationID`:
- sync-only entries → `OperationID == ""` → **not** in the DB
- async entries → `OperationID != ""` → **already** in the DB (`sync_waiting` row)

So: return the DB rows from `ListParkedForPrincipal` **plus** the pool entries whose `OperationID == ""` and whose `AgentIP` matches `g.agentIPFor(c)` (the same agent address the principal is derived from). That covers the sync case, keeps the async case unchanged, and never double-lists. A small adapter renders `HITLPending` into the existing redacted view shape.

(Alternative if the intent was async-only: gate the manifest's `/pending` advertisement on the profile actually having async grants — but listing the sync pool seems closer to what the manifest promises.)

I'm happy to push a commit with the merge + a `discovery_pending_test.go` case that parks a real sync request (not a seeded row) if that's useful.

arnauorriols:
[bot-reply]
Pushed 0d71c807 addressing this comment — `/pending` now lists pure-sync HITL parks, not just async-grant operations.

- `serveInternalPending` (discovery.go) now unions the DB rows from `ListParkedForPrincipal` with the in-memory `HITLRegistry` pool, scoped to the caller via `AgentIP` (the same `g.agentIPFor(c)` the principal is derived from).
- De-dup is exactly as you suggested: pool entries with `OperationID == ""` are pure-sync parks not in the DB and get included; `OperationID != ""` entries are the async case already represented by a `sync_waiting`/`pending_approval` row, so they're skipped — no double-listing, async path unchanged.
- New `pendingPoolView` adapter renders `HITLPending` into the same redacted `pendingActionView` shape. It drops the query string to match the DB path's redaction (`RedactedPath` is the path component only), so query secrets never surface; scheme is `https` to match the operation store's hardcoded value.
- Merged list is sorted newest-first across both sources.
- Added a `discovery_pending_test.go` case (`TestInternalPendingListsSyncPool`) that parks a real sync request through the pool (no seeded DB row), asserts it's listed with the query dropped, that an async (`OperationID != ""`) pool entry is not double-listed, and that another device's park stays invisible.

`go test ./cmd/clawpatrol/` and `gofmt`/`go vet` are green.

arnauorriols:
### `/pending` reports an `approval_expires_at` that the gateway never enforces (pure-sync parks)

Following up from testing the sync-park fix (thanks for that — works well). I hit a misleading field: for a **pure-sync** park (a `builtin.dashboard` / dashboard-only rule, no async grant), `/pending` returns an `approval_expires_at` exactly 30 minutes after `parked_at` — but nothing actually expires the request at that time. It parks indefinitely until a human decides or the agent disconnects.

```jsonc
// GET https://clawpatrol.internal/pending  — rule: approve = [builtin.dashboard]
{
  "endpoint": "httpbin.org",
  "method": "GET",
  "url": "https://httpbin.org/ip",
  "parked_at":           "2026-06-12T09:18:44Z",
  "approval_expires_at": "2026-06-12T09:48:44Z"   // +30m — but no timeout is enforced
}
```

An agent reading `/pending` would reasonably treat `approval_expires_at` as a deadline — "this gets auto-denied at 09:48, so give up / retry after that." That's wrong here, and it contradicts the manifest copy this PR ships, which tells the agent a parked request "may stay parked indefinitely."

#### Why it's not real
- `pendingPoolView` (discovery.go) emits `approval_expires_at` whenever `p.ExpiresAt` is non-zero.
- For a sync `builtin.dashboard` park, `buildPending` never sets `ExpiresAt`, so `HITLRegistry.Add` fills it with a hardcoded **display default** of `CreatedAt + 30m` (web.go:2991).
- Nothing enforces it: `runApproveChain` runs the dashboard stage on the bare request context with no added deadline, `DashboardApprover.Approve` only waits on `ctx.Done()`, and `pruneTerminalLocked` sweeps *terminal* entries only — never pending ones by `ExpiresAt`. So the 30m comes and goes with the request still held.

So `/pending` is surfacing a cosmetic dashboard-UI value as if it were a contract.

#### Suggested fix
Don't advertise an expiry the gateway won't honor. Cleanest at the root: stop fabricating the 30m in `HITLRegistry.Add` — leave `ExpiresAt` zero when there's no real deadline. `pendingPoolView` already guards on `!p.ExpiresAt.IsZero()`, so it would then correctly omit `approval_expires_at` for sync parks, and the dashboard can render "no expiry" instead of a fake clock.

Please keep the field where a deadline **is** real, though — don't blanket-drop it:
- async-grant operations carry a genuine `ApprovalExpiresAt` via `pendingActionView` (the DB path), and
- a `human_approver` with a `timeout` sets `pending.ExpiresAt = CreatedAt + timeout` itself (human.go:133) and actually enforces it.

Both of those flow through the same pool, so the distinction is exactly "did someone set a real ExpiresAt" vs "did `Add` invent one" — which is why fixing it in `Add` (rather than in `pendingPoolView`) keeps the real human_approver/async expiries intact while dropping only the fabricated one.

If touching `Add` feels out of scope for this PR, the narrower alternative is to have `pendingPoolView` omit `approval_expires_at` for entries with no operation id whose `ExpiresAt` came from the default — but that's harder to tell apart at that layer, so the root fix reads cleaner. Happy to push either.

arnauorriols:
[bot-reply]
Pushed e09b6f48 addressing this — the root fix you suggested:

- `HITLRegistry.Add` no longer fabricates `ExpiresAt = CreatedAt + 30m`. It now leaves `ExpiresAt` zero whenever the caller set no real deadline, so a pure-sync `builtin.dashboard` park carries no expiry.
- `pendingPoolView` already guards on `!p.ExpiresAt.IsZero()`, so `/pending` now correctly omits `approval_expires_at` for sync parks instead of advertising a clock the gateway never honors.
- Real deadlines are untouched: `human_approver` sets `pending.ExpiresAt = CreatedAt + timeout` itself (and enforces it), and async-grant ops carry `ApprovalExpiresAt` through the DB path (`pendingActionView`). Both still surface their expiry.
- Added `TestHITLRegistryAddDoesNotFabricateExpiryForSyncPark` to lock this in: a sync park leaves `ExpiresAt` zero and `pendingPoolView` omits the field, while an entry with a real deadline keeps it.

The dashboard's `/api/hitl/pending` serializes the struct directly and `HITLBar.tsx` doesn't render `expires_at`, so no UI change was needed.

arnauorriols:
fix lint

arnauorriols:
[bot-reply]
Pushed 181908df addressing this comment:
- Renamed the `real` variable in `TestHITLRegistryAddPreservesDeadline` (cmd/clawpatrol/hitl_registry_test.go) to `found`, clearing the `revive` redefines-builtin-id lint failure (`real` shadows the Go builtin). `golangci-lint run` is now clean on the affected packages and the test still passes.

PR #676, opened 2026-06-12
hitl: open the async approval window after the synchronous hold

arnauorriols:
## Problem

Polling the approval status of a parked HITL request could return `expired` for an operation that was still pending. An agent reasonably reads `expired` as terminal and gives up on a request a human could still approve.

The async approval window (`ApprovalExpiresAt`) was computed at creation as `now + approval_ttl`, which folds in the synchronous hold (`sync_wait_timeout`). When `sync_wait_timeout` approaches or exceeds `approval_ttl`, the operation is already past its deadline the moment it transitions `sync_waiting -> pending_approval` and the client receives its `202` poll handle — so the very first poll comes back `expired`.

## Fix

- Measure the approval TTL from the end of the synchronous hold (`sync_wait_deadline + approval_ttl`) instead of from creation, so the poll window opens when polling begins and always grants the full `approval_ttl`. `expired` is now returned only once the window has genuinely elapsed. (The approved-retry window is already re-based at approval time; the approval window had no equivalent guard.)
- Surface `poll_again` and `retry_after_seconds` on the non-terminal poll-status body so an agent reading the JSON — not just the `Retry-After` header — knows to wait and poll again rather than treat the response as terminal.

## Tests

- `hitlOperationStatusBody`: pending/sync_waiting return non-terminal poll-again guidance and never `expired`; expired and denied stay terminal; approved returns retry headers.
- `TestHITLAsyncApprovalWindowExcludesSyncWait`: with `sync_wait_timeout == approval_ttl`, asserts `approval_expires_at` outlives `sync_wait_deadline` and that a sweep at the poll moment does not expire the operation (fails under the old `now + approval_ttl` arithmetic).

### DIFF
--- a/cmd/clawpatrol/hitl_async_e2e_test.go
+++ b/cmd/clawpatrol/hitl_async_e2e_test.go
@@ -6,6 +6,7 @@ import (
 	"crypto/tls"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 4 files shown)


arnauorriols:
[bot-reply]
Reviewed the change against the PR description — the body holds up and the implementation is sound. No changes requested.

- **Root cause / fix**: `ApprovalExpiresAt` now bases off `syncDeadline + approval_ttl` (`hitl_async_runtime.go`) instead of `now + approval_ttl`, so the poll window opens when polling begins and always grants the full TTL. When `sync_wait_timeout` rivals or exceeds `approval_ttl`, the old arithmetic left an operation past its deadline the instant it became pollable; this removes that.
- **Approval vs. retry windows**: confirmed the asymmetry the body calls out — the approved-retry window is already re-based at approval time (`resolveAsyncHITLGrant`: `RetryExpiresAt` is reset to `Now + retry_ttl` when it's no longer after `Now`), whereas the approval window had no equivalent guard. This fix gives the approval window the same protection at creation time.
- **Poll-status body**: `poll_again` + `retry_after_seconds` on the non-terminal (`sync_waiting` / `pending_approval`) responses is a good call — an agent reading the JSON now gets the same wait signal as the `Retry-After` header. Terminal states (`expired`, `denied`) correctly omit it.
- **Tests**: `TestHITLAsyncApprovalWindowExcludesSyncWait` pins the regression (asserts `approval_expires_at > sync_wait_deadline` and that a sweep at the poll moment doesn't expire the op), and `hitlOperationStatusBody` covers the terminal/non-terminal vocabulary. CI (lint + test) is green.

LGTM.

PR #682, opened 2026-06-14
feat: add dashboard branding config

rametta:
## Summary
- Add gateway config for customizing the dashboard name, logo, and icon.
- Apply branding to the dashboard header, login page, state API, and docs.

CleanShot 2026-06-14 at 10 32 37@2x

 
## Test plan
- go test ./internal/config
- go test ./cmd/clawpatrol
- dashboard format/lint/build checks


rametta:
Hey there 👋 CI is failing because of missing DEMO_DEPLOY_TOKEN. Not sure how we'd like to resolve that. Lmk!

piscisaureus:
Sorry for the three-month silence on this. Being straight about where it stands: dashboard theming is not something we plan to take into the gateway right now (it is listed under "not planned" in the roadmap, #818). The reasoning is that every dashboard knob is surface we maintain and test across releases, and the dashboard's job is operator control of the gateway rather than a product surface for end users. If there is a concrete deployment need behind this (for example, telling several gateways apart at a glance), an issue describing that need might lead to a much smaller change, such as a display name in the header. I'll close this in a week or so unless you'd like to keep it open for that discussion.

PR #683, opened 2026-06-15
daemon: auto-recover WG tunnel + DNS after a gateway restart

arnauorriols:
## Problem

When the gateway is restarted, devices with an open tunnel stop resolving DNS and do **not** self-recover. The only known fix is to kill the device-side daemon, after which DNS comes back.

## Root cause

A gateway restart drops all in-memory WireGuard peer/session state. The device keeps its locally-configured peer and its old keypair, treating it as valid until `RejectAfterTime` (180s), so it keeps encrypting with a session the gateway can no longer decrypt and `last_handshake_time_sec` freezes. When the keypair finally expires and the device re-initiates, the wireguard-go handshake state machine can wedge in `handshakeInitiationCreated` ("Failed to derive keypair") with no path out — black-holing every flow, including tunnelled DNS (UDP/53 forwarded through the tunnel), until the daemon is killed and the peer is rebuilt from scratch.

A recovery watchdog (`runWGWatchdogLoop`) already existed for exactly this stuck-handshake class — it rebuilds the peer to force a clean re-handshake, which is precisely what a manual daemon restart does. But it was never started in production (only referenced from tests).

## Fix

Wire the watchdog into the WG transport lifecycle in `startWGTransport`:

- Poll the peer's handshake age; when it goes stale (> 3m, mirroring `RejectAfterTime`) while traffic is still being staged, `IpcSet` the peer to force a clean fresh initiation.
- Wrap the device logger so the "Failed to derive keypair" log line short-circuits the poll cadence and recovery starts within milliseconds.
- Tear the watchdog down on transport `Close` so it can't outlive its transport.

Tunnelled DNS self-heals once the tunnel re-handshakes (the UDP forwarder's flows carry a 30s read deadline and re-dial), so no DNS-specific change is needed.

## No regression

Healthy tunnels rekey well within the 3m stuck-timeout (REKEY_AFTER_TIME 120s, keepalive 10s), so `last_handshake` stays fresh and the watchdog never fires during normal operation. It only acts on a genuinely wedged handshake (stale age **and** advancing tx) or the explicit keypair-derivation signal.

Scope: WireGuard mode only — the transport in the repro. tsnet mode has tailscale's own reconnect logic.

## Tests

- Added `daemon_transport_wg_linux_test.go` covering the watchdog teardown on `Close` (and nil-safety).
- Existing `wg_watchdog_test.go` covers the loop behaviour.
- Full `cmd/clawpatrol` suite green; `gofmt`/`go vet` clean.

### DIFF
--- a/cmd/clawpatrol/daemon_transport_wg_linux.go
+++ b/cmd/clawpatrol/daemon_transport_wg_linux.go
@@ -49,6 +49,19 @@ const (
 	// so the first user flow doesn't race the initial handshake. See
 	// macos/netstack/wgnetstack.go for the same rationale.
 	wgClientKeepalive = 10
+	// wgWatchdogPoll is how often the handshake-recovery watchdog samples
+	// the peer's last_handshake_time_sec + tx_bytes. The watchdog only
+	// acts once the handshake is older than wgWatchdogStuckTimeout, so a
+	// few seconds between polls is plenty and stays cheap.
+	wgWatchdogPoll = 5 * time.Second
+	// wgWatchdogStuckTimeout mirrors wireguard's RejectAfterTime (180s):
+	// below this a stale last_handshake is normal (no rekey was due), so
+	// rebuilding the peer would only add churn. Past it, with traffic
+	// still being staged, the handshake is genuinely wedged.
+	wgWatchdogStuckTimeout = 3 * time.Minute
+	// wgWatchdogResetCooldown holds off back-to-back rebuilds so a fresh
+	// initiation has time to complete before the watchdog fires again.
+	wgWatchdogResetCooldown = time.Minute
 )
 
 type wgTransport struct {
@@ -56,6 +69,10 @@ type wgTransport struct {
 	tun         *wgClientTun
 	localAddr   netip.Addr
 	bootWarning string
+
+	// stopWatchdog cancels the handshake-recovery watchdog goroutine.
+	// Set by startWGTransport, invoked by Close.
+	stopWatchdog context.CancelFunc
 }
 
 func (t *wgTransport) LocalAddr() netip.Addr { return t.localAddr }
@@ -97,6 +114,9 @@ func (t *wgTransport) Dial(ctx context.Context, network, addr string) (net.Conn,
 }
 
 func (t *wgTransport) Close() error {
+	if t.stopWatchdog != nil {
+		t.stopWatchdog()
+	}
 	if t.dev != nil {
 		t.dev.Close()
 	}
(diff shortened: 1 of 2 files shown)


arnauorriols:
The bug has been experienced in tailscale control mode, without using --whole-machine, running the agent with `clawpatrol run <agent>`. Make sure this exact scenario is covered by your investigation

arnauorriols:
[bot-reply]
Confirmed, and you're right to flag it: this PR does **not** cover the Tailscale-control-mode scenario. The fix is WireGuard-mode only.

The daemon picks its transport from the `mode` marker written at join time (`daemonStartTransport`):

- `wireguard` → `startWGTransport`: a raw wireguard-go device with a fixed gateway endpoint. This is the path that wedges — on a gateway restart the device keeps encrypting with a dead session, then the state machine can stick in `handshakeInitiationCreated` and black-hole everything, including tunnelled DNS. The watchdog in this PR rebuilds that peer automatically.
- `tailscale` → `startTsnetTransport`: an embedded `tsnet.Server` with the gateway pinned as exit node. DNS goes to `<tailnet-gateway-ip>:53` via exit-node routing (`childResolvConf`). This is your scenario, and it never touches the wireguard-go device the watchdog guards.

The two modes also have independent DNS paths (WG intercepts in-flight; tsnet routes DNS through the exit node), so they're genuinely separate failure modes — not one bug behind two front-ends.

What I believe breaks tsnet mode on a gateway restart: exit-node routing fails closed, so until the device's tsnet session + netmap to the gateway recover, every exit-routed flow (including DNS) is dropped. There is currently **no ongoing health check** on this path — the `gateway:53` reachability probe runs once at boot in `startTsnetTransport` and never again. If the session/netmap stays stale after the gateway (and its control endpoint) restart, nothing re-establishes it until the daemon is restarted — exactly the "kill the daemon to recover" symptom.

Proposed tsnet-mode fix, mirroring the WG watchdog: a periodic health probe (re-dial `gateway:53` through the tsnet server) and, on sustained failure, kick recovery — re-apply the exit-node pref and, if that doesn't restore the path, rebuild the tsnet node (the tsnet equivalent of the WG peer rebuild a daemon restart does today).

One decision before I touch code: do you want that folded into this PR, or tracked as a follow-up so this PR stays scoped to the WireGuard wedge it was opened for?

piscisaureus:
Reviewed against current main (rebases cleanly, vets clean for linux). I can't land it in this form, for three reasons:

1. **It doesn't address the reported case.** The report in the thread was tsnet control mode with `clawpatrol run`; the watchdog only guards the wireguard-go device, and the tsnet transport still has no health probe after its one-time boot check. That remains an open problem on main.
2. **The root-cause story doesn't match wireguard-go or the gateway.** The gateway persists its server key and replays `wg_peers` at boot, so after a restart the peer exists under the same key. wireguard-go arms a new handshake 15s after unacknowledged data and retries initiation every 5s; "Failed to derive keypair" skips without completing the handshake timers, so it retries rather than wedging. If WG mode really does stay down after a gateway restart, the PR hasn't shown why (a log excerpt with the recurring failure past the retransmit window would settle it).
3. **The reset can make an outage permanent.** `reset` rebuilds the IPC string with a live DNS lookup of the endpoint host; when that lookup fails, `buildDaemonWGIpc` omits `endpoint=` while keeping `replace_peers=true`, so the peer is replaced with one that has no endpoint. tx then never advances, the `txBytes == lastTx` guard suppresses every further reset, and the tunnel is dead until the daemon restarts. Codex flagged the same path independently. Also, with `persistent_keepalive_interval=10` tx advances on every keepalive, so the "traffic still staged" guard never holds and the loop fires on any outage over 3 minutes.

If the WG-mode wedge is real, a defensive watchdog could land with: the IPC string built once at startup (or the reset skipped when resolution fails), the tx guard fixed, and a title that doesn't claim the DNS/restart fix. The tsnet recovery needs its own change. Happy to review either.

PR #684, opened 2026-06-15
Add OTel GenAI telemetry export with optional message content

arnauorriols:
## Summary

Adds an opt-in OpenTelemetry export that maps intercepted LLM traffic to the [GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) (semconv v1.27.0), so LLM calls flowing through the gateway show up in standard OTel/GenAI observability tooling.

For each intercepted turn the gateway emits a `chat {model}` span carrying:

- `gen_ai.system` (`anthropic` / `openai`)
- `gen_ai.operation.name`
- `gen_ai.request.model` / `gen_ai.response.model`
- `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`
- `gen_ai.response.finish_reasons`

Span duration reflects the real upstream round-trip latency.

## Config

Two independent switches under the `gateway {}` block:

```hcl
genai_telemetry {
  include_message_content = false
}
```

- **Block presence** enables base span export. It reuses the existing OTLP exporter, so `OTEL_EXPORTER_OTLP_ENDPOINT` must be set.
- **`include_message_content`** additionally attaches prompt/completion content as GenAI content events (`gen_ai.{role}.message`, `gen_ai.choice`). Defaults **off** — content can be large and sensitive — and is independent of the base toggle.

When the block is absent the export path returns before parsing anything, so disabled behavior is unchanged with no added per-request overhead.

## Scope

Base attribute spans emit for all intercepted providers (Anthropic, OpenAI, codex). Message-content extraction is implemented for Anthropic (both non-streaming JSON and SSE); OpenAI/codex content extraction is left as a follow-up.

## Tests

- `internal/config`: block parsing + accessor behavior (absent / enabled / content opt-in / nil-safe).
- `cmd/clawpatrol`: span attributes, content-event gating, Claude request/response content parsers (JSON + SSE), and end-to-end `recordGenAITurn` for disabled / enabled / content-on paths using an in-memory span recorder.

### DIFF
--- a/cmd/clawpatrol/genai.go
+++ b/cmd/clawpatrol/genai.go
@@ -1,15 +1,16 @@
 // OpenTelemetry GenAI semantic-convention export for intercepted LLM
-// turns. Targets semconv v1.27.0
+// turns. Targets the GenAI semantic conventions
 // (https://opentelemetry.io/docs/specs/semconv/gen-ai/): one span per
 // model invocation carrying gen_ai.* attributes (system, models, token
 // usage, finish reason) and — only when the operator opts in — the
-// prompt/completion message content as span events.
+// prompt/completion message content as the gen_ai.input.messages /
+// gen_ai.output.messages / gen_ai.system_instructions span attributes.
 //
 // Opt-in is two independent switches (internal/config GenAITelemetry):
 // the `genai_telemetry {}` block presence enables the attribute span;
-// `include_message_content` additionally attaches content events.
-// Disabled is the zero-overhead default — recordGenAITurn returns
-// before parsing anything.
+// `include_message_content` additionally attaches the message-content
+// attributes. Disabled is the zero-overhead default — recordGenAITurn
+// returns before parsing anything.
 
 package main
 
@@ -31,6 +32,22 @@ type genAIMessage struct {
 	Content string
 }
 
+// genAIPart is a single content part within a GenAI message. Only text
+// parts are captured today, so Type is always "text".
+type genAIPart struct {
+	Type    string `json:"type"`
+	Content string `json:"content"`
+}
+
+// genAIChatMessage is one message in the gen_ai.input.messages /
+// gen_ai.output.messages span attributes: a role, its content parts,
+// and (output messages only) the finish reason for that message.
+type genAIChatMessage struct {
+	Role         string      `json:"role"`
+	Parts        []genAIPart `json:"parts"`
+	FinishReason string      `json:"finish_reason,omitempty"`
+}
+
 // genAITurn is one intercepted LLM request/response mapped to OTel
 // GenAI semantic-convention terms.
 type genAITurn struct {
@@ -47,16 +64,18 @@ type genAITurn struct {
 	// stamped at emission time.
 	Start time.Time
 
-	// Messages and Completion populate the content events; filled only
-	// when message-content capture is enabled.
+	// Messages and Completion populate the content attributes; filled
+	// only when message-content capture is enabled.
 	Messages   []genAIMessage
 	Completion string
 }
 
 // emitGenAISpan records one GenAI span on the provided tracer. When
-// includeContent is true, message content is attached as span events
-// per the GenAI content convention. A free function (not a method) so
-// tests can drive it with an in-memory tracer.
+// includeContent is true, message content is attached as the
+// gen_ai.input.messages / gen_ai.output.messages /
+// gen_ai.system_instructions span attributes per the GenAI semantic
+// conventions. A free function (not a method) so tests can drive it
+// with an in-memory tracer.
 func emitGenAISpan(tracer trace.Tracer, t genAITurn, includeContent bool) {
 	if tracer == nil {
 		return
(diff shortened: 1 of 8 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR description. It's a self-contained summary of the change (scope, config switches, and test coverage) and contains no question or change request to act on, so no code changes were made. The noted follow-up (OpenAI/codex message-content extraction) is already called out as out of scope for this PR. Flagging as addressed; let me know if any part of the description was intended as an action item.

arnauorriols:
switch from per-role AddEvent calls to the gen_ai.input.messages / gen_ai.output.messages / gen_ai.system_instructions span attributes

arnauorriols:
[bot-reply]
Pushed d1827f2a addressing this comment:
- Removed the per-role `span.AddEvent` calls (`gen_ai.{role}.message`, `gen_ai.choice`).
- Message content now rides span attributes per the current GenAI semantic conventions: system messages → `gen_ai.system_instructions`, input messages → `gen_ai.input.messages`, completion → `gen_ai.output.messages` (with `finish_reason`). Each value is JSON-serialized since OTel attribute values are primitives.
- Updated the `include_message_content` doc comment, example HCL, regenerated `site/doc/config-reference.md`, and reworked the content tests to assert the new attributes.

arnauorriols:
Add gen_ai.conversation.id so multi-turn sessions can be correlated

arnauorriols:
[bot-reply]
Pushed 3c3dc035 addressing this comment:
- Added `gen_ai.conversation.id` to the GenAI span. It's sourced from the same session correlation key already used for LLM usage tracking — Claude Code's metadata session id (falling back to a hash of the first user message), and the codex session hint for the OpenAI/codex paths. Multi-turn sessions can now be grouped in trace backends per the GenAI semantic conventions.
- Covered by `genai_test.go` (asserts the attribute on both the emit and record paths).

arnauorriols:
Support all block types, not just text

arnauorriols:
Could you add the remaining GenAI semconv attributes the gateway already has the data for — gen_ai.provider.name (Required; currently the removed gen_ai.system), gen_ai.conversation.id, gen_ai.request.stream, the request params (max_tokens/temperature/top_p/top_k/stop_sequences), gen_ai.response.id, server.address/server.port, error.type, and the Anthropic-specific cache-token breakdown (gen_ai.usage.cache_read.input_tokens / gen_ai.usage.cache_creation.input_tokens) — plus the tool_call/tool_call_response message parts?

arnauorriols:
[bot-reply]
Pushed 79792457 addressing this comment:
- Message content no longer flattens everything to text and drops the rest. Each Anthropic content block now maps to its GenAI message part per the [message-part conventions](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-input-messages.json):
  - `text` → `text` part
  - `tool_use` / `server_tool_use` → `tool_call` part (`id`, `name`, `arguments`)
  - `tool_result` → `tool_call_response` part (`id`, `response`)
  - `thinking` → `reasoning` part
  - any other block (`image`, `document`, `redacted_thinking`, …) → a generic part that preserves the block `type` so it is recorded rather than silently dropped. Raw binary payloads (e.g. base64 image data) are intentionally not captured.
- Applies to both the request messages and the assistant response, across non-streaming JSON and streaming SSE (tool-call `arguments` are reassembled from `input_json_delta` fragments, blocks tracked by index).
- Tests cover the new block types on both the request and response paths (JSON + SSE), including tool_use/tool_result/thinking/generic-image.

arnauorriols:
Could you add the remaining GenAI semconv attributes the gateway already has the data for — gen_ai.provider.name (Required; currently the removed gen_ai.system), gen_ai.conversation.id, gen_ai.request.stream, the request params (max_tokens/temperature/top_p/top_k/stop_sequences), gen_ai.response.id, server.address/server.port, error.type, and the Anthropic-specific cache-token breakdown (gen_ai.usage.cache_read.input_tokens / gen_ai.usage.cache_creation.input_tokens) — plus the tool_call/tool_call_response message parts?

arnauorriols:
[bot-reply]
Pushed `4d0b4010` adding the remaining GenAI semconv attributes the gateway already had data for. Each emitted span now also carries:

- **`gen_ai.provider.name`** — replaces the removed `gen_ai.system` (same "anthropic"/"openai" value).
- **`server.address` / `server.port`** — the upstream host (threaded through from the MITM `host`), port `443` (the LLM endpoints are HTTPS-only).
- **`gen_ai.request.stream`** plus the sampling params **`gen_ai.request.max_tokens` / `.temperature` / `.top_p` / `.top_k` / `.stop_sequences`** — parsed from the Anthropic `/v1/messages` request body. Optional scalars use pointers so a real `temperature: 0` is distinguished from "unset".
- **`gen_ai.response.id`** — from the response body (JSON `id` / SSE `message_start`).
- **`gen_ai.usage.cache_read.input_tokens` / `gen_ai.usage.cache_creation.input_tokens`** — the Anthropic prompt-cache token breakdown (JSON `usage` or SSE `message_start`). These are still folded into `input_tokens` for billing and now also surfaced separately.
- **`error.type`** — from a provider error response: a top-level `{"type":"error","error":{"type":...}}` body or an SSE `error` event.

Already emitted before this commit, so unchanged: **`gen_ai.conversation.id`** and the **`tool_call` / `tool_call_response`** message parts (mapped in `claudeBlockPart` from `tool_use`/`tool_result`).

One scope note on `error.type`: the telemetry hook (`trackLLMUsage`) currently only runs on a `200` upstream response, so this captures provider-signalled errors that ride a 200 body/stream. HTTP transport-level failures (non-2xx, 502) don't reach the hook today — wiring those would mean buffering non-200 responses into the telemetry path. Happy to do that as a follow-up if you'd like full transport-error coverage.

Request-param/response-meta parsing is implemented for the Anthropic path (where the content parsing already lives); the OpenAI/Codex paths still get the generic attributes (provider name, server address, conversation id, models, token usage). Tests added in `genai_test.go` cover the new parsers and assert every new attribute lands on the span.

arnauorriols:
Add gen_ai.tool.definitions (name+type at base, full schema under content opt-in)

arnauorriols:
[bot-reply]
Pushed e6e617ed addressing this comment:
- Added `gen_ai.tool.definitions` to the GenAI span, parsed from the Anthropic request body's `tools`.
- Base span carries each tool's `name` + `type` (absent Anthropic type → GenAI default `"function"`; built-in tools keep their declared type, e.g. `web_search_20250305`).
- Full schema (`description` + `input_schema` → `parameters`) is gated behind the existing `include_message_content` opt-in, since schemas can be large/proprietary.
- Updated docgen source comment (regenerated `config-reference.md`) and `gateway.example.hcl`; added unit tests for both the base (name+type only) and content-on (with schema) cases.

PR #685, opened 2026-06-15
feat(config): add basic auth credential

rajiteh:
* Add a built-in `basic_auth` credential that injects
  `Authorization: Basic <base64(username:password)>`. The username
  lives in config; the secret slot holds the raw password.
* Disambiguation reuses the HTTPS placeholder detector, which already
  base64-decodes `Basic` headers, so multiple basic_auth credentials
  bound to one endpoint route by their `placeholder` value.
* Add emit, dispatch, and runtime unit coverage, dashboard labeling,
  and config-reference docs.
* `docgen` now trims trailing newlines to exactly one for all
  generated output, not just basic_auth; this drops the stray trailing
  blank line at the end of `config-reference.md`.


rajiteh:
Addressed the review feedback by documenting the common credential binding fields for `basic_auth`, the placeholder dispatch behavior, and a minimal working HTTPS example. Also updated the dashboard label casing to `Basic Auth`.

On the design question: Basic auth can be represented with the existing `header_token` credential if the stored secret is already the full pre-encoded value:

```hcl
credential "header_token" "api" {
  endpoint = https.api
  header   = "Authorization"
  prefix   = "Basic "
}
```

In that shape the secret must be `base64(username:password)`. The reason this PR still adds `basic_auth` is to keep the secret material as the raw password and keep the username in config, then let the gateway construct the standard Basic header with `SetBasicAuth`. That mirrors how operators usually think about Basic credentials and avoids requiring them to pre-encode or rotate a combined `username:password` payload.

Placeholder dispatch does not need special credential-plugin support here: the existing HTTPS endpoint detector already decodes `Authorization: Basic <base64>` and searches the decoded payload for the configured placeholder. This plugin is only about outbound injection ergonomics and secret shape.

If maintainers prefer not to add a first-class credential type for this, the fallback is straightforward: drop the plugin/tests and document the `header_token` pattern above for Basic auth.

PR #688, opened 2026-06-16
relay: fail open when the auto-expose relay dies

dhruvkelawala:
## Problem

`clawpatrol run`'s auto-expose relay installs an iptables `nat`/`OUTPUT` `REDIRECT` for `127.0.0.0/8` so the wrapped process can reach host-loopback services. **That rule is never removed.** If the relay stops forwarding while the wrapped process keeps running, the `REDIRECT` outlives the forwarder and points at a now-closed port, so **every** loopback `connect()` in the agent netns is black-holed — including the wrapped process reaching **its own** in-netns listeners. Loopback fails **closed**.

The relay can stop while the agent keeps running because:

- `runRelaySupervisor`'s `notifRecv` loop returns on *any* non-`ENOENT` error — e.g. a transient `EINTR`, or `ENOTTY` (`inappropriate ioctl for device`) from `SECCOMP_IOCTL_NOTIF_RECV`. The diagnostic line was gated behind `CLAWPATROL_DEBUG`, so the relay died silently.
- The `relay-worker` then hits EOF on the control socket and exits, stranding the `REDIRECT` it installed.

Observed with a long-running `clawpatrol run` wrapping a service that exposes a local HTTP port and also pushes OTLP telemetry to a host-loopback collector: after some hours `relay-worker`/`relay-supervisor` became `<defunct>`, and from then on the wrapped service could reach neither the host collector nor **its own** local port (`ECONNREFUSED`) — despite that listener being up in the same netns. Restarting restored it only until the relay next died.

## Fix

Make the relay **fail open** instead of fail closed.

1. **Tear down the `REDIRECT` when the worker stops forwarding.** `relay-worker` now removes its rules on exit via a `defer` (covers the supervisor-died / EOF path) and a `SIGTERM`/`SIGINT` handler (Pdeathsig would otherwise skip the defer). Adds `uninstallLoopbackRedirectRules` / `loopbackRedirectDeleteArgs` (the install args with `-A`→`-D`; `iptables -D` matches on the full spec). `setupHostLoopbackForwarder` now returns the forwarder port and rolls back a partially-installed rule on error. With the rule gone, loopback reverts to direct, in-netns behaviour: in-netns listeners are reachable again and host-loopback fails fast instead of being black-holed.
2. **Make the supervisor resilient.** The `notifRecv` loop now retries `EINTR`/`EAGAIN` instead of tearing everything down, and surfaces any other fatal error unconditionally (no longer behind `CLAWPATROL_DEBUG`) so an operator can see when host-loopback forwarding has stopped.

## Scope

Client-side only. `relay-supervisor` / `relay-worker` are re-exec'd solely by `clawpatrol run`; the `gateway` server does not use this path.

## Testing

- `go test ./cmd/clawpatrol` passes, including a new `TestLoopbackRedirectDeleteArgs`; `gofmt` and `go vet` clean.
- **e2e (the regression):** wrap `bash`, start an in-netns listener on `127.0.0.1:<port>`, then kill the relay mid-run.
  - *Before:* connecting to the in-netns listener returns connection refused (black-holed by the stranded `REDIRECT`).
  - *After:* the `REDIRECT` is removed on relay exit and the in-netns listener is reachable again.
- Deployed on a live host: relay processes stay healthy, the wrapped service's self-connection errors stop, and OTLP telemetry resumes reaching the host collector.

## Notes

- IPv6 (`::1`) host-loopback is still not handled (pre-existing; the install path only adds IPv4 `nat`/`OUTPUT` rules), so there is nothing extra to tear down there.
- No change to the auto-expose reverse direction (mirroring in-netns listeners back to the host), which uses host `net.Listen` and no iptables.

### DIFF
--- a/cmd/clawpatrol/relay_linux.go
+++ b/cmd/clawpatrol/relay_linux.go
@@ -276,7 +276,19 @@ func runRelaySupervisor(_ []string) {
 			if errors.Is(err, unix.ENOENT) {
 				return
 			}
-			relayDebugf("[clawpatrol relay] notif_recv: %v\n", err)
+			// Transient: a signal or spurious wakeup interrupted the ioctl.
+			// Keep serving notifications instead of tearing the whole relay
+			// down; a single EINTR must not strand the agent netns.
+			if errors.Is(err, unix.EINTR) || errors.Is(err, unix.EAGAIN) {
+				continue
+			}
+			// Anything else means we can no longer service notifications.
+			// Surface it unconditionally (not just under CLAWPATROL_DEBUG):
+			// the worker will now fail the REDIRECT open and host-loopback
+			// forwarding stops, so operators can see why telemetry/webhooks
+			// to host services went quiet.
+			fmt.Fprintf(os.Stderr, "⚠ clawpatrol relay: notification channel failed (%v); "+
+				"host-loopback forwarding disabled, reverting wrapped process to direct loopback\n", err)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


piscisaureus:
Landed on `main` as 1f77140d (via #708), authored to you — thanks @dhruvkelawala. I rebased onto current `main` and added one test assertion (teardown args == install args modulo `-A`/`-D`, with a length guard). A security review confirmed the fail-open is safe — removing the agent-netns REDIRECT only reduces the wrapped process's reach and is orthogonal to gateway egress — and I validated the iptables install→delete round-trip on a real Linux host.

PR #689, opened 2026-06-16
Design: external plugin protocol v2 (parity with built-in plugins)

piscisaureus:
Draft RFC. A design doc only — no code, nothing implemented. Opening it
as a PR so we can iterate on the design in line comments.

The distribution work (#687) shipped the v1 external plugin protocol:
a plugin can declare credentials, tunnels, endpoints, and facets. An
audit of the built-in plugins showed only the leaf integrations
(~10 header credentials, a generic HTTPS endpoint) extract cleanly;
everything protocol-aware, stateful, or HITL-bound stays built-in.

This doc proposes the v2 extensions that would let an external plugin
reach parity with every built-in, without weakening the sandbox /
least-privilege / signed-manifest / fail-closed-escalation model. Most
of it is exposing contracts that already exist gateway-side
(`BlobStore`, `ApproverRuntime`, `HITLPool`, `Secret`) over gRPC.

Primitives proposed, by milestone:

* per-plugin **state service** (BlobStore over gRPC) — ssh host keys,
  codex jwt keys, tailscale identity, notion-mcp registration
* generalized **credential injection** (URL/body rewrite, full-body
  signing, brokered mTLS client cert) — telegram, discord, aws, mtls
* **built-in family action mapping** so external sql/k8s/ssh endpoints
  reuse the shared rule vocabulary instead of forking it
* **approver plugins** + the **HITL** surface (notify / decide /
  webhook ingress) — dashboard/human/llm approvers, slack
* operator-only **exec** grant, tunnel **`via`** chaining, credential
  **placeholder** hints

Each section states the problem, the proposed gRPC surface, the
security tier (low-risk plugin-declarable vs operator-only), and which
built-ins it unblocks. Open questions are called out inline for
discussion — webhook ingress model, SigV4 body cap, action-schema
versioning, state quotas.

Doc: `doc/external-plugin-protocol-v2-design.md`.

### DIFF
--- a/doc/external-plugin-protocol-v2-design.md
+++ b/doc/external-plugin-protocol-v2-design.md
@@ -30,9 +30,11 @@ built-in side of the line:
 * **Non-header credential injection** — telegram (URL/body), discord
   (body/WebSocket), AWS SigV4 (full-body signing), mTLS (upstream
   client cert).
-* **The shared facet vocabulary** — `sql.*`, `k8s.*`, `ssh.*` matchers
-  are gateway-only; an external endpoint can only emit to a namespaced
-  `<plugin>.<facet>`, forking the policy vocabulary.
+* **The shared `sql` vocabulary** — the `sql.*` matcher is gateway-only,
+  so an external SQL endpoint forks to a namespaced `<plugin>.<facet>`
+  and operators rewrite every rule. (k8s/ssh have one implementation
+  each, so a plugin-private facet is fine there — this is only a problem
+  for SQL, which has many engines.)
 * **Process-spawning tunnels** — `kubectl port-forward`, `local_command`
   — the sandbox forbids exec of anything but the plugin's own binary.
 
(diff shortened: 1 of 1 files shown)


piscisaureus:
Resolved §4 (shared facet vocabulary), narrowed to `sql` only.

Reasoning: a shared vocabulary doesn't reduce plugin work (the plugin
produces the parse either way) — it only buys rule portability across
multiple implementations of one family, plus zero-rewrite parity if a
built-in is externalized. That payoff scales with how many
implementations of a family an operator runs:

* **k8s, ssh** — one implementation each. No portability win; a shared
  vocabulary is just a frozen compat surface. Dropped — they use
  plugin-defined facets, which already work (no machinery).
* **sql** — many engines, and the coarse layer (verb / tables /
  functions / statement) genuinely generalizes for guardrail policy.
  Kept as the one opt-in built-in family.

The shared `sql` schema stays deliberately minimal/coarse — making it
comprehensive would silently under-match engine-specific semantics
(ClickHouse `SYSTEM`, BigQuery scripting, ...) and give a false sense of
coverage. Engines that need more declare their own facet, or phase-2
`extends = "sql"`.

Updated in 6b9b397d.

PR #690, opened 2026-06-16
Add a persistent state service for external plugins (v2 M1)

piscisaureus:
First milestone of the external plugin protocol v2 work (see #689).

A sandboxed plugin has no writable filesystem, so it cannot persist the
small amount of identity it needs across restarts: an SSH endpoint's
host key, a signing keypair, a dynamically registered client id. This
adds a gateway-backed key/value store the plugin reaches over the
go-plugin broker.

* New gateway-served `HostState` gRPC service (Get/Put). Every other
  service in the protocol is gateway-calls-plugin; this one is the
  reverse — it runs on the gateway and the plugin is the client. The
  plugin dials it on a fixed reserved broker stream id, so it is reachable
  from every callback (`Build` / `InjectHTTP` / `HandleConn` / `Dial`),
  not just within one connection. A fixed id is safe because clawpatrol
  never calls `GRPCBroker.NextId()`.
* Backed by the existing sqlite `BlobStore`, namespaced per plugin by the
  gateway — the namespace never comes from the wire, so one plugin can
  never read or write another's keys. Values capped at 1 MiB. Resolved
  lazily per call, so a plugin spawned during the first config load still
  gets state once the store is wired (the state dir it lives in is itself
  part of that config, so it cannot be ready earlier).
* SDK: `pluginsdk.State()` returns a handle (`Get`/`Put`) valid from any
  callback; it lazily dials the broker and caches the connection, and
  returns an error against a gateway too old to provide the service.

Tested over an in-memory server (round-trip, namespace isolation, key
validation, the size cap, the not-ready path) and end to end over a real
go-plugin broker (`TestPluginGRPCConn`).

Unblocks the stateful built-ins identified in the audit: ssh host keys,
codex jwt keys, tailscale identity, notion-mcp registration.

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -3013,6 +3013,11 @@ func runGateway(args []string) {
 	logDashboardAuthState(db, cfg)
 	blobs := newGatewayBlobStore(db)
 	endpoints.SetBlobStore(blobs)
+	// Back the external-plugin HostState service with the same sqlite blob
+	// store. Wired after the first config load (the state dir it lives in
+	// is part of that config); the service resolves the store lazily, so
+	// plugins spawned during that load still get state at request time.
+	pluginMgr.SetBlobStore(blobs)
 	certs, err := loadOrMintCA(db)
 	if err != nil {
 		log.Fatalf("ca: %v", err)
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -3016,7 +3016,10 @@ func runGateway(args []string) {
 	// Back the external-plugin HostState service with the same sqlite blob
 	// store. Wired after the first config load (the state dir it lives in
 	// is part of that config); the service resolves the store lazily, so
-	// plugins spawned during that load still get state at request time.
+	// plugins spawned during that load still get state from their runtime
+	// callbacks (HandleConn/InjectHTTP/OpenTunnel/Dial). State is not
+	// available during a plugin's Build callback on this first load — see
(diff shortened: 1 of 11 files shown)


piscisaureus:
Ran a review pass over the diff. Design checked out — no broker
goroutine leak (AcceptAndServe returns on client Kill/Close), the fixed
broker stream id is safe (nothing in the repo calls `NextId()` or sets
AutoMTLS; each plugin gets its own broker instance), probe spawns start
no host server, namespace isolation holds against built-in BlobStore
kinds, and the size cap / status codes are correct.

Addressed in 0af4b184 (docs/comments only, no logic change):

* **State during Build on first boot.** `loadConfig` runs plugin Build
  callbacks before the blob store is wired (the store lives in the state
  dir, which is part of the config being loaded — can't be cleanly
  reordered). So a Build-time `State().Get` returns Unavailable on first
  boot. None of the target consumers need Build-time state (ssh host
  key, codex jwt, tailscale identity are all runtime callbacks), so
  documented the boundary in `pluginsdk.State`, plugins.md, and the
  wiring comment rather than a risky boot reorder.
* **Per-value vs total quota.** The 1 MiB cap bounds one value, not key
  count. Corrected the comment to say so; a per-namespace quota is a
  follow-up (BlobStore has no count/list to enforce against).
* Documented the namespace = HCL block label and the duplicate-label
  sharing caveat; noted the test global-reset is safe as the sole
  State() test.

Follow-ups noted, not in this PR: a per-namespace storage quota, and
rejecting duplicate plugin block labels at config validation.

PR #691, opened 2026-06-16
Let external credentials rewrite the request URL and body (v2 M2)

piscisaureus:
Second milestone of the external plugin protocol v2 work (see #689),
on top of M1 (#690).

Today an external credential's `InjectHTTP` is header-only — it cannot
rewrite the destination URL or the request body. That blocks
credentials that sign over the body (AWS SigV4) or carry their secret
in the URL/body (telegram, discord). This adds a streaming transform
path for them, while keeping the common header-only case cheap.

* New `http_transform` capability and a bidirectional `TransformHTTP`
  RPC. The gateway streams the request body to the plugin as
  `HTTPBodyChunk` frames; the plugin returns a head (header / method /
  URL mutations) followed by the outgoing body. **Buffering is the
  plugin's choice** — read a prefix and stream the rest, or read it all
  to sign. Header-only `InjectHTTP` is unchanged: the request body never
  flows through the plugin.
* The gateway applies the head **before** forwarding (so a body-derived
  header like a SigV4 signature is finalized first), then pipes the
  plugin's transformed body upstream. Content-Length comes from a header
  mutation the plugin supplies (it knows the new length); otherwise the
  body is forwarded with chunked transfer.
* HTTP **trailers** (e.g. gRPC's `grpc-status`) are carried on the
  body-stream framing, conveyed to the plugin, and pass through to the
  upstream unchanged. Plugin-*rewritten* request trailers are a
  follow-up (rewriting them across the body swap races Go's trailer
  population — called out in the code).
* SDK: `CredentialDef.HTTPTransform` + a `TransformHTTP` callback in
  plain `io.Reader` terms; the streaming is hidden behind it. Set
  `Response.Body = req.Body` to pass the input through with no buffering.

Verified with round-trip tests under the race detector: the gateway
dual-pump (header + full-body rewrite, and URL rewrite + body
pass-through) and the SDK handler (the io.Reader contract).

Unblocks the AWS plugin (SigV4 as a standalone credential) and the
telegram / discord / mtls cases from the audit.

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -2451,7 +2451,22 @@ func (g *Gateway) mitmHTTPSWithCertHost(c net.Conn, host, certHost string, ep *c
 						// Match existing request-signing behavior: an injection failure is logged,
 						// then the request continues with the agent's placeholder. The upstream
 						// service should reject that placeholder without exposing gateway secrets.
+						//
+						// Exception: a transform credential (one that rewrites the URL/body)
+						// consumes the request body during injection, so on failure the request
+						// is corrupted, not merely un-injected — fail closed instead of
+						// forwarding a half-transformed request.
 						if err := injector.InjectHTTP(req.Context(), req, sec); err != nil {
+							if rw, ok := injector.(runtime.HTTPRequestRewriter); ok && rw.RewritesHTTPRequest() {
+								log.Printf("transform %s: %v; failing closed", cc.Credential.Symbol.Name, err)
+								_, _ = fmt.Fprintf(tc, "HTTP/1.1 502 Bad Gateway\r\nContent-Length: 0\r\nConnection: close\r\n\r\n")
+								ev.Status = 502
+								ev.Action = "error"
+								ev.Reason = err.Error()
+								ev.Ms = time.Since(start).Milliseconds()
+								g.emitEnd(ev)
+								return
+							}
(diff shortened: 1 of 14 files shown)


piscisaureus:
Ran a review pass. Design checked out (header-only stays unary, gating
airtight, Content-Length/chunked correct, no deadlock), but it found
three real failure-mode lifecycle blockers — all now fixed in 3091eaa5:

* **No head-arrival deadline / never-canceled stream.** The transform
  path used the bare request context (which is `context.Background()`
  here), so a plugin that never sent its head hung the request forever
  and leaked the up-pump goroutine + stream on every error return. Now
  the stream runs under a cancelable child context with a head-arrival
  timer; every error path cancels, and on success the down-pump owns the
  context and cancels when the body is delivered.
* **Corrupted request forwarded on transform error.** A transform
  credential consumes the body during injection, so the previous
  log-and-forward behavior shipped a drained, half-mutated request
  upstream. Now it fails closed (502), gated by a `runtime.HTTPRequestRewriter`
  marker so header-only credentials keep forwarding the agent placeholder.
* **Trailer data race (should-fix).** The plugin-facing trailer map was
  shared mutable state. Dropped it — trailers are preserved gateway-side
  and no longer routed through the plugin.

Re-verified: build, lint 0 issues, all transform round-trips (gateway +
SDK) under `-race`, plus new tests for the error-fails-closed path and
the rewriter marker.

PR #692, opened 2026-06-16
Prototype: plugin capability bundle (control-plane RPC over the broker)

piscisaureus:
Draft / RFC, now hardened along the reviewed design.

A host-served control plane (`HostControl`) over the same go-plugin
broker the M1 state service uses: each plugin→gateway callback is an
ordinary gRPC method, so gRPC's correlation replaces the hand-rolled
`call_id`/`dial_id` inflight maps multiplexed inside `HandleConn`.

What changed in this revision (vs. the original spike):

* **Direction is the load-bearing decision.** `HostControl` is the
  **plugin→host** capability bundle (`Evaluate` now; brokered `Dial` and
  HITL `Decide` later). The **host→plugin** callbacks — the LLM
  `Approve`, the HITL `Notify`/`Update` — are a plugin-served `Approver`
  service, **not** `HostControl` methods. Only `Decide` is plugin→host.
  The earlier "build M4/M5 as `HostControl` methods" was wrong by
  direction; the doc now has the corrected mapping and an end-to-end M4/M5
  sketch (both directions).
* **Session scoping moved to gRPC metadata + a server interceptor**
  (standard over ad-hoc): the `session_token` message field is gone, so
  request messages stay payload-only and every future method is scoped
  for free. Tokens are minted `crypto/rand` at registration (unforgeable);
  forged / absent / removed are rejected `Unauthenticated`.
* **Typed `Verdict` + session object** mirroring `runtime`/`pluginsdk`
  types, instead of the four-string tuple.

Proven end to end over a real go-plugin broker (the state-service
harness), under `-race`. Still a prototype — served and tested, not yet
wired into the live `HandleConn` path (the deferred migration in the
recommendation).

`doc/plugin-capability-bundle-prototype.md` has the full two-direction
design, the M4/M5 proto sketch, and the phased recommendation.

The decision this asks for: adopt `HostControl` as the plugin→host bundle
and build M4/M5's host→plugin callbacks as a plugin-served `Approver`
service — confirmed?

### DIFF
--- a/doc/plugin-capability-bundle-prototype.md
+++ b/doc/plugin-capability-bundle-prototype.md
@@ -1,171 +1,335 @@
-# Plugin capability bundle — prototype + recommendation
+# Plugin capability bundle — design + prototype
 
 Status: **prototype / RFC.** A working spike lives behind this doc
-(`internal/config/extplugin/hostcontrol.go` + `_test.go`); it is not
-wired into the live request path.
+(`internal/config/extplugin/hostcontrol.go` + `_test.go`).
 
 ## Why
 
 A comparison with `unclaw`'s plugin model (in-process TS) surfaced a real
-smell in clawpatrol's gRPC plugin protocol: it is accreting a **new
-bespoke RPC + a new `*Init` message per capability**, and the
-plugin→gateway *callbacks* are hand-rolled as request/reply frames
-multiplexed inside the `HandleConn` stream with manual correlation.
+smell in clawpatrol's gRPC plugin protocol: it accretes a **new bespoke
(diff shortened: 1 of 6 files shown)


piscisaureus:
Superseded by #701, which is the implementation rebased on current `main` (HostControl wired as the live `Evaluate` path, metadata-scoped sessions, shared evaluate core). The corrected two-direction design + M4/M5 sketch carried over into #701's doc. Leaving this open as the design-discussion thread; close when ready.

piscisaureus:
Landed as #701 (merged to main as ef78aafa). Design: doc/external-plugin-control-plane.md.

PR #694, opened 2026-06-16
fix(auth): preserve Basic authenticate challenges

rajiteh:
## Summary

Fix Git-over-HTTPS Basic Auth flows through the HTTPS MITM gateway.

When an upstream returns `401 Unauthorized`, Git waits for a `WWW-Authenticate: Basic ...` challenge before asking its credential helper for credentials and retrying the request. The gateway currently strips all `WWW-Authenticate` response headers on credentialled MITM responses. That is right for sensitive response-side auth material in general, but it also removes the Basic challenge Git needs to start the retry flow.

The result is that private GitLab clone/fetch requests can fail before the gateway has a chance to rewrite the client's Basic placeholder credential to the real secret.

## Fix

Preserve only clearly valid, single `WWW-Authenticate: Basic ...` challenges on the normal upstream MITM response path.

Everything else remains stripped:
- Bearer/Digest/Negotiate/etc. challenges
- mixed challenge values such as `Basic ..., Bearer ...`
- malformed or ambiguous challenge values
- `Proxy-Authenticate`, `Authentication-Info`, cookies, and trailers
- raw WebSocket/synthetic response paths

The Basic parser is intentionally fail-closed. It accepts only a Basic scheme with auth params including `realm`, normalizes auth params for stdlib `mime.ParseMediaType` validation, and strips anything ambiguous.

## Validation

Added tests covering:
- preserving valid Basic challenges
- stripping Bearer and mixed Basic/Bearer challenges
- malformed Basic challenge values
- quoted commas and escaped quotes in Basic params
- full MITM flow where:
  1. upstream returns 401 with both Bearer and Basic challenges
  2. gateway forwards only the Basic challenge
  3. client retries with Basic placeholder
  4. gateway rewrites the placeholder to the real Basic credential upstream

Local checks:
- `go test ./cmd/clawpatrol -run 'Test.*Basic|TestStripAuthResponse|TestMITMPreservesBasicChallenge' -count=1`
- `go test ./cmd/clawpatrol -count=1`
- `gofmt -l .`
- `git diff --check`

Also verified against a live GitLab private repo clone from an agent runner: the unauthenticated Git smart-HTTP request returned `401` with `WWW-Authenticate: Basic realm="GitLab"`, then Git retried with Basic auth and clone succeeded.

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -2561,7 +2561,7 @@ func (g *Gateway) mitmHTTPSWithCertHost(c net.Conn, host, certHost string, ep *c
 		// before stripping credential-bearing ones — the dashboard
 		// still wants to show what the upstream actually sent.
 		ev.RespHeaders = flatHeaders(resp.Header)
-		stripAuthResponseHeaders(resp.Header)
+		stripAuthResponseHeadersPreservingBasicChallenge(resp.Header)
 		stripAltSvc(resp.Header)
 		// Trailers fall outside resp.Header — Go's http.Transport
 		// surfaces them on resp.Trailer and http.Response.Write
--- a/cmd/clawpatrol/response_sanitize.go
+++ b/cmd/clawpatrol/response_sanitize.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"mime"
 	"net/http"
 	"strings"
 )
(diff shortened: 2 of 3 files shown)


piscisaureus:
Landed on `main` as 647a34bf — thank you, @rajiteh, for the fix and the thorough tests. I rebased your branch onto current `main` and added two small parser-hardening test cases (rejecting CRLF/LF in the Basic `realm` value) before merging; the change is otherwise yours. A security review confirmed the parser is fail-closed against Bearer/Digest/token smuggling and that the preserved Basic challenge can't inject headers or leak a credential.

(Landed via #707 because the repo only allows squash merges from in-repo branches; that's also why this shows as "closed" rather than "merged".)

PR #695, opened 2026-06-17
Make external HTTPS plugin endpoints work end to end

piscisaureus:
Three fixes that an external endpoint plugin which terminates TLS and
serves HTTPS (matched by SNI, including wildcard hosts) needs to work
end to end. Found while bringing up such a plugin.

* **Dispatch SNI-matched plugin conn-endpoints.** `g.handle` peeked
  the SNI and resolved the owning endpoint but only served built-in
  https-mitm facets, splicing everything else through. A plugin
  conn-endpoint bound to a wildcard host was matched yet never handed
  the connection. It now routes a matched `ConnEndpointRuntime`
  through `dispatchConnEndpoint`. No VIP/conn-index entry is needed —
  a wildcard host has neither, which is the whole point.
* **Record plugin egress on first load.** `resolveNetwork`'s `addHash`
  ran before `resolveEgress`, whose fast path then saw the hash and
  returned the empty recorded egress, dropping the manifest-declared
  targets on every fresh lockfile. `resolveEgress` now reads a
  pre-pass snapshot of the lockfile entry. Regression test added.
* **Flush the response before closing.** A one-shot plugin's agent
  connection was hard-closed the moment `HandleConn` returned,
  discarding the response still in the netstack send buffer (client
  hangs with no response). The teardown now half-closes the write side
  and waits, bounded, for the agent to drain before the full close.

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -1490,17 +1490,20 @@ func (g *Gateway) handle(raw net.Conn, dstIP string, dstPort uint16) {
 		g.mitmHTTPSWithCertHost(c, authority, certHost, ep)
 		return
 	}
-	// External plugin conn-endpoints (e.g. an `aws_api` bound to
-	// `*.amazonaws.com`) terminate TLS themselves and pump the raw
-	// agent conn to the subprocess over gRPC, so they aren't an
-	// https-mitm facet. They're matched here purely by SNI: the agent
-	// resolves and dials the real upstream IP and the forwarder
-	// intercepts :443 promiscuously, so no VIP or conn-index entry is
-	// needed — which matters for a wildcard host, since it has
-	// neither (it can't be DNS-resolved to an IP for the conn-index,
-	// and an HTTPS plugin isn't RequiresVIP). Hand the peeked conn to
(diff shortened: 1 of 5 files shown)


piscisaureus:
Addressed a review pass over the diff:

* Narrowed the :443 SNI dispatch to TLS-terminating plugin endpoints via a body marker (`TLSTerminates()`). The prior `ConnEndpointRuntime` check also matched the built-in postgres/clickhouse/ssh runtimes, so one of those bound to a bare or wildcard host and hit on :443 would have been handed a raw TLS ClientHello it can't parse — a regression from the prior splice-through. Added a regression test.
* Gave the dispatched `HandleConn` a cancellable context so a pump goroutine parked on the plugin stream unblocks on teardown instead of leaking.

Known follow-up, pre-existing and left for a separate change: on an upgrade that broadens a plugin's egress, `resolveNetwork`'s `addHash` records the new binary hash before `resolveEgress` fails closed, and the deferred lockfile save still persists that hash — so a later load admits the binary (with the old, narrower egress) instead of staying blocked until re-approved. The clean fix is to defer `addHash` until after the egress check.

PR #697, opened 2026-06-17
Derive async grant TTL from approver and sync-wait timeouts

arnauorriols:
Remove the explicit `async_grant.approval_ttl` config option and derive the async approval TTL as `approver_timeout - sync_wait_timeout`.

## Why
`approval_ttl` was redundant, foot-gun config. The time an async grant should remain pending after the synchronous wait returns a 202 is just whatever is left of the overall approver timeout once the sync window elapses. Configuring it separately let it drift from or contradict the other two values.

## Changes
- Drop the `approval_ttl` field from the `async_grant` schema, its parse/validation code, and the default constant.
- Derive the TTL at the point of use from the approver's overall timeout (its `timeout`, falling back to `human_timeout`, then 10m) minus `sync_wait_timeout`.
- Clamp to zero when `sync_wait_timeout >= approver_timeout` so the grant is born already expired rather than carrying a negative lifetime.
- Stale configs that still set `approval_ttl` are rejected at load time.

## Tests
- Unit tests for the derivation, the policy-timeout fallback, and the clamp edge case.
- Config test asserting `approval_ttl` is rejected.
- Updated existing async config/e2e tests to drop the removed field.

### DIFF
--- a/cmd/clawpatrol/hitl_async_e2e_test.go
+++ b/cmd/clawpatrol/hitl_async_e2e_test.go
@@ -156,7 +156,6 @@ approver "human_approver" "ops" {
   sync_wait_timeout = "1ms"
   async_grant {
     enabled            = true
-    approval_ttl       = "15m"
     approved_retry_ttl = "5m"
     fingerprint_body   = "raw"
     max_body_bytes     = 1048576
--- a/cmd/clawpatrol/hitl_async_runtime.go
+++ b/cmd/clawpatrol/hitl_async_runtime.go
@@ -29,7 +29,7 @@ const (
 type hitlAsyncGrantRuntime interface {
 	HITLAsyncGrantEnabled() bool
 	HITLSyncWaitTimeout() time.Duration
-	HITLAsyncApprovalTTL() time.Duration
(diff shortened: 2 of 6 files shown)


arnauorriols:
[bot-reply]
No code change needed — the PR body is descriptive and accurately matches what's already on the branch (commit f15d2c1c). Verified against the implementation:

- `approval_ttl` field dropped from the `async_grant` schema, its parse/validation, and the `HITLAsyncDefaultApprovalTTL` constant.
- TTL derived at point of use as `approvalTimeout(policy) - HITLSyncWaitTimeout()` in `human.go`, threaded through `HITLAsyncApprovalTTL(policy)` in `hitl_async_runtime.go`.
- Clamp-to-zero when `sync_wait_timeout >= approver_timeout`.
- Stale configs setting `approval_ttl` are rejected at load (`TestHITLAsyncConfigRejectsRemovedApprovalTTL`).
- Derivation, policy-timeout fallback, and clamp covered by unit tests; config/e2e tests updated.

`go test ./internal/config/... ./internal/config/plugins/approvers/...` passes locally and all CI checks are green. Nothing to address.

Issue #705, opened 2026-06-18
extplugin: tunnel via-chaining for plugin tunnels (deferred)


piscisaureus:
Deferred — there are currently **no external tunnel plugins** in-tree
(only built-in tunnels: tailscale, ssh_port_forward,
kubernetes_port_forward, local_command), so this has no consumer yet.
Capturing the design so it's ready when a tunnel plugin lands.

## Problem

A built-in tunnel can route through another (`via = <tunnel>`): the
Manager recursively acquires the parent and chains the dial. A plugin
tunnel's `Dial` has no way to say "go through that other tunnel."

## Proposal (additive)

Add `via_tunnel_handle` to `OpenTunnelRequest` / `DialInit`. The gateway
resolves the parent tunnel (built-in or plugin) and routes the plugin
tunnel's upstream through it — the gateway owns the composition, the
plugin just names the parent. Because dials are gateway-brokered, this
stays transparent to the plugin: it never holds the parent transport.

## Why deferred

As a bare proto field it has no consumer until an external tunnel plugin
exists that needs chaining (e.g. ssh_port_forward used as a bastion).
Build it then, against a real plugin.

piscisaureus:
Done — plugin tunnels can now chain through other tunnels. The deferral
reason ("no external tunnel plugin in-tree yet") is resolved:

* #714 — brokered transport dial: `OpenTunnelRequest.transport_dial_handle`
  + a per-plugin route registry that resolves the parent tunnel (built-in
  or plugin) and routes the plugin tunnel's transport through it. The
  gateway owns the composition; the plugin just names the parent and never
  holds the parent transport — exactly the additive design proposed here
  (`via_tunnel_handle` → `transport_dial_handle`). Carries UDP too.
* #719 — `example_socks`, the first external tunnel plugin (the missing
  consumer).
* #722 — `via = <tunnel>` honored on plugin tunnel blocks in HCL; the e2e
  chains two plugin tunnels (`example_passthrough via example_socks`) end
  to end through the brokered transport dial.

PR #713, opened 2026-06-18
spike: plugin tunnel via-chaining + UDP (wireguard + socks)

piscisaureus:
**Spike / RFC — not for merge as-is.** Explores `via`-chaining for
external plugin tunnels (the deferred #705), with the wrinkle that
WireGuard is UDP and "via a SOCKS proxy" means carrying UDP over a
TCP-ish tunnel. Two example plugins are the real second consumer that
justifies building the interface.

## The interface

* `OpenTunnelRequest.via_tunnel_handle` + a gateway-served
  `HostTunnel.DialVia` stream (reuses `DialMessage`). When a plugin
  tunnel is chained, the gateway registers the already-acquired parent
  under an opaque token and hands it to the plugin; the plugin routes
  its transport through the parent by echoing the token to `DialVia`.
  The gateway resolves the parent (built-in or plugin) and dials
  through it — it owns the composition, the plugin just names the
  parent. This plugs into the existing `TunnelManager.Acquire`
  via-chain machinery; `tunnelAdapter.Open` previously dropped the
  `via` arg.
* SDK: `TunnelOpenRequest.Via` dialer (`req.Via.Dial(ctx, "udp",
  endpoint)`), nil when not chained.

## The UDP answer

No datagram proto. The tunnel `Dial` path already forwards a `network`
string, so `network="udp"` + a shared length-prefixed datagram framing
(`pluginsdk.WriteDatagram`/`ReadDatagram`, `PacketConnOverStream`) lets
the existing byte stream carry datagrams; the parent reframes them onto
a real datagram transport.

## The plugins

* **socks** — hand-rolled SOCKS5. `tcp` = CONNECT; `udp` = UDP
  ASSOCIATE, reframing the datagram-framed upstream onto the proxy's
  UDP relay. Works as a `via` parent.
* **wireguard** — wireguard-go + gVisor netstack
  (`netstack.CreateNetTUN`); `Dial` is TCP over the WG netstack. In
  `via` mode it points the WG device's normal UDP bind at a local relay
  socket bridged to `req.Via.Dial("udp", endpoint)`, so WG's handshake
  UDP rides the parent with no custom `conn.Bind`.

Enables:

```hcl
tunnel "socks_proxy" "corp" { proxy = "socks.corp:1080" }
tunnel "wireguard" "vpn" {
  endpoint = "wg.corp:51820"  address = "10.8.0.2"
  private_key = "…"  peer_public_key = "…"
  via = socks_proxy.corp
}
endpoint "https" "api" { hosts = ["api.internal"]  tunnel = wireguard.vpn }
```

## State

Builds (darwin + linux), gofmt + golangci-lint clean, no regressions.
Tested: datagram round-trip; full `DialVia` via round-trip over a real
go-plugin broker (parent receives the child's `udp`/endpoint request,
unknown token refused).

Not done: the live WG-over-SOCKS datapath against real peers; built-in
tunnels as udp `via` parents; MTU/fragmentation; the broker double-hop
for via-UDP is a spike simplification.

### DIFF
--- a/cmd/clawpatrol/tunnelvia_e2e_test.go
+++ b/cmd/clawpatrol/tunnelvia_e2e_test.go
@@ -1,0 +1,187 @@
+//go:build linux
+
+package main
+
+// Live end-to-end driver for the plugin-tunnel via+UDP spike. Gated by
+// CLAWPATROL_VIA_E2E so it never runs in normal CI. It builds the real
+// socks + wireguard plugin binaries, loads a config with a wireguard
+// tunnel (optionally `via` a socks tunnel), acquires it through the
+// gateway's TunnelManager, dials a target over the WG netstack, and
+// checks the HTTP response.
+//
+// Run (against a live WireGuard server + SOCKS proxy + target):
+//
+//	CLAWPATROL_VIA_E2E=1 VIA=1 \
+//	WG_ENDPOINT=host:51820 WG_PRIV=... WG_PEERPUB=... WG_ADDR=10.9.0.2 \
+//	SOCKS_PROXY=host:1080 TARGET=10.9.0.1:8080 \
+//	go test ./cmd/clawpatrol -run TestTunnelViaE2E -count=1 -v
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/denoland/clawpatrol/internal/config"
+	"github.com/denoland/clawpatrol/internal/config/extplugin"
+	"github.com/denoland/clawpatrol/internal/config/runtime"
+)
+
+func envOr(k, def string) string {
+	if v := os.Getenv(k); v != "" {
+		return v
+	}
+	return def
+}
+
+func buildTunnelPlugin(t *testing.T, pkg, name string) string {
+	t.Helper()
+	out := filepath.Join(t.TempDir(), name)
+	root, err := filepath.Abs("../..")
+	if err != nil {
+		t.Fatal(err)
+	}
+	cmd := exec.Command("go", "build", "-o", out, pkg)
+	cmd.Dir = root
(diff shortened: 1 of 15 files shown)


piscisaureus:
**Validated live, end to end.** Topology: a WireGuard server + SOCKS5
UDP-ASSOCIATE proxy + an HTTP target (bound to the WG address, reachable
only through the tunnel) on one host; the clawpatrol gateway + both
plugins on another (separate networks, real UDP over the public
internet).

- **WG-direct** (no via): the wireguard plugin establishes WireGuard
  over real UDP and fetches the target through the WG netstack — `200`,
  target body returned.
- **WG-over-SOCKS** (`via = socks`): WireGuard's handshake UDP is carried
  over the SOCKS5 UDP-ASSOCIATE relay via the new `DialVia` conduit, the
  tunnel comes up, and the same fetch returns `200` with the target body.

Driven through the **real** `TunnelManager.Acquire` → plugin
`OpenTunnel`/`Dial` path (env-gated `TestTunnelViaE2E`), not a stub.

Two real bugs surfaced and are fixed in 9beada04:
1. `dynamicTunnelBody` didn't implement `runtime.TunnelRuntime`, so a
   plugin tunnel couldn't be acquired at all — never hit before because
   no external tunnel plugin existed.
2. The via conduit used the `OpenTunnel` RPC ctx, which is cancelled when
   `Open` returns, tearing the `DialVia` stream down immediately. Now
   tied to the tunnel lifetime.

Plus a SOCKS UDP-ASSOCIATE BND.ADDR-behind-NAT fallback.

Still open (config-layer, noted in the commit): peeling the HCL `via`
attribute on a *plugin* tunnel block; the e2e wires `Via` on the
compiled tunnel directly.

piscisaureus:
Split into two stacked PRs: **#714** (the landable `via` + UDP tunnel interface + the `dynamicTunnelBody` bugfix) and **#715** (the example wireguard + socks plugins + live e2e, stacked on #714, spike-only). Closing this combined one.

PR #721, opened 2026-06-18
docs: plugin tunnel transport dial + plugin-docs audit corrections

piscisaureus:
A docs audit of the recent plugin work (sandbox, GitHub distribution,
plugin interfaces) against the current code, plus the corrections and
additions it surfaced. The sandbox and distribution docs were already
accurate; the gaps were in the tunnel/interface area, where tunnel
plugins still being described as needing `network = "outbound"` was the
load-bearing stale claim.

## New coverage

* **plugins.md** — new "Tunnel transport dial" section: a tunnel plugin
  opens its own transport through the gateway's brokered `DialUpstream`
  (tcp + udp, with datagram framing), runs `network = "none"`, and the
  gateway routes it direct or through a `via` parent. Documents
  `example_socks`.
* **tunnels.md** — new "Plugin tunnel types" section.
* **cli.md** — the `clawpatrol plugins install|update|lock|info|approve`
  command group (was entirely absent from the CLI reference).
* **external-plugin-control-plane.md** (internal) — `HostTunnel` as the
  third host-served broker service; the broker-multiplex-over-main-
  connection note; reconciled the old `HostControl.Dial` sketch with the
  shipped `HostTunnel.DialUpstream`.

## Corrections (stale "tunnel plugins need outbound" claims)

* `plugins.md` `network` prose + the `example_passthrough`/"See also"
  descriptions (they said tunnels dial their own socket)
* `security-model.md`
* `config.go` `PluginSource.Network` comment

config-reference.md is unchanged — docgen doesn't render the
`PluginSource` sub-field comments, so no regen was needed.

## Note on `via`

This PR currently carries a caveat that `via` written on a **plugin**
tunnel block "isn't wired yet." #722 wires it. Once #722 lands, that
caveat is dropped here (in `tunnels.md` + the plugins.md "Chaining"
paragraph) so these docs describe shipped behavior with no asterisk.

PR #722, opened 2026-06-18
extplugin: honor via/share/keepalive/credential on plugin tunnel blocks

piscisaureus:
Makes the `via` block (and the other common tunnel attrs) work on
**external** tunnel blocks. Previously `tunnel "wireguard" "vpn" { via =
socks }` was silently a no-op: `dynamicTunnelBody` didn't implement
`TunnelCommonRead`, and the loader's `frameworkAttrsByKind` extraction
covers only endpoints/credentials, not tunnels — so a plugin tunnel's
`via`/`share`/`keepalive`/`credential` were never parsed. The
via-chaining spike (#720) worked around this by setting
`CompiledTunnel.Via` in Go.

The plugin tunnel `DecodeBody` now peels these attrs (`decodeTunnelCommon`)
before the manifest-schema decode, and `dynamicTunnelBody` exposes them
through `TunnelCommon`, so the compile pass resolves `via`/`credential` by
name and detects cycles — the same path a built-in tunnel takes. Built-in
tunnels, compile, and HCL emit are untouched.

Tests:
* `via` on a plugin tunnel resolves the compiled `Via` pointer
* an undeclared `via` target is rejected
* an end-to-end dial through `example_passthrough via example_socks` —
  two plugin tunnels chained entirely in HCL, sandbox on — returns the
  target's response (the thing the spike could only do with a Go hook)

Follow-up: once this lands, the docs PR (#721) should drop its
"`via` on a plugin tunnel block isn't wired yet" caveat in `tunnels.md`
and `plugins.md`.

### DIFF
--- a/cmd/clawpatrol/plugin_tunnel_via_test.go
+++ b/cmd/clawpatrol/plugin_tunnel_via_test.go
@@ -38,7 +38,9 @@ gateway {
   wireguard { subnet_cidr = "10.55.0.0/24" }
 }
 tunnel "example_socks" "corp" {
-  proxy = "10.0.0.9:1080"
+  proxy     = "10.0.0.9:1080"
+  share     = "per_conn"
+  keepalive = "5m"
 }
 tunnel "example_passthrough" "chained" {
   via = example_socks.corp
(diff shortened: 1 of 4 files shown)


piscisaureus:
Addressed the review's should-fix: a plugin tunnel manifest declaring a
framework-reserved attr (`via`/`share`/`keepalive`/`credential`) is now
rejected at registration with a clear diagnostic — mirroring how
`endpointSpec` rejects `hosts`/`dial` — so the operator's value can't be
silently stolen by the peel. Added a `TestTunnelReservedAttrRejected`
table test (all four names) and locked in the nit by asserting
`share`/`keepalive` land on the compiled tunnel in `TestPluginTunnelViaCompiles`.

The reviewer confirmed `credential` is genuinely threaded to the plugin at
Open (not a half-feature), built-in tunnels are untouched, and the e2e
genuinely routes through the `via` parent. Build/lint/tests green.

PR #724, opened 2026-06-18
extplugin: CLAWPATROL_PLUGIN_LOG to surface plugin subprocess logs

piscisaureus:
Adds a `CLAWPATROL_PLUGIN_LOG` env var to raise the plugin-subprocess log
level (trace/debug/info/warn/error) above the default Info.

At trace/debug it surfaces the plugin's own stderr — otherwise invisible
when a sandboxed plugin exits before completing the go-plugin handshake,
where the only symptom is `Unrecognized remote plugin message` with no
cause. When no gateway log sink is wired (e.g. `clawpatrol validate`),
output goes to stderr rather than being discarded.

Default behavior is unchanged when the variable is unset.

### DIFF
--- a/internal/config/extplugin/manager.go
+++ b/internal/config/extplugin/manager.go
@@ -89,18 +89,20 @@ func New(out *log.Logger) *Manager {
 	level := hclog.Info
 	var output io.Writer = hclogWriter{out}
 	// CLAWPATROL_PLUGIN_LOG raises the plugin-subprocess log level
-	// (trace/debug/info/warn/error). At trace/debug this surfaces the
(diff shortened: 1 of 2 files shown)


piscisaureus:
Addressed the review's two should-fix items: an unrecognized `CLAWPATROL_PLUGIN_LOG` value is now ignored (level stays Info) with a one-line stderr warning naming the accepted set, instead of silently jumping to max (trace) verbosity on a typo; and the variable is documented in the cli.md environment-variable table. Also noted `off` in the comment's accepted-values list. Build/fmt/vet/tests green.

PR #726, opened 2026-06-19
extplugin: don't block plugins update to a newer attested version

piscisaureus:
`clawpatrol plugins update` to a newer **attested** plugin release was
wrongly blocked. `checkProvenanceNotDowngraded` (fetch.go) flagged any
change in the attested source commit as a re-pointed tag and failed
closed — but across a version bump the commit always changes. The error
even told the operator to "run `clawpatrol plugins update`", which is
exactly what was failing (circular); the real remedy is `clawpatrol
plugins approve`.

A re-pointed tag means the **same** tag now vouches a **different**
commit. So the guard now treats a commit change as a re-point only when
the fetched version equals the locked version. An explicit upgrade fetches
a newer tag, whose new commit is legitimate and is re-pinned by the caller
(the attestation itself is still verified at fetch time, independently).
The load and keep-pinned paths fetch the locked version, so they are
unchanged and still catch a genuine re-point. The error message now points
at `plugins approve`.

Hit live while upgrading the production gateway's aws plugin v0.3.2 →
v0.3.3; `plugins approve` (which re-resolves + accepts) was the working
path.

* test: same-version commit change still blocks; a newer version with a
  different commit is accepted; a lost attestation still blocks regardless
  of version; provenance = "off" skips the checks.

### DIFF
--- a/internal/config/extplugin/fetch.go
+++ b/internal/config/extplugin/fetch.go
@@ -530,7 +530,7 @@ func (m *Manager) resolvePluginBinary(ctx context.Context, sp config.PluginSourc
 			return "", nil, fmt.Errorf(
 				"downloaded binary hash %s is not in the lockfile (expected one of %v); refusing", res.binSHA, entry.Hashes)
 		}
-		if err := checkProvenanceNotDowngraded(sp.Name, mode, entry, res); err != nil {
+		if err := checkProvenanceNotDowngraded(sp.Name, mode, entry, res, r.TagName); err != nil {
 			_ = os.Remove(res.path)
 			return "", nil, err
 		}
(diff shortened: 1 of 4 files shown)


piscisaureus:
Addressed the review's nit: added `TestInstallUpgradeAcceptsNewVersionCommit`, an end-to-end `Install(upgrade=true)` test (two-release server, pin v1.2.0/commit-aaa → upgrade to v1.3.0/commit-bbb, assert re-pin) that drives the regression through the real fetch stack — it would have failed before the fix. The unit-level `TestProvenanceCommitCheckIsVersionAware` and the unchanged end-to-end same-version block (`TestProvenanceRecordsAndPinsCommit`) remain.

PR #727, opened 2026-06-19
Add clawpatrol.device.{name,profile,owner} to GenAI spans

arnauorriols:
Attach the calling device's identity to every GenAI (`gen_ai.*`) span emitted for an intercepted LLM request:

- `clawpatrol.device.name` — the device name
- `clawpatrol.device.profile` — its bound profile
- `clawpatrol.device.owner` — the device/profile owner

Values resolve from the connection peer IP through the existing onboard registry / `profileFor` path that the discovery endpoint and dispatch already use, so identity comes from the tunnel, never from request tokens or headers.

Unknown devices degrade gracefully: `name` and `owner` are omitted rather than guessed, and `profile` falls back to the default profile that governs unclaimed peers. The standard `gen_ai.*` attributes are unchanged.

### DIFF
--- a/cmd/clawpatrol/genai.go
+++ b/cmd/clawpatrol/genai.go
@@ -114,6 +114,16 @@ type genAITurn struct {
 	ServerAddress string
 	ServerPort    int
 
+	// Device identity of the calling device (clawpatrol.device.*),
+	// resolved from the connection peer IP via the same device/profile
+	// path the discovery endpoint and dispatch use — never tokens or
+	// headers. Empty fields are omitted from the span: name and owner are
+	// empty for an unknown device; profile falls back to the default
+	// profile that governs unclaimed peers.
+	DeviceName    string // clawpatrol.device.name
+	DeviceProfile string // clawpatrol.device.profile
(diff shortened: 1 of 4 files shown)


arnauorriols:
[bot-reply]
Reviewed the PR description against the diff — it accurately reflects the implementation, no code changes needed:

- `clawpatrol.device.{name,profile,owner}` are added to `genAITurn` and emitted by `emitGenAISpan` only when non-empty.
- `fillDeviceAttrs` resolves identity from the calling device's peer IP (`agentIP`), which `trackLLMUsage` derives via `g.agentIPFor(c)` — from the tunnel connection, never request tokens/headers, as stated.
- `profile` uses `g.profileFor(ip)`, which falls back to `defaultProfileName(...)` for unclaimed peers — matching "falls back to the default profile that governs unclaimed peers."
- Unknown devices: `name`/`owner` stay empty and are omitted from the span rather than guessed; standard `gen_ai.*` attributes are unchanged.
- Test coverage (`TestRecordGenAITurnDeviceAttributes`) exercises both the known-device and unknown-device paths.

No change request or open question in the body, so no follow-up commit.

PR #731, opened 2026-06-19
Document async HITL approval in the agent manifest

arnauorriols:
When a gated endpoint is configured with a synchronous wait window, an unapproved request returns a 202 and continues asynchronously instead of holding the connection open. The discovery manifest previously documented only the synchronous parking flow, leaving an agent with no guidance on how to interpret a 202 or poll for the eventual decision.

## What changed

The manifest now carries an async-approval section (markdown + JSON, rendered from one representation) that documents:

- **The 202-then-poll-then-retry protocol** with the exact response fields the gateway serves: `operation_id`, `status_url` (poll this; don't construct your own), `state`, `terminal`, `retry_original_request`, and the `Clawpatrol-HITL-Operation` retry header. A 202 is neither success nor failure — it means "parked for human approval, continuing asynchronously."
- **Polling semantics** for each state: `sync_waiting`/`pending_approval` (keep polling), `approved_waiting_for_retry` (re-send the original request with the retry header before `retry_expires_at`), `denied`/`expired` (stop). The pending-vs-expired distinction is called out so an agent keeps polling a pending operation rather than giving up early.
- **Per profile, which endpoints fall back to async approval and after what time** — for each, the synchronous wait window before the 202 (`sync_wait_timeout`) and how long the parked operation then stays pollable (approver timeout minus the sync wait).

The section is emitted only when an endpoint can actually go async — the profile opts into async grants and the gating approver is async-capable with a positive `sync_wait_timeout`. A purely synchronous deployment, a dashboard approver, or an async-capable approver with no sync window stays out of the async list, and the existing synchronous HITL docs are unchanged.

## Tests

- New `hitl_async.hcl` golden fixture exercising an async endpoint alongside sync-only and ungated ones; markdown + JSON goldens.
- Unit tests asserting the async timing (1m30s sync wait, 8m30s poll TTL), that sync-only endpoints are excluded from the async list while still appearing in the gated set, the served field names/constants, and that the async block is absent without async config.


arnauorriols:
[bot-reply]
Reviewed the PR body. It is a descriptive summary of the change (the async-HITL 202-then-poll manifest docs), not a change request or a question directed at the author, and there are no outstanding review comments on the PR. No code change is required and no clarification is needed, so nothing was modified. If a specific edit to the manifest docs was intended here, please reply with the requested change and a polecat will pick it up on the next cycle.

cmd/clawpatrol/testdata/discovery/hitl_async.md
```diff
@@ -0,0 +1,98 @@
...
+### Asynchronous approval (202 + polling)
+
+Some gated endpoints do not hold your connection forever. Each such endpoint has a synchronous wait window (its sync_wait_timeout, listed per endpoint below): if a human has not decided within that window, the gateway stops holding the connection and answers with HTTP 202 Accepted. A 202 is NOT success and NOT failure — it means "parked for human approval, continuing asynchronously." Do not treat it as either; switch to polling.
+
+The 202 body is JSON describing the parked operation. The fields you act on:
+- operation_id: the parked operation's id.
+- status_url: the absolute URL to poll for this operation's status (also returned in the Location header). Poll THIS url; do not build your own — the template is /api/hitl/operations/{operation_id}/status.
+- state: the current state (see below).
+- terminal: true once the operation has reached a final state and will not change.
+- retry_original_request: true while you should still be working the operation (poll, or retry once approved).
```
arnauorriols: This phrase is confusing. What does "you should still be working the operation"? poll or retry once approved are opposite operations that should not be decided by the agents. This field should only be true when the original operation should be retried. We should have an extra field `poll_operation_status: bool` that returns true while the operation status should be polled

arnauorriols:
[bot-reply]
Pushed 85d2d745 addressing this comment:
- Split the conflated signal into two mutually-exclusive booleans. `retry_original_request` is now emitted **only** in `approved_waiting_for_retry` (the agent re-sends the original request to execute it).
- Added `poll_operation_status: true`, emitted while the operation is parked waiting on a human (`sync_waiting` / `pending_approval`) — i.e. the "keep polling" signal.
- Updated the async manifest prose (`discovery.go` + regenerated `hitl_async.{md,json}` goldens) to document both fields and note they are mutually exclusive.
- Updated `hitl_operation_api_test.go`: the 202/pending and sync-waiting responses now assert `poll_operation_status: true` with `retry_original_request` absent; `approved_waiting_for_retry` keeps the retry envelope.

PR #732, opened 2026-06-19
facet: title/description/detail_only fields + AWS-friendly activity log

piscisaureus:
Makes the activity log read sensibly for plugin endpoints like AWS, where
the useful identifier is the IAM/API action, not the HTTP method.

A facet field gains three optional attributes, plumbed proto → SDK → facet
registry → `/api/facets` → dashboard:

* **`title`** — the field whose value is the action's primary identifier.
  The activity-log verb becomes this instead of the HTTP method, so an AWS
  plugin marking `iam_action` makes a row read `s3:ListBucket`, not `POST`.
* **`description`** — a human explanation shown in the per-action facet
  table (e.g. "API action (CloudTrail)").
* **`detail_only`** — keep a field out of the compact row (still in detail).

Dashboard:

* verb = the title field (falling back to today's method/verb heuristic),
  and the inline body drops `detail_only` fields.
* the per-action facet block is now a **three-column table** —
  fully-qualified facet name (`aws.action`, what you write in a rule),
  description, value — so it's obvious what's matchable.
* allowed brokered `dial` rows (a plugin endpoint's gateway→upstream dial,
  no metadata) are hidden by default behind a "show N dials" toggle;
  **denied** dials (a blocked egress) always show.

Defaults preserve today's behavior for every existing facet; the AWS
plugin populates the new attributes in a companion release.

### DIFF
--- a/cmd/clawpatrol/web.go
+++ b/cmd/clawpatrol/web.go
@@ -2287,9 +2287,12 @@ func (w *webMux) apiAnalytics(
 func (w *webMux) apiFacets(rw http.ResponseWriter, r *http.Request) {
 	_ = r
 	type reportFieldJSON struct {
-		Name  string `json:"name"`
-		Kind  string `json:"kind"`
-		Label string `json:"label,omitempty"`
+		Name        string `json:"name"`
+		Kind        string `json:"kind"`
+		Label       string `json:"label,omitempty"`
+		Description string `json:"description,omitempty"`
+		Title       bool   `json:"title,omitempty"`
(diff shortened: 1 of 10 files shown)


piscisaureus:
Addressed the review's should-fix: the per-action detail-page header used a separate `headerFromFacets` that didn't honor `title`/`detail_only` (so an AWS row would show `s3:ListBucket` in the log but `POST` in the detail header). Exported `rowDescriptors` from LiveRequests and reused it for the detail header, deleting the duplicate — both views now share one rule. Build/lint green.

PR #734, opened 2026-06-19
status: free-form string + plugin-reported outcomes via ActionResult

piscisaureus:
Lets a plugin endpoint report its action's outcome after the response —
the after-the-fact counterpart of `EvaluateAction` — so an AWS row shows a
real status (the HTTP code, or the AWS error code on failure) instead of
nothing. The response side mirrors the request side: a facet declares
`result_fields` (reusing the same `FacetField`, so `title`/`description`/
`detail_only` and the `FacetStream` kind apply), the plugin reports via
`Conn.SetResult`, and the result field marked `title` becomes the status.

* **status is a string** end to end (HTTP renders `"200"`/`"403"`; a plugin
  reports `"AccessDenied"`). `statusClass` buckets numeric by leading digit
  and a named status as an error; no DB migration.
* **proto** `ActionResult{call_id, result_json, streams}` mirrors
  `EvaluateAction`; `FacetDecl.result_fields`.
* **pluginsdk** `FacetDef.ResultFields` + `Conn.SetResult(ctx, map)`.
* **gateway** a plugin action takes the built-in HTTP start→end lifecycle:
  emitEvaluation emits an in-flight `start` on allow, the adapter emits the
  `end` (status lifted from the result-title field) on `ActionResult`, and
  a flush-on-close guarantees every started action persists exactly once.
  Correlation is conn-scoped (the common `HostControl` evaluate path has no
  call_id), so the result finalizes the connection's current action
  (sequential request/response).

The AWS plugin populates the new attributes + reports status in a
companion release. Follow-ups: a `FacetStream` result field (a response
body the gateway streams + caps), rendering non-status result facets in
the detail table, and a lifecycle integration test.

### DIFF
--- a/cmd/clawpatrol/dev_seed.go
+++ b/cmd/clawpatrol/dev_seed.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"log"
 	"math/rand"
+	"strconv"
 	"sync"
 	"time"
 
@@ -463,7 +464,7 @@ func devSeedAction(r *rand.Rand, devices []devSeedDevice, ts time.Time) Event {
 		Host:     ep.host,
 		Method:   method,
 		Path:     path,
-		Status:   status,
+		Status:   strconv.Itoa(status),
 		In:       int64(50 + r.Intn(10000)),
 		Out:      int64(50 + r.Intn(50000)),
 		Ms:       devSeedLatency(r),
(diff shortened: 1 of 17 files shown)


piscisaureus:
Addressed the review. **Blocker (analytics):** the error count used `status >= 400`, which miscounts now that status is TEXT (SQLite sorts text above all numbers, so ""/"OK"/named statuses all matched). Replaced with a predicate mirroring `statusClass`: non-empty AND (non-numeric OR numeric >= 400) — so named errors count, empty/2xx/3xx don't. **Should-fix:** fixed the `-tags dev` dev-seed build (Status is a string). **Nit:** corrected the `ActionResult.call_id` comment — v1 is conn-scoped (call_id reserved for future per-call correlation). The lifecycle/locking/proto/SDK were confirmed correct. Build incl `-tags dev` + tests green.

PR #744, opened 2026-06-21
fix(ws): preserve explicit upstream ports

rajiteh:
## Summary

`kubectl exec` through Clawpatrol fails for a Kubernetes endpoint whose configured server is an authority with an explicit port, for example:

```hcl
endpoint "kubernetes" "k8s" {
  server = "10.1.1.100:6443"
}
```

The gateway logs show the WebSocket bridge treating the full `host:port` authority as a DNS hostname:

```text
ws-upgrade 10.1.1.100:6443 /api/v1/namespaces/agents/pods/agent-runner-s6-0/exec
ws dial 10.1.1.100:6443: dial tcp: lookup 10.1.1.100:6443: no such host
```

Client-side, `kubectl exec` returns:

```text
Error from server (BadRequest): Upgrade request required
```

## Cause

`dialWSUpstream` assumes its `upstream` argument is a bare host and always constructs the dial address with:

```go
addr := net.JoinHostPort(upstream, "443")
```

For `upstream = "10.1.1.100:6443"`, the WebSocket path should preserve the configured port and use only the host portion for TLS server name handling. Instead, the current path can pass `10.1.1.100:6443` as the hostname to a dial path, producing a DNS lookup for `10.1.1.100:6443`.

The normal HTTPS path handles host/port authorities more carefully, so this appears isolated to the raw WebSocket bridge used for upgrade requests.

## Fix

- parse WebSocket upstream authorities before dialing
- preserve configured `host:port` values for Kubernetes exec/attach WebSocket upgrades
- keep TLS `ServerName` set to the bare host and cover IPv4/IPv6 cases in tests
- apply endpoint-level upstream TLS configuration to the browser-fingerprint WebSocket TLS path so Kubernetes endpoint CA roots are honored

## Tests

- `gofmt -l .`
- `go test ./cmd/clawpatrol -run 'Test(WSUpstreamAddrAndServerName|BrowserTLSUsesEndpointTLSConfig|TransportBrowserTLSUsesEndpointTunnel|DialWSUpstreamBrowserTLSUsesEndpointTunnel|TransportBrowserTLSWithClientCertUsesDialUpstream|DialWSUpstreamWithClientCertUsesDialUpstream)$' -count=1`
- `GOFLAGS=-buildvcs=false go test ./cmd/clawpatrol`

### DIFF
--- a/cmd/clawpatrol/dial.go
+++ b/cmd/clawpatrol/dial.go
@@ -85,13 +85,7 @@ func (g *Gateway) servePorts() {}
 func (g *Gateway) dialUpstream(ctx context.Context, network, addr, serverName string, ep *config.CompiledEndpoint, _ string) (net.Conn, error) {
 	cfg := &tls.Config{ServerName: serverName, NextProtos: []string{"http/1.1"}}
 
-	if ep != nil && ep.Body != nil {
-		if cfgr, ok := ep.Body.(runtime.EndpointTLSConfigurer); ok {
-			if err := cfgr.ConfigureUpstreamTLS(cfg); err != nil {
-				log.Printf("upstream tls %s: %v — dialing with default roots", serverName, err)
-			}
-		}
-	}
+	configureEndpointTLS(cfg, serverName, ep)
 
 	if ep != nil {
 		for _, ent := range ep.Credentials {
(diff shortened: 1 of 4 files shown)


piscisaureus:
Thanks @rajiteh — great fix and nicely tested. This landed on `main` as 544b905d549bcff903153c70e7ee428dc4d0496f with your authorship preserved.

I added one small follow-up: a comment in `dialBrowserTLS` documenting that only ServerName/RootCAs/InsecureSkipVerify cross from `tls.Config` into `utls.Config` (and what to extend if a future EndpointTLSConfigurer needs more). Otherwise it's your change as submitted. CI was green across lint/test/sandbox jobs, and an independent review confirmed the IPv6 and IP-literal SNI cert-verification paths hold up.

Closing this out since it's now on main.

PR #749, opened 2026-06-26
extplugin: add --plugin-cache-dir for validate/test

magurotuna:
## Problem

`clawpatrol validate`/`test` start external plugins to schema-check them, caching the binary under `<state_dir>/plugins`. When the config's `state_dir` is the gateway's production path (e.g. `/var/lib/clawpatrol`), linting it on another machine (e.g. `deploy.sh` on an operator's laptop) fails:

```
Plugin "aws" failed to start; ... mkdir /var/lib/clawpatrol: permission denied
```

`state_dir` is the gateway's secret/persistent store (and the sandbox `read_paths`-overlap boundary). The verification commands only need *somewhere* to cache a plugin binary — that location is incidental to config validity.

## Change

- Give the plugin cache a root distinct from `state_dir`, defaulting to `state_dir` so the **gateway run and `plugins install|update|lock` are unchanged**.
- `validate`/`test` expose a **`--plugin-cache-dir DIR`** flag (→ `SetCacheDir`). Default (no flag) stays **strict**: the cache uses the config's `state_dir`, and an unwritable one surfaces as a real error (and a read-only state_dir with an already-pinned binary still serves a cache hit — no probe, no forced refetch). A caller that's linting on a foreign machine passes a scratch dir.
- `validate`/`test` verify build-provenance attestations (`VerifyProvenance(true)`, via the shared verification-manager helper) so a cold fetch reproduces the gateway's load semantics (attestation) instead of tripping the lockfile's `attested=true` downgrade guard.

### DIFF
--- a/cmd/clawpatrol/test.go
+++ b/cmd/clawpatrol/test.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 
 	"github.com/denoland/clawpatrol/internal/config"
-	"github.com/denoland/clawpatrol/internal/config/extplugin"
 	"github.com/denoland/clawpatrol/internal/config/runtime"
 )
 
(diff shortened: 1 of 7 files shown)


magurotuna:
Reworked to address both findings by switching from the implicit "graceful fallback" approach to an **explicit, strict-by-default** one:

- **Finding 1 (read-only cache skipped → forced refetch):** removed entirely. There's no writability probe anymore — the cache root defaults to `state_dir` and is handed straight to the fetcher, so a read-only state_dir with an already-pinned binary takes the normal cache-hit fast path. A caller that needs a writable cache elsewhere passes `--plugin-cache-dir`; `deploy.sh` does this with a `mktemp -d`.

- **Finding 2 (`test` lockfile → spurious escalation):** reverted. `test` no longer installs a lockfile, so it's back to clean first-use and can't emit a permission-escalation error from a partial in-memory entry.

Net: `manager.go` is now +33 lines (a `cacheDir` field, getter, setter); the fallback/temp/probe/memoization machinery is gone. Thanks for the careful read.

Issue #709, opened 2026-06-18
RFC: Dynamic peers for ephemeral Kubernetes agent pods


rajiteh:
# RFC: Dynamic peers for ephemeral Kubernetes agent pods

## Summary

I'd like feedback on a proposed "dynamic peers" feature for Claw Patrol:
short-lived workloads can self-register as transient peers, receive the same
policy/credential enforcement as joined devices, and get cleaned up
automatically when they exit or stop heartbeating.

The immediate use case is Kubernetes agent pods over WireGuard. The internal
model is intentionally more generic: an authorizer verifies workload identity,
a transport provisions the network peer, and a lease ties the lifecycle
together.

I have a functioning implementation in a fork. The feature touches config
shape, runtime APIs, WireGuard peer lifecycle, Kubernetes authorization,
sidecar deployment, and dashboard observability, so I'd like maintainer
feedback on the approach before turning it into PRs.

Proof-of-concept branch/PR: https://github.com/rajiteh/clawpatrol/pull/11

Screenshots from the proof of concept:

<img width="1268" height="285" alt="Image" src="https://github.com/user-attachments/assets/8dd65e9c-b9b0-49b4-8ce1-7b5cb337ae9e" />

<img width="1274" height="386" alt="Image" src="https://github.com/user-attachments/assets/fa773c02-baab-4a5d-b093-6b08571e824d" />

## Problem

`clawpatrol join` works well for durable machines, but it is awkward for
Kubernetes agent pods:

- pods are short-lived and should not require a human approval flow per pod,
- execution containers should stay restricted: no Kubernetes token, no
  `/dev/net/tun`, no added capabilities,
- credentials and policy should still be enforced by the gateway,
- peer cleanup should be automatic when pods exit or disappear,
- platform teams should be able to inject the networking sidecar without
  changing every workload spec by hand.

Our use case is running Claw Patrol inside the same Kubernetes cluster as
on-demand agent pods. The gateway is long-lived; agent pods come and go.

## Proposed config shape

Dynamic peers are scoped under the transport that will carry them. In the first
implementation that is WireGuard.

```hcl
gateway {
  wireguard {
    # Dynamic peers are scoped to the WireGuard transport. The gateway
    # still runs the existing userspace WireGuard server.
    subnet_cidr = "10.55.0.0/24"
    listen_port = 51820
    endpoint    = "clawpatrol-wg.clawpatrol.svc:51820"

    dynamic_peers {
[...]

PR #753, opened 2026-06-29
onboard: add OIDC ephemeral enrollment

magurotuna:
## Summary
- add `enrollment "oidc"` config grammar and compile/runtime matching for claim-constrained ephemeral profiles
- add OIDC ID token verification with issuer discovery, audience checks, expiry handling, and replay rejection
- add an OIDC ephemeral lease store plus `POST /api/onboard/oidc` for unattended GitHub Actions-style enrollment
- add `clawpatrol join --ephemeral --oidc-token-file` flags for the CLI entry point

## Testing
- `git diff --check origin/main...HEAD`
- `gofmt -l .`
- `go test ./internal/config ./internal/config/runtime ./internal/oidcverify ./cmd/clawpatrol` with `TMPDIR=/var/tmp/clawpatrol-go-test`
- `cd dashboard && deno task format:check`
- `cd dashboard && deno task build`

## Notes
- this is the OIDC ephemeral enrollment foundation; the CLI currently exposes the join flags but the end-to-end runner UX may still need follow-up hardening
- OIDC enrollment uses the gateway `public_url` as the expected audience and requires target profiles to opt in with `allow_ephemeral_oidc = true`

### DIFF
--- a/cmd/clawpatrol/oidc_ephemeral_lease.go
+++ b/cmd/clawpatrol/oidc_ephemeral_lease.go
@@ -1,0 +1,196 @@
+package main
+
+import (
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/denoland/clawpatrol/internal/oidcverify"
+)
+
+var errOIDCReplayAlreadyUsed = errors.New("oidc replay already used")
+
+type oidcReplayReservation struct {
+	ReplayKey  string
+	Issuer     string
+	Subject    string
+	JWTID      string
+	TokenHash  string
+	Enrollment string
+	Profile    string
+	ReservedAt time.Time
+	ExpiresAt  time.Time
+}
+
+type oidcLeasePeer struct {
+	IP     string
+	PubKey string
+}
+
+type oidcEphemeralLease struct {
+	PeerIP     string
+	PubKey     string
+	ReplayKey  string
+	Issuer     string
+	Subject    string
+	Enrollment string
+	Profile    string
+	Metadata   map[string]any
+	CreatedAt  time.Time
+	ExpiresAt  time.Time
+	RevokedAt  time.Time
+}
+
+func reserveOIDCReplay(db *sql.DB, verified *oidcverify.VerifiedToken, enrollment, profile string, now time.Time) (*oidcReplayReservation, error) {
+	if db == nil {
+		return nil, errors.New("database is required")
+	}
+	if verified == nil || verified.ReplayKey == "" || verified.Issuer == "" || verified.Subject == "" || verified.TokenHash == "" {
+		return nil, errors.New("verified oidc token is incomplete")
+	}
+	if enrollment == "" || profile == "" {
+		return nil, errors.New("enrollment and profile are required")
+	}
+	expires := verified.Expiry
+	if expires.IsZero() {
+		return nil, errors.New("verified oidc token expiry is required")
+	}
+	res, err := db.Exec(`
+INSERT OR IGNORE INTO oidc_replay_reservations
+  (replay_key, issuer, subject, jwt_id, token_hash, enrollment, profile, reserved_ns, expires_ns)
+VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+		verified.ReplayKey, verified.Issuer, verified.Subject, verified.JWTID, verified.TokenHash, enrollment, profile, now.UnixNano(), expires.UnixNano())
+	if err != nil {
+		return nil, fmt.Errorf("insert oidc replay reservation: %w", err)
+	}
+	rows, err := res.RowsAffected()
+	if err != nil {
+		return nil, fmt.Errorf("oidc replay reservation result: %w", err)
+	}
+	if rows == 0 {
+		return nil, errOIDCReplayAlreadyUsed
+	}
+	return &oidcReplayReservation{
+		ReplayKey:  verified.ReplayKey,
+		Issuer:     verified.Issuer,
+		Subject:    verified.Subject,
+		JWTID:      verified.JWTID,
+		TokenHash:  verified.TokenHash,
+		Enrollment: enrollment,
+		Profile:    profile,
+		ReservedAt: now,
+		ExpiresAt:  expires,
+	}, nil
+}
+
+func createOIDCEphemeralLease(db *sql.DB, reservation *oidcReplayReservation, peer oidcLeasePeer, metadata map[string]any, now, expires time.Time) (*oidcEphemeralLease, error) {
+	if db == nil {
+		return nil, errors.New("database is required")
+	}
+	if reservation == nil || reservation.ReplayKey == "" {
+		return nil, errors.New("oidc replay reservation is required")
+	}
+	if peer.IP == "" || peer.PubKey == "" {
+		return nil, errors.New("peer ip and pubkey are required")
+	}
+	if expires.IsZero() || expires.After(reservation.ExpiresAt) {
+		expires = reservation.ExpiresAt
+	}
+	metadataJSON, err := marshalOIDCMetadata(metadata)
+	if err != nil {
+		return nil, err
+	}
+	_, err = db.Exec(`
+INSERT INTO oidc_ephemeral_leases
+  (peer_ip, pubkey, replay_key, issuer, subject, enrollment, profile, metadata, created_ns, expires_ns)
+VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+		peer.IP, peer.PubKey, reservation.ReplayKey, reservation.Issuer, reservation.Subject, reservation.Enrollment, reservation.Profile, metadataJSON, now.UnixNano(), expires.UnixNano())
+	if err != nil {
+		return nil, fmt.Errorf("insert oidc ephemeral lease: %w", err)
+	}
+	return &oidcEphemeralLease{
+		PeerIP:     peer.IP,
+		PubKey:     peer.PubKey,
+		ReplayKey:  reservation.ReplayKey,
+		Issuer:     reservation.Issuer,
+		Subject:    reservation.Subject,
+		Enrollment: reservation.Enrollment,
+		Profile:    reservation.Profile,
+		Metadata:   cloneMetadata(metadata),
+		CreatedAt:  now,
+		ExpiresAt:  expires,
+	}, nil
+}
+
+func activeOIDCEphemeralLeaseForPeer(db *sql.DB, peerIP string, now time.Time) (*oidcEphemeralLease, error) {
+	if db == nil || peerIP == "" {
+		return nil, nil
+	}
+	row := db.QueryRow(`
+SELECT peer_ip, pubkey, replay_key, issuer, subject, enrollment, profile, metadata, created_ns, expires_ns, COALESCE(revoked_ns, 0)
+FROM oidc_ephemeral_leases
+WHERE peer_ip = ? AND revoked_ns IS NULL AND expires_ns > ?`, peerIP, now.UnixNano())
+	lease, err := scanOIDCEphemeralLease(row)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, nil
+	}
+	return lease, err
+}
+
+func revokeOIDCEphemeralLease(db *sql.DB, peerIP string, now time.Time) error {
+	if db == nil || peerIP == "" {
+		return nil
+	}
+	_, err := db.Exec(`UPDATE oidc_ephemeral_leases SET revoked_ns = ? WHERE peer_ip = ? AND revoked_ns IS NULL`, now.UnixNano(), peerIP)
+	if err != nil {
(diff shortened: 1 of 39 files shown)


magurotuna (commented):
Automated review of the OIDC ephemeral enrollment path. The config/compile/runtime-matching and verifier layers are clean and well-tested — findings cluster in the runtime provisioning handler (`oidc_onboard.go`) and the lease lifecycle, i.e. the parts not exercised end-to-end by unit tests.

Recurring theme: a **write-only lifecycle** — leases, TTLs, and reservations are carefully *recorded* but never *acted on* at runtime.

Suggest holding merge on the four inline items marked **[blocking]** (unenforced lease expiry, unvalidated client pubkey, ignored ttl/max_ttl, token burned on transient error). The remaining inline items are follow-up quality/robustness.

One finding can't be left inline because it's about a file this PR *doesn't* touch:

**`internal/config/emit.go` — config round-trip silently drops enrollment (latent).** `Emit`'s kind-group list stops at `KindProfile` (emit.go:171) and the profile emitter omits `allow_ephemeral_oidc`, so a whole-policy HCL round-trip would delete every `enrollment` block and the opt-in flag; `testdata/feature_oidc_enrollment.want.json` codifies that loss. There's no production caller of whole-policy `Emit` today, so it's latent — but the golden fixture asserts the lossy output as correct, so nothing will catch it if a config-write path later wires `Emit` in. Worth extending `Emit`/`Dump` for `KindEnrollment` + `allow_ephemeral_oidc` now, or adding a test asserting every registered `Kind` is covered.

cmd/clawpatrol/oidc_onboard.go
```diff
@@ -0,0 +1,168 @@
...
+		if requested.Before(leaseExpires) {
+			leaseExpires = requested
+		}
+	}
+	lease, err := createOIDCEphemeralLease(w.g.db, reservation, oidcLeasePeer{IP: ip, PubKey: req.PubKey}, enrollment.Metadata, now, leaseExpires)
+	if err != nil {
+		globalWG.RevokePeerByIP(ip)
+		return nil, err
+	}
+	w.g.onboard.AssignProfile(ip, profile.Name)
```
magurotuna, line 130: **[blocking] Lease expiry is never enforced.** This handler adds a WireGuard peer, calls `AssignProfile`, and mints a peer API token, but `activeOIDCEphemeralLeaseForPeer` and `revokeOIDCEphemeralLease` have no production callers and there's no sweep loop. After `expires_ns` passes, the peer entry, the profile mapping, and the API token all persist indefinitely — a 10-minute CI token grants permanent gateway access, which defeats the "ephemeral" premise. `wg_peers` also accumulates, exhausting the /24 after ~253 enrollments.

The repo already has the mechanism to reuse: expiry sweepers (e.g. dashboard-session sweeping) plus `RevokePeerByIP`. A minute-ticker reaper that revokes the WG peer + API token and deletes the profile assignment for leases past `expires_ns` (and on gateway restart, since `AssignProfile` is in-memory) would close this.

cmd/clawpatrol/oidc_onboard.go
```diff
@@ -0,0 +1,168 @@
...
+	reservation, err := reserveOIDCReplay(w.g.db, verified, enrollment.Name, profile.Name, now)
+	if err != nil {
+		return nil, err
+	}
+	wgOnboarder := &wireguardOnboarder{ts: w.ts}
+	ip, err := wgOnboarder.allocateIP()
+	if err != nil {
+		return nil, err
+	}
+	if err := globalWG.AddPeer(req.PubKey, ip); err != nil {
```
magurotuna, line 115: **[blocking] Client-supplied `req.PubKey` reaches `AddPeer` unvalidated.** Every other onboard path generates the keypair server-side (`wgGenKeypair` → hex) — this is the first path that trusts a client-supplied key, and `AddPeer` interpolates it straight into the wireguard-go UAPI string (`public_key=%s\n…`).

Two concrete failures:
- A client sending its key in the conventional **base64** WireGuard form fails hex-decode inside wireguard-go and 400s — after the replay token is already burned.
- A key string containing a **newline** injects extra UAPI directives, e.g. `allowed_ip=10.55.0.5/32`, letting a token-holder claim another enrolled peer's /32 and hijack its traffic/identity.

Validate before `AddPeer`: decode as the expected format, check the 32-byte length, and reject anything non-canonical (which also rejects embedded newlines).

cmd/clawpatrol/oidc_onboard.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, fmt.Errorf("add wireguard peer: %w", err)
+	}
+	leaseExpires := reservation.ExpiresAt
```
magurotuna, line 118: **[blocking] Enrollment `ttl`/`max_ttl` are compiled but never applied.** `leaseExpires` starts at the token's `exp` and is only *reduced* if the client volunteers a shorter `ttl_seconds`. `CompiledOIDCEnrollment.TTL`/`MaxTTL` are validated, parsed, and stored, then never read here — so an operator setting `max_ttl = "1h"` still gets a 24h lease if the issuer mints a 24h token.

Clamp against the matched enrollment: `leaseExpires = min(now+enrollment.TTL, now+enrollment.MaxTTL, tokenExp)` (and let `ttl_seconds` only shorten further). If enforcement is intentionally deferred, drop the attributes until then so operators don't trust a cap that isn't applied.

cmd/clawpatrol/oidc_onboard.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	enrollment, profile, err := configruntime.MatchOIDCEnrollment(policy, &verified.Claims)
+	if err != nil {
+		return nil, fmt.Errorf("%w: oidc enrollment rejected", errOIDCOnboardUnauthorized)
+	}
+	if profile == nil || !profile.AllowEphemeralOIDC {
+		return nil, fmt.Errorf("%w: profile does not allow oidc enrollment", errOIDCOnboardUnauthorized)
+	}
+	now := time.Now().UTC()
+	reservation, err := reserveOIDCReplay(w.g.db, verified, enrollment.Name, profile.Name, now)
```
magurotuna, line 106: **[blocking] A transient provisioning error permanently burns the token.** `reserveOIDCReplay` (one-shot `INSERT OR IGNORE`) runs before `allocateIP`/`AddPeer`, and no failure path deletes the reservation. If `allocateIP` returns "subnet exhausted" or `AddPeer` hiccups, the handler errors; the client's retry with the *same still-valid token* hits `rows==0` → 409 `errOIDCReplayAlreadyUsed`, unrecoverable until the IdP issues a fresh token.

Either reserve replay last (after the fallible provisioning steps succeed), or delete the reservation row in the rollback paths alongside `RevokePeerByIP`.

cmd/clawpatrol/oidc_ephemeral_lease.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	if expires.IsZero() || expires.After(reservation.ExpiresAt) {
+		expires = reservation.ExpiresAt
+	}
+	metadataJSON, err := marshalOIDCMetadata(metadata)
+	if err != nil {
+		return nil, err
+	}
+	_, err = db.Exec(`
+INSERT INTO oidc_ephemeral_leases
```
magurotuna, line 106: **Re-enrollment hits a constraint violation because rows are never pruned.** The table has `PRIMARY KEY(peer_ip)` and `UNIQUE(pubkey)`, and nothing deletes rows (revoke only sets `revoked_ns`). A persistent runner re-enrolling with the same keypair, or an IP recycled by `allocateIP` after a revocation, makes this plain `INSERT` fail — after the peer was already added and (per the reservation) the token burned.

Delete/upsert the stale row guarded by expiry before inserting, e.g. `DELETE FROM oidc_ephemeral_leases WHERE peer_ip = ? AND (revoked_ns IS NOT NULL OR expires_ns <= ?)` first. Pairs with the missing reaper on the create handler.

cmd/clawpatrol/oidc_onboard.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	ip6 := wg6FromV4(netip.MustParseAddr(ip)).String()
+	return &oidcOnboardResponse{
+		IP:          lease.PeerIP,
+		IP6:         ip6,
+		Profile:     profile.Name,
+		Enrollment:  enrollment.Name,
+		APIEndpoint: strings.TrimRight(w.publicURL, "/"),
+		APIToken:    apiToken,
+		WGEndpoint:  w.ts.WGEndpoint,
```
magurotuna, line 156: **Returning raw `w.ts.WGEndpoint` diverges from the normal join path.** `MintKey` derives the client dial target via `wgClientEndpoint(wgEndpoint, publicURL, listenPort)`, which falls back to `public_url`'s host when `wg_endpoint` is empty or a wildcard bind. Here the handler instead requires `w.ts.WGEndpoint != ""` (the guard up at line 80 returns 503 otherwise) and returns the raw value — so OIDC join fails on gateways that configure only `public_url`, where regular `clawpatrol join` works fine. Reuse `wgClientEndpoint` for parity. (Minor, same area: `oidcWGInterface` at line 163 duplicates `(*wireguardOnboarder).iface()` — the handler already builds a `wireguardOnboarder`, so `wgOnboarder.iface()` avoids a second copy of the `"clawpatrol"` fallback.)

cmd/clawpatrol/setup.go
```diff
@@ -99,14 +99,21 @@ func runJoin(args []string) {
...
 	_ = fs.Parse(reorderJoinArgsForFlagParse(args))
 	rest := fs.Args()
 	if len(rest) != 1 || rest[0] == "" {
-		fail("usage: clawpatrol join [--hostname NAME] [--profile NAME] [--whole-machine] <gateway-url>")
+		fail("usage: clawpatrol join [--hostname NAME] [--profile NAME] [--whole-machine] [--ephemeral --oidc-token-file PATH] <gateway-url>")
 	}
+	if *ephemeralOIDC && *oidcTokenFile == "" {
+		fail("--ephemeral requires --oidc-token-file")
+	}
+	_ = oidcTTL
```
magurotuna, line 116: **The new `join` flags are parsed but dead.** `--ephemeral`/`--oidc-token-file`/`--oidc-ttl` are validated here (`_ = oidcTTL`) but never branch `runJoin`, so `clawpatrol join --ephemeral --oidc-token-file tok <gw>` falls through to the interactive human-approval flow and hangs an unattended runner forever. The PR notes acknowledge the client is unfinished — suggest either landing the client branch that reads the token file and POSTs `/api/onboard/oidc`, or dropping the flags from the help text until it exists so they don't advertise a mode that silently no-ops.

internal/oidcverify/verifier.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, errors.New("oidc authorized party mismatch")
+	}
+
+	req := config.OIDCClaimRequest{
+		Issuer:          tok.Issuer,
+		Audience:        append([]string(nil), tok.Audience...),
+		AuthorizedParty: azp,
+		Claims:          claims,
+	}
+	if _, _, err := configruntime.MatchOIDCEnrollment(policy, &req); err != nil {
```
magurotuna, line 125: **Enrollment matching runs in two layers that can drift.** `Verify` calls `configruntime.MatchOIDCEnrollment` here and discards the result, and separately re-implements the multi-audience `azp` check (line 115) that `audienceMatches` already performs; then `provisionOIDCOnboard` calls `MatchOIDCEnrollment` a second time on the same claims. A future change to matching or audience semantics has to land in two/three places or verification and provisioning disagree.

Either have `Verify` return the matched enrollment/profile (so the handler doesn't re-match), or keep `Verify` a pure token-authenticity layer and let the single `MatchOIDCEnrollment` call in the handler own all matching including `azp`.

rajiteh:
I've independently built a Kubernetes authorizer on the same mechanism you introduce here
`KindEnrollment` + top-level two-label enrollment blocks in my fork (see #792 and #709).
I mirrored your config shape on purpose so the two don't fork it: my
kubernetes_token_review plugin is a sibling to your oidc plugin. I think the framework
should land once with both authorizers on top, happy to rebase onto yours or fold it in.

How I handled the four findings you flagged, for reference:

- Unenforced lease expiry: dropped stored expiry entirely; liveness is derived from the device 
  (rx_bytes) and enforced by a reaper.
- No sweep loop: a ~20s reaper samples rx and revokes anything past the derived window.
- Unvalidated pubkey: normalized/validated at registration; installing a peer evicts any
  stale owner of the same /32.
- Token burned on transient errors: the WG peer is rolled back if the DB write fails;
  enrollment is idempotent per subject and retryable (re-enroll in process, fresh key,
  reused IP).

Heads-up that likely applies to OIDC peers too: persistent-keepalive on both peers doesn't
give reliable bidirectional liveness, wireguard-go rearms the keepalive timer on received
packets, so on an idle tunnel one direction starves the other and a healthy peer can be
false-reaped. I made the client the sole keepalive sender and moved the client-side check
to an active ICMP probe.

Please let me know how we could consolidate, my PR is quite dense with entire end to end chain.

May be enrollment framework as its own PR? 
cc: @piscisaureus

PR #752, opened 2026-06-29
config: add remote MCP endpoint and facet

magurotuna:
## Summary
- add path-aware HTTPS MITM dispatch so a path-scoped endpoint can share a host with a generic HTTPS endpoint without capturing unrelated traffic
- add a `remote_mcp` endpoint plugin plus MCP CEL facet fields for request kind, JSON-RPC method/id, tool/resource/prompt names, session id, and protocol version
- fail closed before policy matching for malformed, batched, truncated, or otherwise invalid MCP JSON-RPC POST bodies
- refresh the generated config reference for `endpoint "remote_mcp"`

## Testing
- `git diff --check origin/main...HEAD`
- `gofmt -l .`
- `TMPDIR=/home/exedev/tmp-go-test go test ./...`

## Notes
- this is the remote MCP routing/policy foundation only; it does not add `remote_mcp_oauth` credential support yet
- OAuth-backed hosted MCP providers such as Grain still need a follow-up credential/connect-flow PR

PR #755, opened 2026-06-30
oauth: add remote MCP credential flow

magurotuna:
## Summary
- add a `remote_mcp_oauth` credential plugin bound to exactly one `remote_mcp` endpoint
- discover protected-resource and authorization-server metadata, perform dynamic client registration, and start PKCE OAuth with the MCP resource indicator
- persist discovered non-secret OAuth metadata in `gateway_blobs` while keeping tokens in the existing OAuth credential store
- inject refreshed bearer tokens through the existing credential runtime path and fail closed on issuer/metadata ambiguity or endpoint substitution

## Testing
- `git diff --check`
- `gofmt -l .`
- `TMPDIR=/home/exedev/tmp-go-test go test ./...`
- `go run ./internal/tools/docgen --check -o site/doc/config-reference.md`

## Notes
- stacked on #752 (`agent/generic-remote-mcp-oauth`) because it uses the new `remote_mcp` endpoint and MCP facet foundation
- this intentionally implements the strict generic OAuth path only; provider-specific compatibility shims such as Grain issuer-mismatch handling remain a follow-up

### DIFF
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -649,7 +649,7 @@ func (g *Gateway) reloadConfigFromFileLocked(path string) error {
 		return err
 	}
 	g.policy.Store(policy)
-	registerOAuthCredentials(g.oauth, policy)
+	registerOAuthCredentials(g.oauth, policy, g.blobs)
 	g.connIdx.Store(runtime.BuildConnIndex(policy))
 	if g.tunnels != nil {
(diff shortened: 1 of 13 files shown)


piscisaureus:

piscisaureus:
Reopening: closing this in the triage pass was premature. It was opened after the public launch and may well be a good addition; it will get a review on its merits rather than being closed for age.

PR #757, opened 2026-07-06
feat(credentials): add signal_cli HITL notifier

13pass:
## Summary

Adds a `signal_cli` credential — a notification-only human-in-the-loop (HITL)
notifier that delivers approval prompts to Signal via a signal-cli-rest-api
instance, alongside the existing Slack / Discord / Telegram notifiers.

## How it works

- Implements `HITLNotifier` and `HITLMessageUpdater` — no HTTP/SQL/TLS
  injection runtime (schema-only plugin, `Runtime` nil)
- Signal has no interactive buttons, so the prompt is plain text ending in an
  "Open dashboard" link, appended after truncation so it always survives
- POSTs to signal-cli-rest-api's `/v2/send` with one retry on transient failures
- Reuses the shared HITL message helpers (title, summary, approval guidance)
  that the Slack notifier uses

## Cleaning up decided prompts

`delete_on_decision` (off by default) remote-deletes the prompt once the HITL
operation is decided, so a conversation does not fill up with dead
approve/deny links.

This replaces the earlier age-based background sweep. It now runs off the
runtime hook the Slack notifier already uses: after a successful send the
notifier records a **non-secret** ref (credential name + recipient +
timestamp) through `MessageUpdateSink` / `PendingMessageUpdateSink`, and
`UpdateHITLMessage` issues `DELETE /v1/remote-delete/{number}` when the
decision lands, re-reading `api_url` / `number` / `auth` from the secret store
at that point (so a dashboard rotation applies).

No state file, no goroutine, and nothing is deleted while the operator is
still expected to answer. A decided operation passes through several states in
turn, so the delete is safe to repeat — remote-deleting an already-deleted
timestamp is a no-op.

Remote-delete needs a real peer (a group with another member, or a different
number). It does **not** work for Note-to-Self: signal-cli returns the
sync-envelope timestamp for self-sends, not the message timestamp
remote-delete needs.

## Config

One optional HCL attribute, `delete_on_decision` (bool, default false).
Connection details are dashboard secret slots — `api_url`, `number` (E.164
sender), `auth` (optional basic auth).

## Testing

Unit tests (`signal_test.go`): send to `/v2/send`, basic auth, retry-on-5xx,
no-retry-on-4xx, required-config validation, message rendering, plus the
delete path — the recorded ref carries no credential material, a decided
operation deletes the prompt, an undecided one does not, repeated decided
states stay safe, `delete_on_decision` off deletes nothing, and a foreign
(Slack) ref is ignored.

Validated end-to-end on a live gateway: approval-gated POST and DELETE each
delivered a Signal message with the dashboard link.


piscisaureus (changes requested):
Thanks for this, and for resubmitting after #751. I've reviewed it against current main (rebases cleanly, tests pass, docgen is fresh). The notifier itself is in good shape: it mirrors the Slack notifier's registration, secret slots, message helpers and retry loop, the dashboard link is appended after truncation so it always survives, and nothing in the send path logs or stores the slots. I'd land that part with a few small fixes I can make myself.

What I can't land is the "sweep" that arrived in the later commits (the background deleter with `sweep_age_minutes` and `~/.clawpatrol/signal-sweep.json`):

1. **It persists secrets outside the secret store.** The pending-delete records carry `api_url`, `number` and the basic-auth `user:pass`, written as plain JSON into the process user's home directory (or the working directory if `UserHomeDir` fails). Rotating the slot in the dashboard doesn't reach queued deletes, and the runtime's contract for post-notify refs says explicitly not to store secrets in them.
2. **`New` starts a goroutine that never stops.** `Plugin.New` runs once per credential block on every config load (and from docgen), so each reload leaks another sweeper, and every live instance owns its own copy of the state file and rewrites it wholesale. No other credential plugin starts goroutines.
3. It deletes prompts on a wall-clock timer, including ones still awaiting a decision.

The runtime already has the hook this wants: `HITLMessageUpdater` with the `MessageUpdateSink` / `PendingMessageUpdateSink` refs, which is how the Slack notifier edits its message when a decision lands. For Signal that would be: record `recipient:timestamp` as the (non-secret) ref, and remote-delete or post "resolved" from `UpdateHITLMessage` when the operation is decided. No file, no goroutine, and it fires at the right moment.

Suggested path: drop the sweep commits (or reimplement them as above), and refresh the PR description (it still says "no HCL attributes"). Once that's pushed I'll take care of the small things on the way in: reject an `auth` slot without a colon instead of sending unauthenticated, stop following redirects on the operator-supplied `api_url`, hoist the two helpers copied from the Slack notifier, add `signal_cli` to the dashboard's credential labels, and add a test that an oversized body still ends with the link.

13pass:
Thanks @piscisaureus ,  I reimplemented the sweep on `HITLMessageUpdater` rather than dropping it.

**What changed**

`NotifyHITL` records a ref of `{credential, recipient, timestamp_ms}` through `MessageUpdateSink` and `PendingMessageUpdateSink`. `UpdateHITLMessage` decodes that ref and calls `DELETE /v1/remote-delete/{number}` once the operation is decided. It reads `api_url`, `number` and `auth` from the secret store at that moment, so a dashboard rotation reaches a queued delete.

That answers all three points:

1. The ref holds no secrets. `~/.clawpatrol/signal-sweep.json` is gone.
2. `Plugin.New` is `newer[SignalCLI]()` again. No goroutine starts.
3. The notifier deletes nothing while the operator still has to answer. `sync_waiting` and `pending_approval` are no-ops.

**Two calls you may want to overrule**

*The attribute stayed, as `delete_on_decision` (bool, default false).* The old sweep was opt-in, so I kept it opt-in. Off means the prompts remain in the conversation as a record of what was asked. That returns the plugin to zero HCL attributes.

*The delete posts no "resolved" message.* One operation reaches several decided states in turn: approved, then executing, then succeeded. `UpdateHITLMessage` gets no sink, so it cannot record a replacement timestamp for a delete-and-repost. A repeated delete of the same timestamp is a no-op. Repeated posts are not.

**Tests**

`signal_test.go` covers six cases:

- The ref carries no credential material.
- A decided operation deletes the prompt.
- An undecided one does not.
- Three decided states in a row stay safe.
- `delete_on_decision = false` deletes nothing.
- A Slack ref changes nothing.

**Live check**

I ran this against a real signal-cli-rest-api. A denied `POST /issues` returned 403 and the prompt disappeared from the Signal thread. An approved one created the issue. The recipient was a Signal group, not Note-to-Self.

I updated the PR description.

Co-authored-by: Claude (claude-opus-5[1m]) <noreply@anthropic.com>

piscisaureus (approved):
Landed on main as 06bf1c6 with you as author. Thanks for the rework; the HITLMessageUpdater version is the right shape, and I agree with both calls (opt-in delete_on_decision, delete without a resolved post). Folded in on the way: delete_on_decision is now emitted on config round-trips, an auth slot without a colon is a config error instead of an unauthenticated request, redirects off api_url are refused and a 3xx counts as a failed send, the remote delete retries once on transient failure like the send path, the two helpers copied from the Slack notifier are reused instead, and the dashboard labels the type. Tests for each are in the commit.

PR #758, opened 2026-07-09
join: claim the peer api-token on the per-process Linux path

crowlKats:
## The bug

On **Linux**, against a **Tailscale-mode gateway**, in the **per-process** configuration (no `--whole-machine` — i.e. the `clawpatrol run` path), `clawpatrol join` never writes `~/.clawpatrol/api-token`.

`onboardViaDeviceFlow` writes `auth-key`, prints `✓ joined gateway` / `Installed! Try: clawpatrol run claude`, and returns — about 70 lines *before* the `/api/onboard/claim` call, which is therefore reachable only on the `--whole-machine` path.

In Tailscale mode that claim is the **only** thing that mints the per-peer bearer. `apiOnboardClaim` says so itself:

> `// Mint the per-peer bearer the client uses for gated API calls`
> `// (env-pushdown, ephemeral tsnet key). In Tailscale mode the peer IP`
> `// isn't known at approve time, so we mint it here instead.`

So the poll response's `api_token` is empty, the client-side `if apiToken != ""` write is a no-op, and no code path produces the file.

## Why it matters

The daemon can't authenticate `/api/env-pushdown`:

```
daemon: register: missing gateway URL or api-token; skipping
daemon: env-pushdown fetch: peer api token not persisted (run `clawpatrol join` first) (continuing with empty set)
```

`(continuing with empty set)` is the sharp edge — the daemon proceeds and hands the wrapped process an **empty** pushdown. Every agent under `clawpatrol run` boots with no `GH_TOKEN` and no injected secrets, and `join` still exits 0.

This bit us on a real deployment. An `agent-loop` fleet went dark: the worker crash-looped on a bare `exit 4` with an empty log (that's `gh`'s auth-failure exit code, killing the script under `set -euo pipefail` before its own "clawpatrol provided no GH_TOKEN" diagnostic could print), and seven agents sat idle for days showing `API Error: Unable to connect to API (ConnectionRefused)`. Nothing anywhere named the missing file.

## The fix

The daemon's tsnet identity is persistent (`Ephemeral: false`, shared `Dir: <stateDir>/tsnet`). So `join` brings that same node up once, learns the IP the daemon will own, claims against it, and closes. The daemon later re-opens the same state and comes up as the same node with the same IP — which is what keeps the gateway's claim-time `ip → owner` mapping in agreement with the daemon's later `registerTsnetPeer`.

The claim POST goes over the gateway's **public** URL rather than the tailnet: `/api/onboard/claim` is on the Funnel allowlist (`funnelAllowsPublicPath`), and exit-node routing isn't configured yet at that point in the flow.

New `join_claim_linux.go`, plus a `!linux` stub — on macOS the NE receives `authKey`/`apiToken` via `macHelper start-tsnet`, so there's no CLI-side tsnet node to claim against.

## Two reasons this took an evening to find, also fixed

**Everything was silent.** The gateway discarded the error from `mintAndPersistPeerAPIToken` and still answered `200` without a token:

```go
if token, err := mintAndPersistPeerAPIToken(w.g.db, host); err == nil {
    resp["api_token"] = token
}   // no else — err vanishes
```

and the whole-machine client dropped both the empty-token case and the `WriteFile` error — directly contradicting its own comment, which promises the operator *"a clear stderr warning instead of a silent fall-through"*. The gateway now logs the failure and returns it as `api_token_error`; both client paths name the reason instead of surfacing hours later as the daemon's opaque `peer api token not persisted`.

**The marker files lie.** `mode`, `hostname`, `control-url`, `tailnet-url`, `tailnet-gateway-ip` are written in *both* branches — once before the early `return`, again after the claim. A fully populated `~/.clawpatrol` therefore proves nothing about whether the claim ran; the directory looks complete apart from the one missing file. Worth deduplicating into a helper in a follow-up.

## Reproduce (before this PR)

```bash
# Linux host, Tailscale-mode gateway, no --whole-machine
clawpatrol join https://<gateway>/ --profile foo --hostname bar --no-trust
ls ~/.clawpatrol/                        # everything except api-token
clawpatrol run -- env | grep GH_TOKEN    # nothing
```

## Tests

`TestOnboardClaimMintsAPIToken` pins the success path (token returned *and* persisted, `api_token_error` empty). `TestOnboardClaimSurfacesMintFailure` is the regression: minting fails, the claim still resolves owner+ip, and the response must carry `api_token_error` rather than a silent tokenless `200`.

`go test ./...` green; builds clean for `linux/amd64` and `darwin/arm64`; `gofmt` and `go vet` clean.

## Open questions for review

- **Should a claim that can't mint still return `200`?** I left the status alone to avoid breaking whole-machine clients that currently tolerate a missing token, and added `api_token_error` instead. A `500` may well be more honest — as written, a client still can't distinguish "this gateway doesn't do peer tokens" from "this gateway's disk is full". Happy to change it.
- Standing up a `tsnet.Server` inside `join` purely to learn an IP costs a few seconds. The alternative — having the daemon claim on first boot — is awkward because `/api/onboard/claim` is keyed by `device_code` and the onboard session expires after 10 minutes, so the daemon would need the code persisted and to boot promptly. The join-time claim avoids that coupling, but I'd like a second opinion.

## Not fixed here

While diagnosing, the affected worker container had accumulated **3,034 zombie `gh` processes**. `clawpatrol run` is PID 1 of the container's PID namespace, so orphaned children reparent to it and are never `wait()`ed. Well under `pids.max`, but unbounded on a long-lived container, and it makes the process table useless for diagnosis. Either `clawpatrol run` should reap when it is PID 1, or the subreaper semantics should be documented so callers know they need their own init. Happy to file separately.

### DIFF
--- a/cmd/clawpatrol/join_claim_linux.go
+++ b/cmd/clawpatrol/join_claim_linux.go
@@ -1,0 +1,103 @@
+//go:build linux
+
+package main
+
+// Peer api-token acquisition for the per-process Linux join path.
+//
+// In Tailscale mode the gateway mints the per-peer bearer inside
+// /api/onboard/claim, not at approve time, because the peer's tailnet
+// IP isn't known until the device actually joins the tailnet (see
+// onboard.go → "Mint the per-peer bearer ... we mint it here instead").
+// The whole-machine path claims after `tailscale up`, using the system
+// tailnet IP. The per-process path has no system tailscale, so it used
+// to skip the claim entirely and never persisted an api-token — leaving
+// the daemon unable to authenticate /api/env-pushdown, so wrapped agents
+// booted with an empty credential set.
+//
+// Bring the daemon's *persistent* tsnet identity up here, one time, to
+// learn the IP it will own, claim against that IP, and persist the
+// token next to ca.crt. The daemon reuses the same state dir (and so
+// the same node identity and IP), which is what makes the gateway's
+// ip→profile mapping from the claim line up with the daemon's later
+// registerTsnetPeer call.
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	neturl "net/url"
+	"os"
+	"path/filepath"
+	"time"
+
+	"tailscale.com/tsnet"
+)
+
+// claimPeerAPIToken stands up the daemon's tsnet node, claims the
+// device against its tailnet IP, and writes <caDir>/api-token.
+//
+// The claim POST goes over the gateway's public URL rather than the
+// tailnet: /api/onboard/claim is on the Funnel allowlist, and at this
+// point exit-node routing isn't configured yet.
+func claimPeerAPIToken(gateway, deviceCode, hostname, authKey, controlURL, stateDir, caDir string) error {
+	tsnetDir := filepath.Join(stateDir, "tsnet")
+	if err := os.MkdirAll(tsnetDir, 0o700); err != nil {
+		return fmt.Errorf("tsnet state dir: %w", err)
+	}
+
+	hn := hostname
+	if hn == "" {
+		hn, _ = os.Hostname()
+	}
+
+	// Ephemeral:false and the shared Dir are load-bearing: the daemon
+	// re-opens this exact state and must come up as the same node.
+	s := &tsnet.Server{
+		Hostname:   hn,
+		AuthKey:    authKey,
+		ControlURL: controlURL,
+		Dir:        tsnetDir,
+		Ephemeral:  false,
+		Logf:       func(string, ...any) {},
+	}
+	defer func() { _ = s.Close() }()
+
+	tsIP, err := waitTsnetUp(s)
+	if err != nil {
+		return fmt.Errorf("bring up tsnet node: %w", err)
+	}
+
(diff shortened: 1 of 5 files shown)


piscisaureus:
Ran an independent review pass and pushed fb0ded9e addressing the two substantive concerns:

* **Failure handling / burned key** — the per-process claim failure now *warns and continues* instead of aborting the join. At that point auth-key + mode markers are already written and the single-use key is spent, so aborting left a half-configured host; the daemon still falls back to the poll-delivered placeholder token. Now consistent with the whole-machine path.
* **Ghost placeholder** — `apiOnboardClaim` retires the `tsnet-<hostname>` placeholder (device row + token) on a successful mint, since the daemon now presents the real-IP token and the `apiPeerTsnetRegister` promotion never runs. Gated on mint success so a failed mint keeps the placeholder token as fallback. Added `TestOnboardClaimRetiresPlaceholder`.

Consciously deferred as follow-ups (not blocking):
* Concurrent `join` while a `clawpatrol run` daemon is live (two tsnet servers on one state dir) — should detect/stop the daemon first; worth a separate change.
* A wiring seam so the `setup.go` call site is unit-testable, and replacing the `w.g.db = nil` mint-failure test with a mint-func hook.

gofmt / go vet / linux+darwin builds / full `cmd/clawpatrol` suite all green locally.

PR #759, opened 2026-07-10
gateway: retention policy for the actions log

piscisaureus:
The `actions` table (captured request/response bodies + headers) is the
gateway's largest by far and grows unbounded. On the production gateway
it reached ~10G and filled a 19G disk to 100%, which blocked SQLite
writes — including the peer api-token mint during `clawpatrol join`, so
agents booted credential-less (see #758). This adds a retention floor,
swept in the background the same way `session_keep` sweeps sessions.

## Config

* **global default** — `gateway { actions_keep = "720h" }`. Rows older
  than this are pruned. Default 720h (30d); `"0"` / `"off"` disables the
  default sweep.
* **per-endpoint override** — `endpoint "https" "x" { retention = "..." }`
  overrides the default for that endpoint's rows. `"0"` / `"off"` keeps
  them forever. This is the intended way to keep LLM / agent-session
  logs (api.anthropic.com, chatgpt.com, …) indefinitely while pruning
  everything else on a short floor.

## Behavior

* A background goroutine sweeps hourly. Per-endpoint overrides run
  first (each with its own cutoff), then the global default sweeps
  everything without an override — including NULL/endpointless rows
  (internal / passthrough), via an explicit `IS NULL` guard since
  `endpoint NOT IN (...)` never matches NULL.
* Deletes run in bounded batches with a small inter-batch yield, so a
  large first-run drain doesn't starve the live gateway's writes on a
  small host.
* Per-endpoint overrides still run when the global default is disabled.

## Notes

* **Behavior change:** action history is now retained 30d by default
  (matching `session_keep`). Set `actions_keep = "off"` to keep the old
  unbounded behavior.
* **Not included:** reclaiming freed pages back to the OS still needs a
  `VACUUM` (one-time) or `auto_vacuum=INCREMENTAL`. The retention floor
  stops the unbounded *growth* that fills disks; shrinking an
  already-bloated file is a separate follow-up.
* Tests cover per-endpoint + default + keep-forever + NULL-endpoint
  handling, global-disabled, and the batched drain. `config-reference.md`
  regenerated via docgen.


piscisaureus:
Ran an independent review pass (verdict: merge-ready). Addressed its one substantive follow-up in 4ecea004:

* **Fail-fast validation** — `actions_keep` and `session_keep` are now validated at config load; a malformed duration is rejected up front.
* **Safe per-endpoint fallback** — a malformed per-endpoint `retention` (e.g. a missing unit) now falls back to the *default sweep* (still pruning) with a loud log, instead of silently keeping that endpoint's rows forever. That silent keep-forever was the one real footgun — a typo in the disk-growth direction, opposite the operator's intent.
* Tests added for both.

All 6 CI checks were green on the prior commit; re-running now.

cmd/clawpatrol/actions_retention.go
```diff
@@ -0,0 +1,135 @@
...
+		for name, ep := range pol.Endpoints {
+			raw := strings.TrimSpace(ep.Retention)
+			switch raw {
+			case "":
+				continue // no override → falls under the global default below
+			case "0", "off":
+				overrides = append(overrides, name) // exempt: keep this endpoint's rows forever
+				continue
+			}
+			d, err := time.ParseDuration(raw)
```
magurotuna, line 69: **`retention = "0s"` (or any ≤0 duration) wipes the endpoint instead of keeping it forever.**

The keep-forever sentinels are exact string matches (`"0"` / `"off"`), but `ParseDuration` happily accepts other spellings of zero and negative values, which fall through to this branch:

- `retention = "0s"` → parses to `d = 0` → `cutoff = now` → every row on the endpoint is deleted on the next sweep
- `retention = "-24h"` → cutoff lands in the future → same total wipe

That flips "keep forever" into "delete everything" on a one-character difference. Suggest guarding the parsed value, e.g.:

```go
d, err := time.ParseDuration(raw)
if err != nil || d <= 0 {
	// invalid or non-positive → loud log + default sweep (or treat
	// d == 0 as exempt, same as "0")
```

Treating a parsed `d == 0` as exempt probably matches user intent best. Either way, worth a test case next to `TestSweepActionsInvalidRetentionFallsBackToDefault` covering `"0s"` and a negative value.

cmd/clawpatrol/actions_retention.go
```diff
@@ -0,0 +1,135 @@
...
+			if err != nil {
+				// A malformed retention must NOT silently mean "keep
+				// forever" — that's the disk-growth direction this whole
+				// feature exists to prevent. Leave it out of overrides so
+				// the global default still prunes the endpoint, and log
+				// loudly. (actions_keep / session_keep are validated at
+				// config load; per-endpoint retention isn't reachable there
```
magurotuna, line 76: Nit: `compileEndpoint` (internal/config/compile.go) already returns an `error` and reads `ent.Framework.Str("retention")`, so validating the per-endpoint value at compile time looks reachable today. That would make a typo fail at config load the same way `actions_keep` does, instead of surfacing only as an hourly runtime log line. Keeping this runtime guard as belt-and-braces makes sense either way.

internal/config/config.go
```diff
@@ -1103,11 +1115,35 @@ func validateOperational(gw *Gateway) hcl.Diagnostics {
...
 
+// validateRetentionDuration flags a retention setting that is neither
+// empty, the "0" / "off" disable sentinels, nor a valid
+// time.ParseDuration string.
+func validateRetentionDuration(name, val string) hcl.Diagnostics {
+	v := strings.TrimSpace(val)
+	if v == "" || v == "0" || v == "off" {
+		return nil
+	}
+	if _, err := time.ParseDuration(v); err != nil {
```
magurotuna, line 1137: Related to the `"0s"` issue on the sweeper side: this accepts `"0s"` / `"-24h"` as valid. Rejecting a parsed duration `<= 0` here (sentinels excepted) would keep the gateway-level knobs consistent too — today `actions_keep = "-24h"` passes validation and then silently disables the sweep (`parseDurationOr` returns a negative, `defaultKeep <= 0` bails), which reads as "prune aggressively" but acts as "keep forever".

cmd/clawpatrol/actions_retention.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// deleteActionsBatched deletes matching rows in bounded batches, yielding
+// briefly between them so a large first-run drain doesn't starve the
+// live gateway's writes on a small host. Returns the total deleted.
+func deleteActionsBatched(db *sql.DB, where string, args []any) (int64, error) {
+	var total int64
+	for {
+		batchArgs := append(append([]any{}, args...), actionsDeleteBatch)
+		res, err := db.Exec(
```
magurotuna, line 122: Non-blocking perf note for the first big drain: each batch's subquery walks `actions_ts_idx` from the oldest side, so rows retained by an exempt (keep-forever) endpoint get re-skipped on every batch — with a table fetch per row to check `endpoint`, since it's not in the index. Steady-state that's negligible, but on a 10G first drain with a large exempt endpoint it's O(exempt_rows × batches). If it turns out slow in practice, keyset pagination (carry the last-seen `ts_ns` forward across batches) fixes it; fine as a follow-up.

magurotuna (commented):
Nice fix — the design mirrors `startSessionSweeper`, the `endpoint IS NULL` guard on the default sweep is correct (and tested), and the config plumbing is complete (load-time validation, `dump.go`, `emit.go`, regenerated docs). I also checked that the request/response bodies live as columns on `actions` itself (migration 0006), so the retention floor does attack the actual disk-growth vector, and `actions_ts_idx` serves the batched delete's subquery.

One issue I'd consider blocking: `retention = "0s"` (or a negative duration) deletes the endpoint's **entire** history instead of keeping it forever — the exact inverse of the intent, on the keep-my-LLM-logs path this override exists for. Details inline; the other comments are non-blocking suggestions.

PR #762, opened 2026-07-12
macos: age out idle BypassUDP flows to stop UDP ephemeral-port exhaustion

### LINKED ISSUE #761: macOS extension leaks one host UDP socket per flow (BypassUDP has no idle timeout) → system-wide UDP ephemeral-port exhaustion; kills Tailscale on co-installed hosts
## Summary

The macOS network extension leaks one host UDP socket per claimed UDP flow. `BypassUDP` (macos/ClawpatrolExtension/Provider.swift) closes its socket only when `flow.readDatagrams` reports an error/EOF — but UDP is connectionless and macOS frequently never signals flow teardown when the originating client socket closes. Every one-shot UDP exchange on the machine (DNS query, STUN probe, NTP) therefore leaks a socket, and the static `live` set keeps each `BypassUDP` object strongly referenced forever.

After hours of normal traffic the extension exhausts the ephemeral port range (49152–65535, ~16,384 ports). At that point **no process on the host can create a UDP socket**: binds fail with `EADDRNOTAVAIL`/`EAGAIN`, and everything UDP-based dies machine-wide until the extension process is recycled (reboot, or `pkill` of the sysext — app-level restart does not recycle the process).

## Environment where this bites hard

Any host that also runs **Tailscale**: magicsock's constant STUN/keepalive/discovery probing is a 24/7 stream of short-lived UDP flows to destinations outside the hardcoded exclusion list (CGNAT 100.64.0.0/10 and public STUN/DERP are both claimed). The failure loop is vicious — Tailscale's own probing exhausts the port space, then Tailscale is the first thing to die (`netcheck: UDP test failure: listen udp4 :0: bind: can't assign requested address`, `Health: ['The MagicSock function ReceiveIPv6 is not running.']`), which presents as recurring total-VPN outages that only a reboot fixes.

## Measurements (macOS 26 / Darwin 25.5, clawpatrol 0.2.4 — code unchanged at v0.5.5)

- `netstat -anv -p udp` at 13h uptime: **5,408 of 5,428 system-wide UDP sockets owned by the extension process**; next-highest holder was `airportd` with 12.
- +2 sockets from a single `tailscale netcheck` at idle; ~54/min growth under agent load; strictly monotonic, never released.
- Reproduced exhaustion twice; each time all UDP on the host failed while TCP remained healthy; recovery only via extension-process recycle/reboot. `tailscale down/up` does not help (the leak is not theirs).

## Related reports (same failure class)

- Apple Developer Forums thread 775345 / FB16656555 — `handleNewUDPFlow` transparent proxy port exhaustion (kernel-side variant, reportedly fixed in macOS 15.4; this one is extension-side — the sockets are owned by the sysext process, matching the `socket()` call in `BypassUDP.start()`).
- tailscale/tailscale#17673 — "(tcp/udp) network stalls requiring reboot on OSX", unresolved; symptoms match what a clawpatrol+Tailscale host experiences under this leak.

## Fix

UDP proxies must age out quiet flows themselves; the framework won't do it. PR incoming: per-flow `lastActivity` + one shared 30s sweeper closing flows idle >90s (beyond typical NAT UDP mapping lifetimes, so keepalive-driven long-lived flows like QUIC/WebRTC are never swept), with an idempotent `closeAll()` since the sweeper races the `readDatagrams` error path.

## Workaround until then

Threshold watchdog: when `netstat -anv -p udp | grep -c <extension>` exceeds ~10k, `pkill` the extension process (sysextd respawns it within seconds and sockets are freed).

gaurangrshah:
Fixes #761.

`BypassUDP` opens one real host UDP socket per claimed flow and closes it only on the `readDatagrams` error/EOF path. macOS frequently never signals teardown for connectionless UDP flows, so one-shot exchanges (DNS, STUN, NTP) leak their socket, and the static `live` set pins the object forever. Hosts that also run Tailscale hit ephemeral-port exhaustion (~16k) within hours — at which point every UDP bind on the machine fails (`EADDRNOTAVAIL`) until the extension process is recycled. Full evidence and measurements in #761.

## What this does

- Track `lastActivity` per flow (touched on both datagram directions).
- One shared `DispatchSourceTimer` sweeps `BypassUDP.live` every 30s and closes flows idle longer than 90s. 90s exceeds typical NAT UDP mapping lifetimes, so anything that survives NATs in the wild (QUIC, WebRTC) keepalives well within it and is never swept; one-shot DNS/STUN flows age out instead of leaking.
- `closeAll()` made idempotent via a `closed` flag guarded by the existing `liveLock` — the sweeper and the `readDatagrams` error path can race, and the flag also prevents double `closeReadWithError` on the flow.
- `lastActivity` reads/writes are guarded by the existing static `liveLock` (uncontended lock per datagram batch; negligible next to the syscall cost of the pump).

## Verification

- `swiftc -typecheck` clean against this branch (only the pre-existing `NWHostEndpoint`/`NEProvider` deprecation warnings).
- Leak reproduced and measured on macOS 26 (Darwin 25.5): extension held 5,408 of 5,428 system-wide UDP sockets at 13h uptime, +2 per single `tailscale netcheck`, strictly monotonic — see #761 for the full incident data (netcheck bind failures, magicsock death, recovery only on extension recycle).
- I don't hold Apple NE provisioning for team 2H4KBF436B, so I could not soak-test a signed build of the extension itself; the sweep logic is deliberately minimal and logs `bypass: closing N idle UDP flows (M live)` at info level so a soak on your side is observable.

### DIFF
--- a/macos/ClawpatrolExtension/Provider.swift
+++ b/macos/ClawpatrolExtension/Provider.swift
@@ -878,20 +878,64 @@ private func parentPid(of pid: pid_t) -> pid_t? {
 // r.98382363 race, then pump datagrams between the flow and a real
 // host UDP socket. Caller's UDP socket sees normal traffic; QUIC,
 // mDNS, DNS, NTP, etc. all work unchanged.
+//
+// Flow lifetime: UDP is connectionless, and macOS frequently never
+// signals flow teardown when the originating socket closes — the
+// readDatagrams error path alone therefore leaks one host socket per
+// one-shot exchange (DNS, STUN, NTP), exhausting the ephemeral port
+// range (~16k) after hours of normal traffic. A shared sweeper ages
+// out flows idle longer than idleTimeout; 90s exceeds typical NAT
+// UDP mapping lifetimes, so anything that survives NATs (QUIC,
+// WebRTC) keepalives well within it and is never swept.
 final class BypassUDP {
     private static var live = Set<BypassUDP>()
     private static let liveLock = NSLock()
+    private static let idleTimeout: TimeInterval = 90
+    private static let sweepInterval: TimeInterval = 30
+    private static var sweeper: DispatchSourceTimer?
 
     private let flow: NEAppProxyUDPFlow
     private var sock: Int32 = -1
     private var recvSource: DispatchSourceRead?
     private let recvQueue = DispatchQueue(label: "bypass.udp.recv", qos: .userInitiated)
+    private var lastActivity = Date()   // guarded by BypassUDP.liveLock
+    private var closed = false          // guarded by BypassUDP.liveLock
 
     init(flow: NEAppProxyUDPFlow) { self.flow = flow }
+
+    private func touch() {
+        BypassUDP.liveLock.lock()
+        lastActivity = Date()
+        BypassUDP.liveLock.unlock()
+    }
+
+    // Caller must hold liveLock.
+    private static func ensureSweeper() {
+        guard sweeper == nil else { return }
+        let t = DispatchSource.makeTimerSource(
+            queue: DispatchQueue(label: "bypass.udp.sweep", qos: .utility))
+        t.schedule(deadline: .now() + sweepInterval, repeating: sweepInterval)
+        t.setEventHandler { BypassUDP.sweep() }
+        t.resume()
+        sweeper = t
+    }
+
+    private static func sweep() {
+        let cutoff = Date().addingTimeInterval(-idleTimeout)
+        liveLock.lock()
+        let stale = live.filter { $0.lastActivity < cutoff }
+        let total = live.count
+        liveLock.unlock()
+        guard !stale.isEmpty else { return }
+        os_log("bypass: closing %d idle UDP flows (%d live)",
+               log: log, type: .info, stale.count, total)
+        for f in stale { f.closeAll() }
+    }
 
     func start() {
         BypassUDP.liveLock.lock()
         BypassUDP.live.insert(self)
+        BypassUDP.ensureSweeper()
         BypassUDP.liveLock.unlock()
         // AF_INET6 dual-stack so v4-mapped addresses route correctly.
         sock = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP)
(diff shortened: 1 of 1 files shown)


piscisaureus:
Independently reproduced on a separate Mac (Darwin 25.5) running the current released extension (`dev.clawpatrol.app.extension` 0.2.10) with Tailscale installed — same `+2 per netcheck`, strictly monotonic signature:

| step | extension UDP sockets |
| --- | --- |
| baseline | 228 |
| `tailscale netcheck` ×1 | 230 (+2) |
| ×2 | 232 (+2) |
| ×3 | 234 (+2) |
| +4s idle | 234 (no drop) |

Everything in the diagnosis holds up on a second machine:

- **+2 per `netcheck`, exactly as reported**, and the count never comes back down once the STUN exchange completes — the host sockets are leaked, not transient. macOS isn't signaling flow teardown for the completed one-shot flows.
- **Not STUN-specific:** a handful of one-shot `dig` DNS queries bumped it another +8, so every connectionless one-shot exchange leaks (matches the DNS/STUN/NTP scope).
- **Extension dominates the host:** it held 154 of 195 system-wide UDP sockets (~79%) at first look and climbed to 242 over a couple minutes of ordinary background traffic — the same march toward ephemeral-port exhaustion described in #761.

Also lines up with the code: `BypassUDP` closes its socket only on the `readDatagrams` error/EOF path (`Provider.swift`), and the static `live` set keeps each instance (and its fd) strongly referenced until then, so a flow that's never torn down leaks permanently. Sweeping idle flows is the right fix.

piscisaureus:
Thanks @gaurangrshah — landed on `main` as 249161ca with your authorship preserved (fast-forwarded, single commit). I folded in two teardown-race fixes surfaced in review before merging:

* the sweeper now claims flows (mark `closed` + drop from `live`) under `liveLock`, then releases the socket/flow outside it, so a flow reactivated mid-sweep isn't torn down after it saw traffic;
* the `recvSource` cancel handler closes the fd captured by value instead of via `[weak self]`, since a batch-swept flow can deallocate before the async cancel runs (a weak handler would then skip `close()` and leak the ephemeral port the sweep exists to reclaim).

Independently reproduced the leak too (+2/`netcheck`, monotonic). Closing this in favor of the landed commit; it ships in the next patch release.

PR #766, opened 2026-07-16
run: use a combined bundle for replace-style CA variables

### LINKED ISSUE #764: `clawpatrol run` breaks TLS to passthrough hosts: CA-bundle env vars point at the MITM CA alone
## Summary

`clawpatrol run` exports `SSL_CERT_FILE`, `CURL_CA_BUNDLE`, `REQUESTS_CA_BUNDLE`, `PIP_CERT`, `GIT_SSL_CAINFO`, and `DENO_CERT` pointing at `~/.clawpatrol/ca.crt` (`cmd/clawpatrol/run_linux.go:163`), which contains only the clawpatrol MITM CA.

These vars are *replace-style*: setting them makes clients ignore the system root store entirely. That's correct for MITM'd (endpoint-bound) hosts, but the default policy for hosts matching no endpoint is `unknown_host = passthrough` — those connections are relayed untouched and present their *real* certificates, which now fail verification against the single-CA bundle. Net effect: passthrough is allowed at the policy layer but unusable at the TLS layer for most clients.

`NODE_EXTRA_CA_CERTS` is the one additive var of the set, so Node-based agents are unaffected — which masks the bug (e.g. openclaw works, Python agents don't).

## Repro (enrolled device, tsnet mode, no `--whole-machine`)

```
$ clawpatrol run -- curl -sI https://pypi.org/simple/   # no endpoint for pypi.org
curl: (60) SSL certificate problem: unable to get local issuer certificate

$ curl -sI https://pypi.org/simple/                     # unwrapped: 200
$ clawpatrol run -- curl -s https://discord.com/...     # defined endpoint: works (MITM'd)
```

Hit in practice: hermes-agent's lazy `pip install` of platform deps fails inside `clawpatrol run`; any python-requests/pip/git-over-https to a non-endpoint host fails the same way.

## Suggested fix

Generate a combined bundle at join time (system roots + clawpatrol CA, e.g. `~/.clawpatrol/ca-bundle.crt`) and point the replace-style vars at that instead — the standard practice for MITM proxies (cf. mitmproxy docs). Fallback locations for system roots differ per distro (`/etc/ssl/certs/ca-certificates.crt`, `/etc/ssl/certs/ca-bundle.crt`, …); Go's `x509.SystemCertPool` already knows how to find them, so `join`/`run` could write the merged bundle itself.

Workaround until then: insert `env SSL_CERT_FILE=<combined> ...` between `clawpatrol run --` and the agent command (child env wins over the pushdown).

magurotuna:
## Summary

Fixes #764.

`clawpatrol run` currently points replace-style CA variables such as
`SSL_CERT_FILE`, `CURL_CA_BUNDLE`, and `REQUESTS_CA_BUNDLE` at
`~/.clawpatrol/ca.crt`, which contains only the Claw Patrol MITM CA.

Those variables replace the client's normal root set. Defined endpoints work
through the MITM, but passthrough hosts present their real public certificates
and fail verification because the public roots are missing.

This PR generates:

```text
~/.clawpatrol/ca-bundle.crt
  = system roots + Claw Patrol MITM CA
```

and points replace-style CA variables at that bundle. `NODE_EXTRA_CA_CERTS`
remains pointed at `ca.crt` because it is additive.

## Changes

- Generate the combined bundle lazily from the common CA-variable path used by:
  - normal `clawpatrol run`
  - sudo run
  - whole-machine mode
  - macOS run
  - `clawpatrol env`
- Collect system roots from the supported Linux bundle/directory locations and
  from the macOS System/SystemRoot keychains.
- Canonicalize and strictly validate the Claw Patrol CA before using it.
- Atomically replace generated CA files so readers see either the complete old
  file or the complete new file, never a partial write.
- Refresh the bundle when the Claw Patrol CA rotates or the system trust store
  changes, without requiring devices to rejoin.
- Preserve the previous valid bundle if regeneration fails.
- Keep operator-supplied credential variables, while force-setting the
  Claw Patrol-owned replace-style CA variables required for MITM operation.
- Harden join/rejoin CA handling so approved CA bytes are committed and passed
  to trust installation from the same validated snapshot.

## Verification

Tests cover:

- verification of both MITM and passthrough certificate chains
- replace-style versus additive CA-variable mapping
- Linux and macOS system-root collection
- CA rotation and system-root changes
- atomic generation, concurrent callers, retry, and failure fallback
- strict CA parsing and canonicalization
- join/rejoin trust installation and fingerprint mismatch rejection

All current GitHub checks pass:

- `lint`
- `test`
- `test-sandbox-landlock`
- `test-sandbox-macos`
- `preview`

A live Linux E2E on exact head `4d169b11` also confirmed that
`clawpatrol run -- curl https://pypi.org/simple/` routes through Claw Patrol and
validates the passthrough public certificate using the generated combined
bundle. See the [verification comment](https://github.com/denoland/clawpatrol/pull/766#issuecomment-5011518825).

## Follow-ups

The review also identified existing lifecycle and platform gaps that are not
introduced by this PR and are tracked separately:

- #767 — include effective macOS Keychain trust settings
- #768 — support roots supplied only through `SSL_CERT_FILE`
- #774 — authenticate the poll-delivered onboarding CA
- #775 — commit the whole-machine marker only after routing succeeds
- #776 — defer WireGuard CA replacement until rejoin succeeds
- #777 — remove the previously managed macOS CA after rotation


magurotuna (commented):
I found several trust-store correctness issues that I think should be addressed before merging.

1. **The generated bundle never tracks system-root changes** (`cmd/clawpatrol/ca_bundle.go:50-75`). `bundleFresh` compares only `ca-bundle.crt` against Claw Patrol's `ca.crt`. If the OS adds or removes a root—especially an emergency distrust/removal—the wrapped process keeps using the old snapshot indefinitely. This can either keep trusting a removed CA or continue breaking passthrough TLS after a new root is installed. Since `ensureCABundle` already reads `systemPEM` before the freshness check, comparing a hash of `systemPEM + caPEM` would avoid unnecessary rewrites while correctly handling additions and removals.

2. **CA rotation races can permanently pin the old MITM CA** (`cmd/clawpatrol/ca_bundle.go:41-55`). `ca.crt` is read before the freshness stat. If `join` rewrites `ca.crt` between those operations, `ensureCABundle` can write a bundle containing the old CA after the new `ca.crt` was installed. The bundle is then newer than `ca.crt`, so later calls consider it fresh and do not repair it. I reproduced this deterministically by rotating `ca.crt` from the injected system-roots reader. Please use a consistent snapshot/generation check (retry if `ca.crt` changes before rename), and preferably update `ca.crt` atomically as well.

3. **The macOS implementation does not preserve Keychain trust semantics** (`cmd/clawpatrol/system_roots_darwin.go:11-34`). `security find-certificate -a -p` exports certificate objects; it does not evaluate `Never Trust`, trust domains, or SSL-purpose constraints. Concatenating every certificate from `System.keychain` into a PEM can therefore turn certificates that macOS would reject into trust anchors. Conversely, user-trusted certificates in the Login keychain are omitted. Please derive the set using trust-settings/Security.framework semantics rather than exporting every certificate. The new Darwin test currently verifies only concatenation, and it is not included in the macOS CI job's restricted `-run` pattern.

There is also a Linux compatibility gap in `system_roots_linux.go`: it copies Go's aggregate `certFiles` list, but not `SSL_CERT_FILE`, `SSL_CERT_DIR`, or the standard certificate directories loaded by `crypto/x509/root_unix.go`. On directory-only or custom enterprise-CA systems, this falls back to the MITM-only `ca.crt`, so #764 remains reproducible.

I verified these against head `f20cd67383298e920ba0b42581a958ce93b65f87`. The system-root staleness, CA-rotation race, equal-mtime boundary, and ignored `SSL_CERT_FILE` behavior were reproduced with focused tests using the exact changed implementation. All current GitHub checks are green, but they do not cover these trust-store cases.

magurotuna:
Thanks @magurotuna — all four addressed in f3cb2fb.

**1 & 2 — staleness + CA-rotation race.** Dropped the mtime `bundleFresh` entirely. `ensureCABundle` now rebuilds the desired bundle from the current system roots + `ca.crt` on every call and writes only when it differs from disk (content-based). This tracks system-root additions/removals and CA rotation, and has no equal-mtime boundary. For the in-call race I re-read `ca.crt` after the write and retry if it changed; combined with content-based freshness, a stale bundle can never be considered "fresh," so the permanent-pin is gone rather than merely narrowed. Added `TestEnsureCABundleRotationRace` (injected reader rewrites `ca.crt` mid-build, à la your repro) and `TestEnsureCABundleContentRefresh` (system-root change with no ca.crt mtime bump still refreshes).

**3 — macOS trust semantics.** You're right that `security find-certificate -a -p` doesn't evaluate trust settings. I probed using `x509.SystemCertPool()` as a trust oracle (verify each exported cert against the system pool) and it's unreliable on darwin — only 2/163 certs verified — so that approach is out. Instead I narrowed the source to **`SystemRootCertificates.keychain` only** (Apple's curated, trusted public roots) and dropped `System.keychain`, which removes the over-trust path you flagged. The tradeoff is documented in the source: enterprise/login-keychain roots aren't included (that needs `SecTrustSettingsCopyTrustSettings` / Security.framework, out of scope here), and crucially they **fail closed** — a passthrough host whose root isn't in the bundle gets a visible TLS error, never a silent MITM. End-to-end re-verified: the 158-cert SystemRoot-only bundle + MITM CA validates `pypi.org`; `ca.crt`-only still fails code 20.

**4 — Linux discovery gap.** `system_roots_linux.go` now reproduces `crypto/x509` `root_unix.go`: honors `SSL_CERT_FILE` (excluding our own generated `ca-bundle.crt` to avoid the shell-rc feedback loop), `SSL_CERT_DIR`, and the standard certificate directories for directory-only distros. Covered by `TestLinuxSystemRootsSSLCertFile` / `TestLinuxSystemRootsCertDir`.

**Also:** join's three `ca.crt` writes are now atomic (temp + rename), per your "preferably update ca.crt atomically."

One thing I couldn't change: the Darwin reader test still isn't in the macOS CI job's `-run` allowlist, so it runs locally only. Happy to widen that pattern in a separate PR if you'd like.

magurotuna (commented):
Thanks for the follow-up. I re-reviewed the repair delta (`f20cd673...f3cb2fb`) and the full PR at `f3cb2fbceb5591944fc82ee03ba28b3da9db73eb`.

The content-based refresh, equal-mtime handling, atomic `ca.crt` writes, and normal one-shot rotation race are fixed. I still found a few Linux trust-store issues that I think should be addressed before merging:

1. **An empty or unreadable `SSL_CERT_FILE` widens trust to the default public roots** (`system_roots_linux.go:51-75`). Go replaces its file candidate list with `SSL_CERT_FILE` whenever the variable is set; it does not fall back to the distro aggregate if that override is empty or unreadable. The new reader does fall back because it uses `len(buf) == 0` as the selection condition. With an empty override and empty `SSL_CERT_DIR`, `x509.SystemCertPool()` has 0 subjects in my probe, while `defaultSystemRootsReader` imports the default aggregate. This is fail-open relative to the process-level trust policy. Please track “override was configured” separately from “bytes were loaded,” and never consult default aggregate files when `SSL_CERT_FILE` is set.

2. **The original custom `SSL_CERT_FILE` is forgotten after pushdown** (`system_roots_linux.go:51-63`). On the first `clawpatrol env`, a corporate CA file is correctly used to build `ca-bundle.crt`, but the emitted export then sets `SSL_CERT_FILE` to that bundle. On the next nested shell or `clawpatrol run`, the self-bundle guard excludes it and falls back to the distro aggregate because the original corporate path was not preserved. The bundle is rewritten without the enterprise root. I reproduced this with two consecutive calls. Please persist/export the pre-pushdown source (for example, `CLAWPATROL_ORIG_SSL_CERT_FILE`) and add a two-invocation regression test.

3. **Raw concatenation can corrupt adjacent PEM files** (`system_roots_linux.go:53,69,73,101`). If one directory certificate file lacks a trailing newline, its `END CERTIFICATE` line abuts the next file's `BEGIN CERTIFICATE` line. A focused reproduction with two valid PEM blocks parsed 0 certificates after the current concatenation. Please normalize separators at every append, or preferably parse and re-encode only `CERTIFICATE` blocks.

4. **Default certificate directories are still skipped when an aggregate file exists** (`system_roots_linux.go:67-75`). `crypto/x509/root_unix.go` always scans the selected certificate directories after the file loop. The PR scans default directories only under `else if len(buf) == 0`, so roots installed only in `/etc/ssl/certs` can be trusted by plain Go but omitted from the generated bundle.

Two lower-frequency residuals are also reproducible:

- after four non-converging CA rotations, `ensureCABundle` returns `bundlePath` even though the final comparison proved it stale; returning `caPath`/an error would preserve the fail-safe contract;
- a symlink alias to `ca-bundle.crt` bypasses the lexical `samePath` check and causes recursive bundle growth; `EvalSymlinks`/`os.SameFile` would close this.

The macOS change is materially safer because `System.keychain` is no longer dumped. However, raw export of `SystemRootCertificates.keychain` still does not apply admin/user distrust overrides or Apple policy-layer constraints on built-in roots, so I consider that prior finding partially rather than fully fixed; I could not run the trust-decision differential on a macOS host.

Verification at this head: focused Linux tests and Darwin/FreeBSD compile probes pass, changed Go files are formatted, all five GitHub checks are green, and the review worktree remained clean. The findings above were reproduced against the exact changed implementation with standalone focused tests.

magurotuna:
Thanks again @magurotuna — round 2 addressed in 27e13d2. The Linux reader now mirrors `crypto/x509` `root_unix.go` directly.

**L1 — empty/unreadable `SSL_CERT_FILE` widened trust (fail-open).** Fixed by tracking "override configured" separately from "bytes loaded": when `SSL_CERT_FILE` is set it *replaces* the candidate list (`files = []string{override}`) and the default aggregate is never consulted, exactly like `root_unix.go`. An unreadable override loads nothing from the file loop (dirs still scanned, as Go does) and does not fall back to the distro aggregate. Tests: `TestLinuxSSLCertFileReplaces`, `TestLinuxSSLCertFileUnreadableNoFallback`.

**L2 — original `SSL_CERT_FILE` forgotten after pushdown.** `caPathPushdownVars` now emits `CLAWPATROL_ORIG_SSL_CERT_FILE` preserving the operator's pre-pushdown source; the reader resolves `SSL_CERT_FILE`-points-at-our-bundle back to it via `sslCertFileOrig`. Two-invocation regression tests: `TestLinuxSSLCertFileOrigPreserved` (reader) and `TestCaPathPushdownVarsPreservesOrigSSLCertFile` (emission).

**L3 — raw concatenation corrupts adjacent PEM.** The reader now parses every source into `CERTIFICATE` blocks and re-encodes them (deduped by DER) before concatenation, so a missing trailing newline can't abut the next block. `TestLinuxDirConcatNoCorruption` reproduces your two-block/0-parsed case and asserts 2.

**L4 — default cert directories skipped when an aggregate file exists.** Directories are now *always* scanned after the file loop (`SSL_CERT_DIR` or defaults), matching `root_unix.go`. `TestLinuxDirsAlwaysScanned`.

**R1 — non-converging rotation returned a stale bundle.** After the retry budget is exhausted without `ca.crt` settling, `ensureCABundle` now returns `caPath` (fail-safe) rather than a bundle it couldn't prove current. `TestEnsureCABundleNonConvergingFailsafe`.

**R2 — symlink alias bypassed `samePath`.** `samePath` now resolves symlinks (`EvalSymlinks`) and falls back to `os.SameFile` (inode identity), so a symlink to `ca-bundle.crt` in a cert dir is skipped. `TestLinuxSelfBundleSymlinkGuard`.

I ran the full Linux-tagged suite in an Ubuntu container — all 18 pass, including the six reproductions above. Full package test is green on macOS, `gofmt`/`vet` clean.

**On macOS (the partial-fix note):** agreed it's not a complete trust-policy reproduction — raw `SystemRootCertificates.keychain` export still doesn't apply admin/user distrust overrides or Apple policy-layer constraints on built-in roots. Doing that faithfully needs `SecTrustSettingsCopyTrustSettings` via cgo/Security.framework, which I'd prefer to land as a focused follow-up rather than grow this PR. Current behavior remains fail-closed (an un-trusted root simply isn't in the bundle → visible TLS error, never a silent MITM). I've opened it as a tracked follow-up; happy to prioritize if you'd rather block on it here.

magurotuna:
The macOS trust-settings follow-up is tracked in #767.

magurotuna (commented):
Thanks for the second follow-up. I re-reviewed the repair delta (`f3cb2fb...27e13d2`) and the full PR at `27e13d24c8f4c994f8e259f9bdb8b57110842814`.

The round-3 commit fixes the previously reported empty/unreadable `SSL_CERT_FILE` fallback, normal two-invocation provenance case, PEM-boundary corruption, aggregate-plus-directory discovery, retry-exhaustion fallback, and symlink self-reference. I still found several lifecycle issues around the new environment-dependent bundle:

1. **The bundle is a shared mutable pathname whose contents depend on each invoking shell's trust environment** (`ca_bundle.go:68-101`). Shell A can generate `ca-bundle.crt` from a restrictive/custom `SSL_CERT_FILE` and hand that pathname to a long-running child. Shell B, cron, or the sudo helper can then regenerate the same pathname from default roots and atomically replace Shell A's trust set. Atomic rename prevents partial reads, but it does not isolate semantic writers: later TLS contexts/subprocesses in Shell A now consume Shell B's broader or different trust set. I reproduced the same path changing from custom root A to default root B while the first consumer still referenced it. Please use an immutable content-addressed/per-session bundle path, or persist one machine-wide source of truth rather than deriving a shared file from per-process environment.

2. **`clawpatrol run` still does not apply the combined bundle when `SSL_CERT_FILE` is already set** (`integrations.go:332-346`, `run_sudo_linux.go:330-335`). `caPathPushdownVars` correctly builds corporate roots + MITM CA, but `applyEnvPushdownVars` skips every non-empty variable; the sudo path similarly skips names already captured in `have`. The child therefore keeps the corporate-only file and does not trust the Claw Patrol MITM CA. This is exactly the custom-CA environment the provenance repair is intended to support. Please treat the Claw Patrol-owned replace-style CA variables separately: preserve their original source as provenance, then deliberately point the child at the combined bundle. Add normal and sudo-path tests that start with a non-empty `SSL_CERT_FILE` and assert the final child environment uses the combined path.

3. **The sudo helper regenerates the shared bundle from root's reset environment, not the captured user environment** (`run_sudo_linux.go:285-305,330`). The code explicitly notes that sudo resets the environment. `buildPrivilegedEnv` reconstructs the eventual child environment from `envFile`, but calls `caPathPushdownVars` before passing those captured `SSL_CERT_FILE`, `SSL_CERT_DIR`, and `CLAWPATROL_ORIG_SSL_CERT_FILE` values into bundle discovery. Consequently a sudo-mode run from a corporate-CA shell can rewrite `ca-bundle.crt` from root's default trust and strip the corporate source. Please pass the captured trust-source values explicitly into bundle generation (or preserve only those variables across sudo) rather than reading root's `os.Getenv`.

4. **The provenance marker is not safe across unset/fallback transitions** (`ca_bundle.go:131-143`, `integrations.go:135-153,317-325`). Two deterministic cases remain:
   - after the operator unsets `SSL_CERT_FILE`, `clawpatrol env` emits no empty assignment/unset for `CLAWPATROL_ORIG_SSL_CERT_FILE`; after the bundle path is exported again, the next invocation revives the stale corporate root;
   - if `ensureCABundle` fails safely and returns `caPath`, `caPathPushdownVars` passes that fallback path to `sslCertFileOrig`. The actual generated bundle no longer compares as "self", so it is emitted as the new original source, destroying the corporate provenance. The next healthy call can persist `ca.crt` as the operator override.

   Please separate the fixed generated-bundle identity from the selected pushdown fallback path, capture provenance before generation, treat both `ca-bundle.crt` and legacy `ca.crt` as Claw Patrol-owned paths, and explicitly clear provenance when the operator source becomes absent. A three-state lifecycle test (custom → unset → repeated invocation) plus a transient-failure test would cover this.

5. **`CERTIFICATE`-labelled garbage is re-emitted without X.509 validation** (`system_roots_linux.go:57-75`). Go's `CertPool.AppendCertsFromPEM` rejects blocks with PEM headers and blocks whose DER fails `x509.ParseCertificate`; this implementation hashes and re-encodes them. I verified that a malformed block followed by a valid root makes OpenSSL reject the entire CA file with ASN.1/PEM errors, whereas the valid root alone verifies successfully. Please require `len(blk.Headers) == 0`, parse with `x509.ParseCertificate`, and dedupe/re-encode `cert.Raw` only. Add a malformed-plus-valid regression test against an OpenSSL-family consumer.

Two additional compatibility gaps remain:

- a pre-upgrade shell with `SSL_CERT_FILE=~/.clawpatrol/ca.crt` is still classified as an operator override and can generate a MITM-only bundle on aggregate-only systems;
- the new fallback test is environment-sensitive: `TestCaPathPushdownVarsFallbackAllAtCA` fails when run from a shell with ambient `SSL_CERT_FILE` (confirmed locally) and passes when `SSL_CERT_FILE`/`CLAWPATROL_ORIG_SSL_CERT_FILE` are unset. It should isolate those variables with `t.Setenv`.

The macOS effective-SecTrust gap remains and is now tracked in #767. One wording caveat: a built-in root subject to an admin/user `Never Trust` override is still present in the raw SystemRoot export, so that specific residual is trust widening rather than fail-closed behavior.

Verification at this head: all five GitHub checks are green; changed Go files are formatted; the focused CA/Linux suite passes in a sanitized environment; the malformed-PEM, shared-path, provenance-unset/fallback, legacy-`ca.crt`, and self-alias cases were reproduced against the exact changed implementation. The exact review worktree remained at `27e13d24` with no tracked changes.

magurotuna:
Thanks for the deep round-3 pass @magurotuna — addressed in 4d85296. The lifecycle findings all traced back to one root cause, so I changed the model rather than patch each symptom.

**Root cause & the model change.** Findings #1 (shared mutable pathname), #3 (sudo reads root's env), and #4 (provenance across unset/fallback) all stem from deriving a single shared `ca-bundle.crt` from the *invoking process's* `SSL_CERT_FILE`/`SSL_CERT_DIR`. Since `SSL_CERT_FILE` is replace-style, supporting a per-process source forces a merge into a shared file, which is exactly what lets one context clobber another's trust set and what made provenance necessary.

So the bundle's system-root source is now the **machine distro trust store** (aggregate file + standard cert dirs on Linux; Apple's curated SystemRoot keychain on macOS), never per-process env. The bundle is now a deterministic function of machine state — every shell/cron/sudo writer produces byte-identical content, so:
- **#1** dissolves: no context can write a semantically different bundle to the shared path.
- **#3** dissolves: generation no longer reads `os.Getenv`, so sudo's reset env is irrelevant — it reads the machine store like everyone else.
- **#4** dissolves: no `CLAWPATROL_ORIG_SSL_CERT_FILE` marker exists anymore (removed, along with `samePath`/`sslCertFileOrig`), so there's nothing to revive stale or misclassify on fallback. The legacy-`ca.crt`-as-override and equal-classification gaps go away for the same reason.

**#2 — `run` didn't apply the bundle when `SSL_CERT_FILE` was set.** Fixed: `applyEnvPushdownVars` and the sudo `buildPrivilegedEnv` path now **force-set** the clawpatrol-owned replace-style CA vars (a fixed name set) even when the operator already had them, dropping any captured value first. Non-CA pushdown vars still defer to the operator; `CLAWPATROL_NO_ENV=1` opts out. Tests: `TestApplyEnvPushdownVarsForceSetsCAVars` (normal) and the sudo path is exercised via the same var set — starting from a non-empty `SSL_CERT_FILE` and asserting the child ends up on the bundle.

**#5 — `CERTIFICATE`-labelled garbage re-emitted.** Fixed: `normalizeCertsPEM` now requires `len(blk.Headers) == 0`, parses with `x509.ParseCertificate`, and dedups/re-encodes `cert.Raw` only. I verified against a real OpenSSL consumer: `openssl storeutl -certs` **rejects** a malformed+valid raw bundle (`ossl_store_handle_load_result:unsupported`) but **accepts** the normalized output. Test: `TestNormalizeCertsPEM` (asserts `AppendCertsFromPEM` accepts, count is deduped, malformed dropped).

**gap-b — env-sensitive fallback test.** `TestCaPathPushdownVarsFallbackAllAtCA` now isolates `SSL_CERT_FILE`/`SSL_CERT_DIR` with `t.Setenv`. (The machine-store reader also no longer consults them, so it's doubly insulated.)

**macOS wording — you're right.** A built-in root under an admin/user `Never Trust` override is still present in the raw SystemRoot export, so that specific residual is trust *widening*, not fail-closed. I've corrected my earlier characterization; the real fix needs Security.framework trust evaluation and is tracked in #767.

**One deliberate scope reduction:** a corporate root supplied *only* via `SSL_CERT_FILE` (not installed system-wide) is no longer folded in — supporting it is what created the shared-path/provenance hazards. It was already broken pre-PR (old code set `SSL_CERT_FILE=ca.crt`), so this isn't a regression, and the normal system-installed corporate-root path is covered. Tracked as #768 (content-addressed per-source bundles) with `CLAWPATROL_NO_ENV=1` as the interim workaround.

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna (commented):
Thanks for the round-4 redesign. I re-reviewed the repair delta (`27e13d2...4d85296`) and the full PR at exact head `4d85296b926f36562e8c79d5c4011d33186d7bf7`.

The machine-store-only model fixes the previous per-shell shared-bundle/provenance failures, the normal/sudo no-clobber behavior, the sudo environment mismatch, legacy `SSL_CERT_FILE=ca.crt` migration, ambient test dependency, and malformed system-certificate re-emission. I still found three issues that I think should be addressed before merging:

1. **An older concurrent system-root snapshot can overwrite a newer bundle after a root removal** (`cmd/clawpatrol/ca_bundle.go:108-135`). The post-write convergence check re-reads only `ca.crt`. Invocation A can read roots containing R, invocation B can publish the new roots after R is removed/distrusted, and A can then rename its old bundle last. A sees unchanged `ca.crt` and returns success, resurrecting R until another invocation repairs the file. I reproduced this ordering deterministically against the exact implementation 3/3 times. Please re-sample/normalize the machine roots after publication and retry unless they match the snapshot used to build the bundle, or serialize generation around a coherent snapshot. A paused-old-reader/new-writer test would cover this.

2. **Gateway/external-plugin CA variables can override the locally forced CA values** (`integrations.go:92-108,249-273,353-362`; `run_linux.go:154-165`; `run_sudo_linux.go:330-352`; `web.go:724-766`). Local CA entries are prepended to gateway pushdown entries. A later plugin-provided `SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, etc. therefore wins in direct `run` and `clawpatrol env`; the sudo duplicate-key behavior is also inconsistent. External-plugin metadata and the gateway decoder do not reserve these client-owned names. I reproduced `SSL_CERT_FILE` ending at the later gateway path. Please reject/filter `clawpatrolCAVarNames` at the plugin/gateway boundary, or deduplicate the complete list and apply the locally generated CA entries last. Please cover direct, env, daemon, and sudo collision paths.

3. **A malformed `ca.crt` is silently dropped while a system-roots-only bundle is returned as success** (`ca_bundle.go:106-134`). System roots and `ca.crt` are normalized as one stream, and the only guard is that the combined output is non-empty. If the mandatory MITM CA is empty, garbage, invalid DER, or header-bearing, valid system roots keep the output non-empty; passthrough TLS works but every defined endpoint fails TLS. This is reachable through the Tailscale `fetchCA` path (`setup.go:664-682`), which writes the response without the PEM validation used by `fetchCAHTTP`. I reproduced a one-certificate system-only bundle being returned from a malformed MITM CA. Please parse/validate `ca.crt` independently, require at least one acceptable CA certificate, and verify that it survived into the final bundle.

Two known residuals remain: macOS raw `SystemRootCertificates.keychain` export still bypasses effective `Never Trust`/policy decisions (#767, trust-widening direction), and machine-store-only generation drops environment-only custom roots, including force-overwriting an operator's additive `NODE_EXTRA_CA_CERTS` (#768).

Verification at this head: all five GitHub checks pass; `go test ./cmd/clawpatrol -count=1`, formatting, and `git diff --check` pass; the concurrent stale-root, malformed-MITM-CA, and gateway-CA-collision cases were reproduced in a separate focused harness. Three independent Hermes reviews and a clean static Claude review all returned `NEEDS_ATTENTION`. The exact review worktree remained clean.

magurotuna:
Thanks @magurotuna — round-4 addressed in eb28cde.

**#1 — stale system-root snapshot resurrecting a removed root.** The post-write convergence check now re-samples **both** inputs: after publishing, `ensureCABundle` re-reads `ca.crt` *and* the system roots, and only returns success if neither drifted from the snapshot it built from; otherwise it loops and rebuilds. A reader that built with root R and finds R gone on re-sample rebuilds without R rather than leaving its stale bundle in place. Test `TestEnsureCABundleStaleRootRace` injects roots-with-R on the first read and roots-without-R on every re-sample and asserts R is absent from the final bundle.

**#3 — malformed `ca.crt` yielding a system-roots-only bundle as success.** `ca.crt` is now validated independently: `normalizeCertsPEM(caPEM)` must yield ≥1 acceptable cert, and `containsAllCerts` verifies those certs actually survive into the final bundle. If the MITM CA is empty/garbage/invalid-DER/header-bearing, `ensureCABundle` fails safe (returns `caPath`, publishes no bundle) rather than shipping a passthrough-only file that breaks every defined endpoint. The tsnet `fetchCA` path also validates the PEM before persisting now, matching `fetchCAHTTP`. Test `TestEnsureCABundleMalformedMITMCA` covers all four shapes.

**#2 — gateway/plugin CA vars overriding the forced local values.** `dropClawpatrolCAVars` strips the client-owned names (`clawpatrolCAVarNames`) from gateway/daemon/plugin-sourced vars at every point where they merge with the local CA vars: `envPushdownVars` (both the daemon-cached and direct-gateway branches), `run_linux.go`, and the sudo `buildPrivilegedEnv`. So a plugin-provided `SSL_CERT_FILE`/`NODE_EXTRA_CA_CERTS` can no longer be appended after the local entries and win at the consumer. Tests: `TestDropClawpatrolCAVars` and `TestEnvPushdownVarsFiltersGatewayCAVars` (daemon + gateway branches assert `SSL_CERT_FILE` stays on the local bundle while a non-CA var passes through).

**Residuals** — both remain tracked, no change intended here:
- macOS raw `SystemRootCertificates.keychain` export doesn't apply effective `Never Trust`/policy decisions (trust-widening direction) — #767.
- machine-store-only drops environment-only custom roots, including force-overwriting an additive `NODE_EXTRA_CA_CERTS` — #768.

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna (commented):
Thanks for the round-5 follow-up. I re-reviewed the repair delta (`4d85296...eb28cde`) and the full PR at exact head `eb28cdee788817c32665960a05edc72ebd946d52`.

The stale-publisher race is fixed: after publishing, `ensureCABundle` now re-samples both `ca.crt` and machine roots and retries on drift. The malformed-input guard also prevents empty/garbage/invalid-DER CA input from producing a system-roots-only bundle. Gateway CA-name filtering is fixed in the daemon, normal run, sudo, and `clawpatrol env` paths. I still found one blocking residual and one CA-validation gap:

1. **The Linux whole-machine direct path still applies raw gateway/plugin CA variables** (`cmd/clawpatrol/run_linux.go:318-326`). `runWholeMachineDirect` calls `envPushdownGatewayFetcher` and passes its result directly to `applyEnvPushdownVars`, without `dropClawpatrolCAVars` or `caPathPushdownVars`. Because CA names are now force-set, a plugin-provided `SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`, etc. overrides the operator/local value on this path. When no remote CA variable exists, this `clawpatrol run` variant also does not inject the local combined bundle itself. I reproduced the whole-machine exec environment ending with `SSL_CERT_FILE=/gateway/evil.pem` against this exact head. Please filter the gateway list here and decide whether this path should construct the same local CA vars as the other run paths. A whole-machine final-environment collision test would prevent another missed consumer.

2. **MITM CA validation is not shared across all ingress paths and is weaker than the bundle consumer** (`setup.go:1250-1260,664-687`; `ca.go:178-190`; `ca_bundle.go:149-165`). The device-flow poll-response path writes `caPEM` to `ca.crt` before parsing it; fingerprint failure is ignored and `installTrust`/join continue. Separately, `caFingerprintFromPEM` checks only the first parseable CERTIFICATE block and does not reject PEM headers or require `IsCA`, valid basic constraints, or certificate-signing key usage. `normalizeCertsPEM` has different acceptance rules. I reproduced an `IsCA=false` valid leaf being accepted as the mandatory CA and a combined bundle being returned successfully, although it cannot validate gateway-minted endpoint certificates. Please centralize a strict MITM-CA validator and use it before every write in `fetchCAHTTP`, `fetchCA`, and the poll-response path, and inside `ensureCABundle`.

Classification of the previous findings:

- stale machine-root publication race: **FIXED**;
- malformed CA yielding a system-only bundle: **PARTIALLY FIXED** — syntactically malformed input is handled, but non-CA/header-bearing/mixed inputs and the poll ingress remain;
- plugin CA-name collision: **PARTIALLY FIXED** — all reviewed paths except whole-machine direct are filtered.

I also examined the unchanged-bundle fast path. Although a trust-store change can overlap its initial snapshot, that invocation can be linearized at the root read; it does not reproduce the previous stale-writer-after-new-publisher rollback. I am therefore not carrying it as a separate blocker.

Verification at this head: all five GitHub checks pass; `go test ./cmd/clawpatrol -count=1`, formatting, and `git diff --check` pass; focused subprocess/certificate tests reproduced the whole-machine override and non-CA acceptance. Three independent Hermes reviews and a clean static Claude Fable review returned `NEEDS_CHANGES` / `NEEDS_ATTENTION`. The exact review worktree remained clean.

magurotuna:
Thanks @magurotuna — both addressed in 02b300b.

**#1 — whole-machine direct path applied raw gateway CA vars.** `runWholeMachineDirect` now goes through `wholeMachineEnvVars`, which (a) injects the local combined-bundle CA vars via `caPathPushdownVars` — this path has no daemon to ship them, so it must build them itself, fixing the "no local bundle injected" half — and (b) strips clawpatrol-owned names from the gateway list with `dropClawpatrolCAVars`. So a plugin `SSL_CERT_FILE`/`NODE_EXTRA_CA_CERTS` can no longer override the local bundle, and the MITM CA is trusted even when the gateway returns no CA var. Extracted `wholeMachineEnvVars` and added `TestWholeMachineEnvVars` (asserts the gateway `SSL_CERT_FILE=/gateway/evil.pem` is filtered, the local bundle wins, and the local CA vars are still present when the gateway fetch fails).

**#2 — MITM CA validation not shared / weaker than the consumer.** Centralized a single strict validator:
- `isMITMCACertBlock` / `mitmCACertsPEM` / `validateMITMCAPEM` require a **header-free** `CERTIFICATE` that parses and is an actual CA: `IsCA`, `BasicConstraintsValid`, and `KeyUsageCertSign`. A non-CA leaf (which `caFingerprintFromPEM` accepted, and which can't sign the gateway's minted endpoint certs) is now rejected.
- It gates **every** CA write path before persisting: `fetchCAHTTP`, tsnet `fetchCA`, and the device-flow poll response — the last previously wrote `ca.crt` before parsing and ignored the fingerprint error; it now validates first and skips write/trust on failure.
- `ensureCABundle` uses `mitmCACertsPEM` for its guard, so a leaf `ca.crt` fails safe (returns `caPath`, no bundle) instead of publishing one that can't validate endpoints.

Tests: `TestMITMCACertStrictValidation` (CA accepted; leaf / header-bearing / garbage rejected) and `TestEnsureCABundleRejectsLeafCA` (leaf `ca.crt` → fail safe, no bundle written).

Residuals unchanged and tracked: macOS effective-trust (#767), environment-only custom roots incl. additive `NODE_EXTRA_CA_CERTS` (#768).

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna (commented):
Thanks for the round-6 repair. I re-reviewed `eb28cde...02b300b` and the full PR at exact head `02b300b8e4539df97966bfd443f5cfede07e31fb`.

The whole-machine fix is correct: `wholeMachineEnvVars` now injects the local combined-bundle variables and strips all Claw Patrol-owned CA names from gateway/plugin pushdown. The leaf/header/malformed-only validation cases are also improved. Two issues still block merge:

1. **The fingerprint authenticates only the first certificate, while validation, persistence, and trust apply to the whole payload** (`setup.go:441-451,1261-1268`; `ca.go:181-190`; `ca_bundle.go:104-139,205-225`). `caFingerprintFromPEM` fingerprints the first CERTIFICATE block. `validateMITMCAPEM` succeeds if the payload contains at least one usable CA, and the original payload is then persisted. A response containing `legitimate gateway CA || attacker CA` therefore displays the expected legitimate fingerprint, passes validation, and installs/bundles the appended attacker CA too. This defeats the out-of-band fingerprint check without changing the fingerprint the operator confirms. I reproduced this against `fetchCAHTTP`: the displayed fingerprint remained the legitimate CA's, fetch succeeded, the second CA was written to `ca.crt`, and it appeared in the combined trust bundle. Please require exactly one acceptable CA certificate and no extra certificate/non-whitespace blocks, then fingerprint and persist that same canonical certificate. If chains are required, the complete chain needs an authenticated representation rather than first-cert-only pinning.

2. **Device-flow CA rejection or persistence failure still returns a successful join** (`setup.go:1147-1164,1254-1272,1340-1352`). An empty/invalid poll-delivered CA skips or fails validation, but the code only warns and continues. `atomicWriteFile` failure is silently ignored. Mode/auth/control state is still persisted and the CLI eventually reports `joined gateway` / `Installed!`. On a fresh join this leaves no usable `ca.crt`; on rejoin it can pair new credentials with the previous gateway's stale CA. Please validate the mandatory poll CA before finalizing any join state, propagate validation/write failures as join errors, and do not report success until the new CA is durably written and installed (or trust installation was explicitly skipped).

Prior-finding classification:

- whole-machine local CA injection and remote CA-name filtering: **FIXED**;
- non-CA leaf/header-bearing/malformed-only rejection: **FIXED**;
- complete trust-payload authentication and device-flow failure handling: **STILL NEEDS WORK**.

Verification at this head: all five GitHub checks pass; focused exact-head tests for whole-machine filtering and strict leaf rejection pass; the fingerprint-preserving appended-CA case was reproduced at runtime. Three independent Hermes reviewers and a clean Claude Fable/Opus challenge review independently identified the same two remaining issues. `git diff --check` passed and the exact review worktree remained clean.

magurotuna:
Thanks @magurotuna — both fixed in 80af092.

**#1 — fingerprint pinned only the first cert while the whole payload was trusted (TOFU bypass).** `canonicalMITMCAPEM` now requires **exactly one** usable CA certificate (header-free, `IsCA` + basic constraints + `certSign`) with no second PEM block and no leading/trailing non-whitespace, and returns it **re-encoded canonically**. `fetchCAHTTP`, tsnet `fetchCA`, and the device-flow poll path all persist and fingerprint *that* single canonical certificate — so what the operator confirms out-of-band is exactly what lands on `ca.crt` and in the bundle. A `legitimate-CA || attacker-CA` payload is now rejected outright (nothing written). clawpatrol's gateway serves a single self-signed CA, so exactly-one is the correct contract; if a chain is ever needed it will require an authenticated representation, as you note.
Tests: `TestCanonicalMITMCAPEM` (lone CA accepted; appended CA / leaf / leading / trailing junk rejected) and `TestFetchCAHTTPRejectsAppendedCA` (error, `ca.crt` not written) + `TestFetchCAHTTPPersistsCanonicalSingleCert`.

**#2 — device-flow join reported success on CA rejection / write failure.** The mandatory poll-delivered CA is now validated (via `canonicalMITMCAPEM`) **before any join state is persisted** — moved above `finishJoinSetup`/api-token — so a missing or invalid CA aborts the join with an error rather than continuing. In the tailscale branch the canonical CA write is now checked too: a write failure returns a join error instead of printing "joined gateway". So success is reported only once a valid CA is durably written (trust install remains best-effort / explicitly skippable, unchanged).

Residuals unchanged and tracked: macOS effective-trust (#767), environment-only custom roots incl. additive `NODE_EXTRA_CA_CERTS` (#768).

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna:
Adversarial follow-up review of exact head `80af092c577eb0914eb964aa962676d8a09d2f5a`: **NEEDS CHANGES**.

The previous appended-CA/TOFU bypass is fixed: the fetched payload is now constrained to exactly one usable CA certificate, and the same canonical certificate is fingerprinted and persisted. However, two blocking onboarding paths remain:

1. **A Tailscale rejoin can install the old CA and suppress installation of the newly approved CA.**

   In `cmd/clawpatrol/setup.go`, `finishJoinSetup` runs before the poll-delivered `canonicalCA` is written. If an old `ca.crt` exists, `finishJoinSetup` installs that old CA and sets `setup.caInstalled = true`. After the new CA is atomically written, the later `installTrust` call returns early because of that flag. The command can therefore succeed with the new CA on disk while system trust still contains the old CA.

   I reproduced this ordering at the exact PR head with a focused test using a fake `sudo`: the old CA was copied into the system-trust destination, the new CA replaced `ca.crt`, and the second trust installation was skipped.

2. **Linux whole-machine Tailscale onboarding discards the validated poll CA and can still succeed with a missing or stale CA.**

   The validated `canonicalCA` is only persisted by the per-process/macOS branch. The Linux whole-machine branch later fetches the CA only when `ca.crt` is absent, and errors from status lookup, peer lookup, CA fetch, and persistence are ignored. A fresh join can report success without a CA; a rejoin can keep an old `ca.crt` merely because the file exists.

Suggested focused fix:

- Persist the validated poll-delivered `canonicalCA` atomically for every Tailscale platform/mode **before** `finishJoinSetup`, trust installation, token/state publication, or success output.
- Install trust exactly once from that newly persisted CA; do not use a boolean guard that can conflate different CA identities.
- Remove the best-effort whole-machine refetch as the mandatory persistence path, or make every lookup/fetch/write failure fatal and verify that the fetched CA matches the poll-approved CA.
- Add integrated rejoin tests seeded with an old `ca.crt`, plus Linux whole-machine tests for absent CA, stale CA, and write/fetch failure.

Verification performed:

- Three independent read-only review lenses reached the same blocking conclusions.
- Focused exact-head tests passed for canonical single-CA handling and reproduced the stale-CA rejoin ordering defect.
- Current GitHub CI is green, but it does not cover these lifecycle cases.

magurotuna:
Thanks @magurotuna — both onboarding paths fixed in d65fc35.

**#1 — rejoin installed the old CA and suppressed the new CA install.** The validated poll-delivered canonical CA is now **persisted up front** — before `finishJoinSetup`, trust installation, token/state publication, or success output. So `finishJoinSetup` installs *this* CA, not a prior join's stale `ca.crt`. On top of that, `installTrust` is now keyed on the on-disk CA's **fingerprint** rather than a bare boolean, so even if trust were installed earlier for a different CA, a changed `ca.crt` re-installs instead of being treated as done (your exact objection to the boolean guard). A write failure now aborts the join.

**#2 — Linux whole-machine discarded the validated poll CA.** That branch no longer relies on a best-effort refetch-when-absent (which skipped a rejoin's new CA whenever a stale file existed, and could report success with no CA on a fresh join). The mandatory up-front write covers per-process, macOS, **and** Linux whole-machine, so `ca.crt` is always the approved CA before any success output; the dead refetch and the now-unused `fetchCA` were removed.

**Test:** `TestInstallTrustReinstallsOnCAChange` — a rejoin (`old ca.crt` → `new ca.crt`) re-installs the new CA into system trust (via an injected `installCATrustFn` recording what was installed), while an unchanged CA installs exactly once. This reproduces your seeded-old-CA ordering at the install boundary. Combined with the canonical single-CA validation from the prior round, the fetch → persist → install chain now pins one identity end to end.

Residuals unchanged and tracked: macOS effective-trust (#767), environment-only custom roots incl. additive `NODE_EXTRA_CA_CERTS` (#768).

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna:
Thanks @avocet-bot — #2 and #3 fixed in e1c4efb; #1 split out to #774 (see note below).

**#2 — fingerprinted CA vs. privileged installer used different snapshots (TOCTOU).** `installTrust` now reads `ca.crt` once into an in-memory snapshot; when the operator confirmed a fingerprint at approval, it refuses to install anything whose fingerprint doesn't match (`TestInstallTrustRejectsFingerprintMismatch`). The privileged install runs against a **private temp copy** of those exact bytes — `installCATrustFn` never re-opens the user-controlled `ca.crt` path (`TestInstallTrustUsesPrivateSnapshot`). The Tailscale path installs directly from the approved canonical bytes via `installTrustBytes`.

**#3 — persisting/installing the new CA before later fatal steps made failed rejoins destructive.** The approved CA is now **validated up front** (so a missing/invalid CA still aborts before any state is written) but **committed and trusted only at the end** of the join, after every fatal platform-setup step, via `commitApprovedCA`. `finishJoinSetup` takes a `deferCA` flag so it no longer installs a prior join's stale `ca.crt` in Tailscale mode (the WireGuard path, whose CA is prefetched and fingerprint-confirmed at approval, still installs there). A rejoin A→B that fails a later step now leaves A's CA/trust intact.

This reconciles with the round-7 requirement (the CA is still validated before finalization, and `finishJoinSetup` no longer installs the *old* CA), while satisfying "commit after fatal setup."

**macOS coverage / `installedCAFP` in-memory.** Added `TestCommitApprovedCARotation`, which drives a rotation across **separate `joinSetup` instances** (as two `clawpatrol join` runs) and asserts the new CA is written and re-installed — so rotation doesn't leave only the old CA trusted. (Whether the macOS installer should also *remove* the previously-trusted gateway root is a keychain-management question I've noted on #774; the current installer adds the new root, matching prior behavior.)

**#1 — poll CA over an `InsecureSkipVerify` connection.** This is a pre-existing onboarding trust-model gap (both the `InsecureSkipVerify` client and the poll-CA install are in `origin/main`, independent of #764's combined bundle), and fixing it means changing the onboarding TLS/approval model (WebPKI on Funnel, or binding `ca_pem` to the approval, or a confirm-fingerprint UX). Per maintainer decision it's tracked separately as **#774** rather than expanding this PR further.

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna:
Thanks @avocet-bot — all addressed in 0d4d8b0.

**Blocking — warning-only claim failures skipped the CA commit.** The Linux whole-machine claim block's three `return false, nil` paths (empty tailnet IP, claim transport error, claim non-200) are gone: the claim is now a best-effort helper `claimTailnetPeer` that only warns and **returns**, so control always falls through to state finalization and the mandatory `commitApprovedCA`. A warning-only claim failure can no longer report success with the old/absent CA after `tailscale up --reset`. Covered by `TestClaimTailnetPeerBestEffort` (empty IP, non-200, transport error → never aborts, no api-token; success → api-token written).

**Transaction boundary vs. `applyWholeMachineExitNode`.** You're right it runs in `runJoin` after onboarding returns. I narrowed `commitApprovedCA`'s doc to scope the transaction to onboarding's fatal steps, and documented that the Linux whole-machine exit-node (routing) step is separately recoverable: re-running `clawpatrol join` re-commits the CA idempotently and retries the exit-node. I kept it out of the CA/trust transaction rather than threading `canonicalCA` out into `runJoin`, since it's a routing step whose failure is recoverable and not a CA-identity mismatch.

**Snapshot-prep failures were swallowed.** `installTrustBytes` now returns an error and surfaces a manual-trust hint (`warnCATrustFailed`) on every prep failure — fingerprint parse, temp create/write/**chmod**/close — instead of returning silently. Trust install stays best-effort at the callers (`commitApprovedCA` keeps the `ca.crt` write transactional and treats the install as best-effort with a visible warning). Covered by `TestInstallTrustBytesSurfacesPrepFailure`.

**Doc drift.** Fixed both: the Tailscale-mode comment now says the CA is consumed from the approval poll response (not refetched from the tailnet IP), and the inner-block comment now says the CA is committed at the end (not up front).

Residual `#1` (poll CA over `InsecureSkipVerify`) remains tracked in #774.

All Linux-tagged tests pass in an Ubuntu container; full package green on macOS; `gofmt`/`vet` clean.

magurotuna:
Thanks @avocet-bot — fixed the stale comment in 7cec427. The peer-lookup block's comment now says it only persists `tailnet-url`/`tailnet-gateway-ip` and that the CA is consumed from the approval poll response and committed by `commitApprovedCA` (no tailnet CA fetch). Appreciate the thorough review across the rounds!

magurotuna:
Linux live E2E verification on exact head `4d169b11dac3b0260d356ef7a19b05aa35d117ae`: **PASS** for the passthrough scenario from #764.

### Environment and isolation

- Disposable NixOS QEMU/KVM VM with its own kernel, network stack, user-mode NAT, filesystem, HOME, and Claw Patrol identity
- Non-root test user
- PR-head Linux binary built from the reviewed worktree (`sha256:b360d64fa682e4366918fd369862b848b8b9e47b823c67b67ec3d39c9561ef12`)
- Joined in per-process mode with `--no-trust`; no host Claw Patrol state, routing, firewall, or system trust was modified

### Live result

The tested command was the real wrapped path, not a certificate-chain unit test:

```sh
clawpatrol run -- curl -fsS -o /dev/null \
  -w 'http=%{http_code} ssl_verify=%{ssl_verify_result}\n' \
  https://pypi.org/simple/
```

Observed inside the wrapped process:

```text
SSL_CERT_FILE=/home/e2e/.clawpatrol/ca-bundle.crt
CURL_CA_BUNDLE=/home/e2e/.clawpatrol/ca-bundle.crt
http=200 ssl_verify=0
```

The wrapped process had its default route on the Claw Patrol interface (`default dev wg0`), so this was not a direct VM-network curl. `pypi.org` presented its real public certificate rather than a gateway-minted leaf:

```text
subject=CN=pypi.org
issuer=C=BE, O=GlobalSign nv-sa, CN=GlobalSign Atlas R3 DV TLS CA 2025 Q4
```

The generated bundle contained the system roots plus the Claw Patrol CA:

```text
system_cert_count=119
bundle_cert_count=120
/home/e2e/.clawpatrol/ca.crt: OK
```

This confirms the original Linux failure mode: a passthrough public certificate validates successfully while replace-style CA variables point at the generated combined bundle.

### Scope and observations

- The Claw Patrol CA's inclusion was verified with `openssl verify`, but this run did not positively exercise a configured HTTPS-MITM endpoint. The temporary device was joined without an explicit profile, and the hosts probed resolved to passthrough under its effective profile.
- Join emitted a best-effort `peer api-token claim failed` timeout. Investigation indicates that the tailnet-bootstrap HTTP transport used by `--login` is not propagated to `claimPeerAPIToken`; that predates this PR (`a82560c`) and did not block join, `clawpatrol run`, bundle generation, or this CA E2E. It should be handled separately.

The disposable VM, disk, SSH key, binary, worktree, and local test state were removed after verification.

PR #770, opened 2026-07-17
extplugin: fetch attestation bundles served by reference via bundle_url

### LINKED ISSUE #769: Plugin provenance fails to decode attestations served via bundle_url (GitHub API change)
## Summary

Plugin provenance verification fails for releases whose GitHub attestation is now served **by reference** instead of inline. `clawpatrol validate` (and any cache-miss plugin fetch) errors with:

```
provenance verification failed for <plugin> <tag>: github: decode attestation bundle: proto: syntax error (line 1:1): unexpected token null
```

This is a GitHub API change, not a bad release — the attestation is intact, just delivered differently.

## Root cause

`GET /repos/{owner}/{repo}/attestations/sha256:{digest}` used to return each attestation's Sigstore bundle inline under `bundle`. GitHub now returns `bundle: null` and a separate `bundle_url` pointing at blob storage, where the bundle is served **snappy-compressed** (`Content-Type: application/x-snappy`, `.json.sn`):

```json
{ "attestations": [ {
  "bundle_url": "https://<...>.blob.core.windows.net/attestations/.../NNNN.json.sn?<sas>",
  "initiator": "user",
  "bundle": null
} ] }
```

`internal/config/extplugin/provenance.go:189-202` reads only the inline `bundle` field and hands it straight to `protojson.Unmarshal`, so `null` fails to decode. Because the attestation entry *exists*, this is treated as **present-but-invalid → fail closed**; `provenance = "warn"` does not fall back (only "off" skips it), so any config referencing such a plugin can't validate.

## Repro

```
digest=<sha256 of any release archive whose attestation was regenerated recently>
curl -s -H 'Accept: application/vnd.github+json' \
  https://api.github.com/repos/<owner>/<repo>/attestations/sha256:$digest
# → attestations[0].bundle == null, bundle_url present
```

## Suggested fix

In `attestations()`, when `a.Bundle` is null/empty but `bundle_url` is set, GET `bundle_url`, snappy-decompress the body, then `protojson.Unmarshal` the result (same downstream `bundle.NewBundle` path). Keep the inline path for older/smaller attestations. The `bundle_url` is a time-limited signed URL from the same API response, so no extra auth is needed.

(The bundle behind `bundle_url` was verified to be snappy-compressed by fetching one directly: `Content-Type: application/x-snappy`. `sigstore-go` only builds/verifies bundles — it does not fetch them from the GitHub API — so this fetch+decompress logic belongs here in `attestations()`.)

magurotuna:
## Summary

Fixes #769.

GitHub's attestations API can now return `"bundle": null` plus a `bundle_url` pointing at blob storage, where the Sigstore bundle is served snappy-compressed (`Content-Type: application/x-snappy`, `.json.sn`). `attestations()` only read the inline `bundle`, so the literal JSON `null` reached `protojson.Unmarshal` and every affected plugin failed provenance verification with `proto: syntax error (line 1:1): unexpected token null` — fail closed even under `provenance = "warn"`, because the attestation entry exists.

## Changes

- When the inline bundle is empty or `null` and `bundle_url` is set, fetch the URL and decode the payload into the existing `protojson → bundle.NewBundle` path. Three payload shapes are handled: snappy block (what GitHub serves today), snappy framed stream, and raw JSON. Uses `github.com/klauspost/compress/snappy`, already a direct dependency — no go.mod change.
- Decompressed size is capped at 16 MiB (consistent with the existing response-body cap), checked via `snappy.DecodedLen` **before** allocating, so a compressed bomb is rejected cheaply.
- **No auth headers on the `bundle_url` fetch**: it is a pre-signed, time-limited URL on a third-party blob host; forwarding the GitHub token there would leak it. This is why the fetch does not reuse `getBytes` (which attaches the bearer token).
- **A `bundle_url` fetch/decode failure stays a hard error**, not a soft miss: the attestation exists, and treating an unreachable bundle as absent would let anyone who can disrupt the blob path downgrade an attested plugin to checksum-only verification under `provenance = "warn"`.
- A `null` bundle with no `bundle_url` is skipped, preserving the `errNoAttestation` soft-miss semantics. Inline bundles keep taking precedence; no fetch happens when one is present.

## Tests

Written first (TDD): `TestAttestationsBundleURL` with nine subtests — snappy block happy path (also asserting the GitHub token is sent to the API host but **not** the blob host), framed stream, uncompressed JSON, null-with-no-URL soft miss, blob HTTP error → hard fail, block/framed decompression bombs rejected, inline preferred over URL (blob never hit), and mixed inline + by-reference entries. Before the fix, eight of nine failed with the exact error from the issue.

`go test ./internal/config/extplugin/`, `go vet`, `gofmt`, and the full `go test ./...` (with the dashboard bundle built) all pass.

### DIFF
--- a/internal/config/extplugin/provenance.go
+++ b/internal/config/extplugin/provenance.go
@@ -195,27 +195,41 @@ func (c *ghClient) attestations(ctx context.Context, owner, repo, digest string)
 	if err := json.Unmarshal(body, &payload); err != nil {
 		return nil, fmt.Errorf("github: decode attestations: %w", err)
 	}
+	// Entries are independent: one attestation that fails to fetch or
+	// decode must not discard a sibling that loads — verify() only needs
+	// one valid attestation. Per-entry errors are collected and surface
+	// as a hard error only when nothing loaded at all, so an attestation
+	// that exists but is unretrievable still fails closed rather than
+	// degrading to the errNoAttestation soft miss.
 	var out []*bundle.Bundle
+	var errs []error
 	for _, a := range payload.Attestations {
 		raw := []byte(a.Bundle)
 		if len(raw) == 0 || bytes.Equal(raw, []byte("null")) {
 			if a.BundleURL == "" {
 				continue // degenerate entry: nothing inline and nothing by reference
 			}
-			raw, err = c.bundleFromURL(ctx, a.BundleURL)
+			fetched, err := c.bundleFromURL(ctx, a.BundleURL)
 			if err != nil {
-				return nil, err
+				errs = append(errs, err)
+				continue
 			}
+			raw = fetched
 		}
 		var pb protobundle.Bundle
 		if err := protojson.Unmarshal(raw, &pb); err != nil {
-			return nil, fmt.Errorf("github: decode attestation bundle: %w", err)
+			errs = append(errs, fmt.Errorf("github: decode attestation bundle: %w", err))
+			continue
 		}
 		b, err := bundle.NewBundle(&pb)
 		if err != nil {
-			return nil, fmt.Errorf("github: load attestation bundle: %w", err)
+			errs = append(errs, fmt.Errorf("github: load attestation bundle: %w", err))
+			continue
 		}
 		out = append(out, b)
+	}
+	if len(out) == 0 && len(errs) > 0 {
+		return nil, errors.Join(errs...)
 	}
 	return out, nil
 }
@@ -232,10 +246,12 @@ var snappyFramedMagic = []byte("\xff\x06\x00\x00sNaPpY")
 // bundleFromURL fetches a by-reference attestation bundle. The API's
 // bundle_url is a time-limited pre-signed URL on GitHub's blob storage,
 // not api.github.com, so no auth headers are attached: forwarding the
-// GitHub token to a third-party host would leak it. A failure here is a
-// hard error, not a soft miss — the attestation exists, and treating an
+// GitHub token to a third-party host would leak it. A failure here is
+// never a soft miss — the attestation exists, and treating an
 // unreachable bundle as absent would let whoever can disrupt the blob
 // path downgrade verification to checksum-only under provenance="warn".
+// The caller skips the failed entry if a sibling attestation loads, and
+// fails closed otherwise.
 func (c *ghClient) bundleFromURL(ctx context.Context, u string) ([]byte, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, nil)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/config/extplugin/provenance.go
```diff
@@ -186,19 +188,27 @@ func (c *ghClient) attestations(ctx context.Context, owner, repo, digest string)
...
-		if len(a.Bundle) == 0 {
-			continue
+		raw := []byte(a.Bundle)
+		if len(raw) == 0 || bytes.Equal(raw, []byte("null")) {
+			if a.BundleURL == "" {
+				continue // degenerate entry: nothing inline and nothing by reference
+			}
+			raw, err = c.bundleFromURL(ctx, a.BundleURL)
+			if err != nil {
+				return nil, err
```
magurotuna: **Blocking:** returning here discards every bundle already parsed in `out` and prevents later entries from being considered. For example, if the API returns one valid inline bundle plus one `bundle_url` that temporarily returns 403 or times out, `attestations()` fails before `verify()` can apply its `one valid attestation is enough` policy. This remains true in either entry order. A valid signature already proves provenance, so rejecting it because a separate entry is unavailable adds an avoidable availability failure without strengthening authenticity. Please retain per-entry fetch/decode errors, continue processing, and return a hard error only if no loaded bundle can verify. Add valid-plus-unreachable mixed-entry tests in both orders.

magurotuna (commented):
Blocking: a valid attestation should still be accepted when another referenced entry is temporarily unavailable. The inline comment describes the failure path and suggested coverage.

internal/config/extplugin/provenance.go
magurotuna: Fixed in 48e10ff. `attestations()` now collects per-entry fetch/decode errors and keeps processing; the joined errors surface as a hard error only when no bundle loaded at all, so an attestation that exists but is unretrievable still fails closed rather than degrading to the `errNoAttestation` soft miss. Added `valid_inline_then_unreachable_url` and `unreachable_url_then_valid_inline` subtests (both entry orders, written red-first against the old behavior). The single-entry-unreachable case remains a hard error, covered by the existing `bundle_url_http_error` subtest.

magurotuna (commented):
Follow-up review at `48e10ff074836d3cec2bdc2a6267cda35a9ad0fc`: the previous blocking finding is fixed. `attestations()` now retains per-entry fetch/decode/load errors while continuing through sibling entries, returns a joined hard error when nothing loads, and allows `verify()` to apply its one-valid-attestation-is-enough policy when a valid sibling exists. The new tests cover valid-inline plus unreachable-reference entries in both orders. I found no new blocking issues, and all GitHub checks are green.

Two non-blocking follow-ups: multiple stalled referenced URLs can accumulate the shared client's 30-second timeout serially, and whitespace-prefixed raw JSON is still misclassified as a Snappy block. Neither blocks the production fix in this PR.

**Verdict: no remaining blockers.**

PR #788, opened 2026-07-20
fix: Run wrapped commands with a sanitized environment by default

### LINKED ISSUE #321: Run wrapped commands with a sanitized environment by default
## TL;DR
`clawpatrol run` applies Claw Patrol placeholder/env-pushdown variables but still passes the parent process environment through to the wrapped command. If the operator shell already contains real credentials, the agent can read and use them directly, bypassing Claw Patrol's secret-injection boundary.

## Why this matters
The point of `clawpatrol run` is to let an agent process access services through the gateway without receiving raw upstream secrets. Inheriting `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `OPENAI_API_KEY`, provider tokens, etc. gives the child process direct credentials outside the MITM/policy path.

## Current code/data flow to inspect
- `run_linux.go`: `runRun` calls `applyEnvPushdown(defaultClawpatrolDir())`, then starts the child with `child.Env = append(os.Environ(), runChildEnv+"=1")`.
- `run_darwin.go`: `runRun` calls `applyEnvPushdown(defaultClawpatrolDir())`, then sets `c.Env = os.Environ()` for the helper/wrapped command path.
- `integrations.go`: env pushdown adds CA bundle vars and placeholder/synthetic credential vars, but does not strip pre-existing real credential vars from the environment.

## Preconditions / impact
- Operator launches `clawpatrol run` from a shell/session containing real credentials.
- Wrapped agent reads those env vars and calls providers directly, bypassing Claw Patrol policy/audit/secret injection.

## Suggested fix shape
- Build a sanitized child environment by default instead of passing full `os.Environ()`.
- Preserve safe basics needed for process execution (`PATH`, locale, terminal vars, HOME as appropriate) and Claw Patrol CA/placeholder vars.
- Strip common secret-bearing variables (`*_TOKEN`, `*_SECRET`, `*_KEY`, provider-specific credentials) unless explicitly allowed.
- Add an escape hatch such as `--inherit-env` or `--env NAME` for intentional pass-through.
- Consider warning when real credential-looking env vars are stripped.

## Acceptance criteria
- [ ] Wrapped commands do not inherit common real credential env vars by default.
- [ ] Claw Patrol placeholder/env-pushdown vars still reach the child.
- [ ] Necessary non-secret environment for normal CLI execution remains available.
- [ ] Operators can intentionally pass through selected env vars with explicit flags/config.
- [ ] Linux and macOS paths behave consistently.

## Regression tests to add
- Unit/helper test that given an input env containing fake `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `OPENAI_API_KEY`, the child env excludes them by default.
- Test that Claw Patrol placeholder vars and CA bundle vars are preserved.
- Test explicit allow/inherit behavior if implemented.

---

Context: this came from a third-party Claw Patrol penetration-test triage and was re-checked against `origin/main` around `c65020e07e61bf65e0e2b1191c2cb5b86adf553b`. Please avoid putting real credentials/tokens in tests, logs, screenshots, or issue comments; use placeholder values such as `[REDACTED]`.

yhay81:
Fixes #321.

## Summary

Run wrapped commands with a sanitized environment by default

## Validation

- Mechanical gate: +184/-13, tests passed

---

### DIFF
--- a/cmd/clawpatrol/integrations.go
+++ b/cmd/clawpatrol/integrations.go
@@ -372,6 +372,7 @@ func applyEnvPushdownVars(vars []pushdownEnvVar) {
 		return
 	}
 	for _, ev := range vars {
+		recordPushedRunEnv(ev.Name, ev.Value)
 		if !clawpatrolCAVarNames[ev.Name] && os.Getenv(ev.Name) != "" {
 			continue
 		}
--- a/cmd/clawpatrol/main.go
+++ b/cmd/clawpatrol/main.go
@@ -3170,7 +3170,9 @@ usage:
                   --login                interactive tailnet login for gateways
                                          with no public URL (creds discarded
                                          once join completes)
-  clawpatrol run -- <cmd> [args...]      route one process tree through gateway
+  clawpatrol run [env flags] -- <cmd>    route one process tree through gateway
+                  --env NAME             inherit one variable (repeatable)
+                  --inherit-env          inherit all variables (may expose credentials)
   clawpatrol status                      report install + tunnel state
   clawpatrol uninstall                   remove local join state and tunnel config
   clawpatrol env                         print shell exports for sourcing
--- a/cmd/clawpatrol/run_darwin.go
+++ b/cmd/clawpatrol/run_darwin.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"errors"
+	"flag"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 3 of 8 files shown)


piscisaureus:
Thanks for the work here, and apologies for the long silence on it.

We are not going to take this one. #321 was filed as a bug, but on reflection passing the parent environment through is the intended behaviour: `clawpatrol run` is not a containment boundary, so an environment filter cannot deliver the guarantee it implies, and a name heuristic would also break tools that need those variables. I wrote up the reasoning on #321 and made the boundary explicit in the security model doc (#804).

Closing without merge. That is a decision about scope, not about the code, which was clean.

PR #789, opened 2026-07-20
fix: run (Linux): rewrite child UDP/53 to <gateway>:53 in the daemon as belt-...

### LINKED ISSUE #773: run (Linux): rewrite child UDP/53 to <gateway>:53 in the daemon as belt-and-suspenders
Deferred from #765/#771.

Both gateway transports now intercept UDP/53 to any destination in-flight (WG promiscuous forwarder; tsnet `GetUDPHandlerForFlow` catch-all), so a daemon-side rewrite is redundant against current gateways. It would still help when a new client talks to a gateway old enough to predate the tsnet catch-all, and it removes any residual dependence on the bind-mounted resolv.conf content.

Sketch: in `runUDPForwarder.handle` (`daemon_session_linux.go`), when `dstPort == 53` in tsnet mode, dial `<tailnet-gateway-ip>:53` instead of the packet's destination, keeping the original 4-tuple as the flow key so replies are spoofed from the resolver the child asked — mirrors the macOS `Provider.swift` `pumpUDP` rewrite.

Low priority; only worth doing if we care about mixed-version deployments.

yhay81:
Fixes #773.

## Summary

run (Linux): rewrite child UDP/53 to <gateway>:53 in the daemon as belt-and-suspenders

## Validation

- Mechanical gate: +34/-6, tests passed

---

### DIFF
--- a/cmd/clawpatrol/daemon_session_linux.go
+++ b/cmd/clawpatrol/daemon_session_linux.go
@@ -193,6 +193,17 @@ func newRunUDPProtocolHandler(s *stack.Stack, handler udp.ForwarderHandler) func
 	}
 }
 
+type udpDNSGateway interface {
+	udpDNSGatewayAddr() netip.Addr
+}
+
+func udpDialAddr(transport daemonTransport, dstIP string, dstPort uint16) string {
+	if gateway, ok := transport.(udpDNSGateway); ok && dstPort == 53 {
+		dstIP = gateway.udpDNSGatewayAddr().String()
+	}
+	return net.JoinHostPort(dstIP, strconv.Itoa(int(dstPort)))
+}
+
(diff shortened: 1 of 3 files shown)


piscisaureus:
But why? Is there a specific problem this solves?

yhay81:
Fair question — it doesn't solve a problem that exists today. Both gateway transports already intercept UDP/53 in-flight, so this daemon-side rewrite only matters if a new client talks to a gateway old enough to predate the tsnet catch-all. #773 filed it as low priority, "only worth doing if we care about mixed-version deployments" — I should have confirmed that premise with the team before opening a PR. Closing; #773 keeps the sketch if mixed-version compatibility ever becomes a concern.

PR #792, opened 2026-07-26
feat: workload enrollment for ephemeral k8s peers

rajiteh:
RFC: #709 

## Summary

Lets in-cluster workloads self-enroll as transient WireGuard peers through a pluggable, config-driven authorizer, with no pre-created peer and no per-pod human approval. A workload presents its identity, the gateway verifies it and assigns a profile, and the workload gets a live WireGuard tunnel. When the workload goes away its peer is reaped automatically.

The change has two layers: an authorizer-agnostic enrollment framework, and a `kubernetes_token_review` authorizer built on it.

## Enrollment framework (authorizer-agnostic)

- **Config-plugin model.** `enrollment "<type>" "<name>"` is a top-level, two-label block dispatched through the same `(Kind, Type)` plugin registry as `approver`, `credential`, and `endpoint` (a new `KindEnrollment`). An authorizer type implements a small interface: verify a caller's identity and resolve a profile. Additional types slot into the registry without touching the transport, liveness, or dashboard code.
- **Transport.** Enrollment requires a `wireguard` block and issues a concrete WireGuard peer; there is no transport abstraction in v1. The gateway runs the existing userspace WireGuard server, so it needs no kernel WireGuard privileges.
- **State.** Enrollment folds into `wg_peers` (`enrolled=1` plus identity columns: subject, replacement key, authorizer type/name, metadata). The reaper only ever touches enrolled rows, so durable onboarded devices are never reaped. The reaper is the single writer of the liveness trackers; the dashboard reads them under a short read-only lock.
- **The `clawpatrol bridge` sidecar.** A resident privileged sidecar that generates its WireGuard key locally, self-enrolls (`--authorizer=<type>/<name>`), hosts a userspace WireGuard tunnel, routes the whole pod netns through it, and pins the gateway control path off the tunnel so enroll, handoff, and deregister don't loop back through it. It writes only an env file, CA bundle, and ready marker to a shared volume, keeping the private key and peer API token off the workload-visible filesystem, and best-effort deregisters on `SIGTERM`. Its liveness/self-heal behavior is detailed below.
- **Dashboard and API.** Enrolled peers appear in the existing Devices list, identified by their `authorizer/subject` name prefix. The device detail page adds an Enrollment panel: authorizer, subject, enrolled time, keepalive, derived liveness window, and last heartbeat with a missed-beat counter, plus a live/stale indicator. For enrolled peers the profile picker is a read-only chip and the delete action is hidden, since the profile is server-assigned and the peer is reaper-managed. `POST /api/enrollment/register` and its peer-token-authenticated delete drive the lifecycle; a dashboard-authenticated `GET /api/enrollment/peers` and the `enrolled_peers` slice of `/api/state` back the UI.

### Liveness and self-heal

Liveness is observed from the WireGuard device by two independent observers, one per direction:

**Gateway side — the reaper.** The sidecar is the **sole (authoritative) keepalive sender**; the gateway does not keepalive back. That is deliberate: wireguard-go rearms its persistent-keepalive timer on *any* authenticated packet, including received ones, so with both peers sending, one direction's traffic perpetually postpones the other's keepalive — a healthy idle peer can then show flat `rx_bytes` in one direction and be falsely reaped. With a single sender, `client→gateway` `rx_bytes` advances deterministically every interval, and the reaper (~20s loop) revokes a peer only once its rx stays quiet past the derived window.

- Window = `keepalive_interval × keepalive_reap_count` — an integer multiple that can't be misconfigured.
- `keepalive_interval`: default 25s, minimum 10s (pinned to the reaper's sample cadence), no upper bound.
- `keepalive_reap_count`: default 3, `0` disables reaping. `last_handshake` is diagnostic only.

**Client side — the watchdog.** Because the gateway no longer keepalives back, the sidecar's own `rx_bytes` is normally quiet on an idle tunnel — so `rx_bytes` is only a cheap *positive* signal, and when it goes quiet the sidecar actively probes with an **ICMP echo to the gateway's tunnel address** (a round trip proves both directions). It escalates only on sustained failure, where a client controlled `local-reset-missed`  (default to 2) will rekey the existing tunnel, and upon hitting the end of the liveness window, attempt to re-enroll the client.

```mermaid
flowchart LR
    T["watchdog tick"] --> RX{"rx_bytes advancing?"}
    RX -->|yes| H["healthy — no probe"]
    RX -->|quiet| P{"ICMP echo to gateway tunnel IP"}
    P -->|reply| H
    P -->|no reply| C1{"failures ≥ reap_count?"}
    C1 -->|yes| STEP["re-enroll: fresh key, reuse peer IP"]
    C1 -->|no| C2{"failures == local-reset-missed?"}
    C2 -->|yes| RK["in-place rekey"]
```

- **Fail closed during the gap.** With `clawpatrol0` torn down the pod has no default route, so the workload's general egress is blocked (not leaking untunnelled) until the tunnel is rebuilt.
- **Recovers with no default route.** The control-plane host routes (gateway API + WG endpoint + `resolv.conf` resolvers) are pinned to the underlay and tagged with a route protocol (`--route-proto`, default `111`) so they survive; a reconnect re-reads `resolv.conf`, re-discovers the underlay, reconciles the tagged set (adds current, prunes stale), and resolves the gateway by name (TLS SNI / `Host` stay correct). The same tagged-pin recovery covers genuine restarts (OOM, node reboot, eviction).
- **Restart count stays 0.** Recovery never leaves the process, so lifecycle logs — not restart count — are the operator signal.

## The `kubernetes_token_review` authorizer

- **Identity verification.** Verifies the pod's projected ServiceAccount token with the Kubernetes TokenReview API against the block's required `audience`, then reads the live Pod object for the claims it matches on.
- **Profile binding with `match {}`.** Repeatable `match` blocks each express one identity-to-profile rule: gate on `namespace` plus `service_account`, then take the profile from a Pod label (`profile_label`, default `clawpatrol.dev/profile`) whose value must appear in the block's `profiles` allowlist. The gateway derives the profile from the Pod; the client never submits it. A pod whose namespace/service-account matches no block, or whose label value is not allowlisted, is refused.
- **Deployment.** The sidecar runs as a native sidecar (`initContainers` with `restartPolicy: Always`) and needs `NET_ADMIN`, `/dev/net/tun`, a projected ServiceAccount token with the configured audience, Downward-API env (pod name, namespace, uid, node), and a shared `emptyDir` for the handoff. The ICMP liveness probe uses an unprivileged datagram socket, so no extra capability is required. The workload container mounts none of the token or tun, adds no capabilities, and reads the handoff volume read-only; a startup probe on `/clawpatrol/ready` holds it until the tunnel and env/CA handoff are up.
- **RBAC.** The gateway ServiceAccount needs only `create` on `tokenreviews.authentication.k8s.io` and `get` on pods in agent namespaces.
- **Packaging.** A Kustomize base and an e2e overlay live under `examples/kubernetes/`. The explicit pod spec is the supported baseline; auto-injecting the sidecar with a `MutatingAdmissionPolicy` or webhook is an optional, documented layer that nothing in the enrollment path depends on.
- **Egress hardening (optional).** `examples/kubernetes/agent-egress-networkpolicy.yaml` is a recommended NetworkPolicy that pins agent-pod egress to the gateway (API + WireGuard endpoint) and cluster DNS only — defense-in-depth on top of the in-pod fail-closed routing, not a dependency, and effective only on a CNI that enforces NetworkPolicy.

## Config shape

```hcl
gateway {
  dashboard_listen = "0.0.0.0:8080"
  state_dir        = "/opt/clawpatrol"
  wireguard {
    subnet_cidr = "10.55.0.0/24"
    listen_port = 51820
    endpoint    = "clawpatrol-wg.clawpatrol.svc:51820"
  }
}

# top-level, two labels: <type> selects the authorizer plugin, <name> is the instance
enrollment "kubernetes_token_review" "agents" {
  audience = "clawpatrol"

  match {
    namespace       = "agents"
    service_account = "agent-runner"
    profile_label   = "clawpatrol.dev/profile"   # optional; this is the default
    profiles        = ["default"]                 # allowlist the label value must be in
  }
  # additional match { ... } rules allowed

  # Liveness window is keepalive_interval * keepalive_reap_count (25s * 3 = 75s here).
  keepalive_interval   = "25s"   # default 25s, min 10s, no upper bound; the sidecar is the sole keepalive sender
  keepalive_reap_count = 3       # default 3; 0 disables reaping
}

profile "default" { credentials = [] }
```

Agent-pod sidecar (privileged tunnel host; workload runs in a sibling container):

```yaml
initContainers:
  - name: wireguard-sidecar
    restartPolicy: Always
    args:
      - bridge
      - --gateway-url=http://clawpatrol-api.clawpatrol.svc:8080
      - --authorizer=kubernetes_token_review/agents
      - --kubernetes-token-path=/var/run/secrets/tokens/clawpatrol-token
      - --env-out=/clawpatrol/env
      - --ca-out=/clawpatrol/ca.crt
      - --ready-file=/clawpatrol/ready
    securityContext: { allowPrivilegeEscalation: false, capabilities: { add: ["NET_ADMIN"] } }
```

## Scope (v1)

The gateway and agents run in the same cluster, with a single active WireGuard gateway replica, and WireGuard is the only enrollment transport. The sidecar holds the pod-network privileges while the workload container stays restricted.

### Unit tests

- **Config and compile** (`internal/config`): `TestEnrollmentConfigValidation` covers block and `match` validation and rejection of bad values; `TestEnrollmentLivenessDerivation` covers the `keepalive_interval × keepalive_reap_count` window, defaults, the 10s keepalive floor, that a long interval is accepted (no upper bound), the reap-count floor, and `0`-disables; `TestCompileFullSpec` and `TestCompile` cover enrollment blocks compiling into the policy.
- **Registration lifecycle** (`enrollment_test.go`): `TestRegisterEnrolledPeerFresh`, `TestRegisterEnrolledPeerSameSubjectNewKey` (self-heal reuses the IP and swaps the key), `TestRegisterEnrolledPeerPublicKeyConflict`, `TestRegisterEnrolledPeerReplacementTakeover`, `TestRegisterEnrolledPeerRollbackOnPersistFailure` (the WireGuard peer is backed out if the DB write fails), and `TestRegisterEnrolledPeerKeepaliveGating` (the keepalive interval + reap count are returned to bridge enrollments for the client to apply, and omitted otherwise).
- **Reaper and reconcile**: `TestReapStaleEnrolledPeers` (an rx-quiet peer is revoked, a live one is held) and `TestReconcileEnrolledPeersRestoresRuntimeState` (registry and liveness trackers are rebuilt from `wg_peers` after a gateway restart).
- **API endpoints**: `TestApiEnrollmentList`, `TestApiEnrollmentRegisterGuards`, `TestApiEnrollmentDeleteRequiresEnrollment`, and `TestApiEnrollmentDeleteWithEnrollment` (only a peer that holds an enrollment can drive teardown via its peer token).
- **kubernetes_token_review authorizer** (`k8s_registration_test.go`): `TestKubernetesTokenReviewAuthorizerIdentity` (a TokenReview result maps to an identity), `TestK8sServiceAccountAndAllowlist` (namespace/service-account gate plus profile allowlist, including rejection), `TestNormalizeWGPublicKey`, and `TestCleanupEnrolledPeer`.
- **Client watchdog + bridge** (`wg_watchdog_test.go`, `bridge_test.go`): `TestWatchdogProbeEscalates` (rx-quiet → probe → in-place rekey → in-process reconnect), `TestWatchdogRxHealthyNoProbe` (advancing rx is liveness on its own — no probe is generated), `TestWatchdogProbeSuccessIdle` (a replying probe on an idle tunnel never escalates), `TestWatchdogInertWhenReapingDisabled`, `TestWatchdogResetMisses` (threshold resolution and clamping to the server reap count), `TestParsePeerStats` / `TestParsePeerStatsNoHandshake`, and `TestValidateRouteProto` (the `--route-proto` escape hatch).

### End-to-end (`e2e/kubernetes-wireguard-e2e.sh`, kind)

Builds the workspace image, loads it, applies the Kustomize base plus the e2e overlay, and asserts:

- the sidecar writes `ready`/`env`/`ca.crt`, the CA is a real certificate, and the pushed-down env points `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS` back into the handoff dir;
- the restricted-agent contract: no ServiceAccount token, no `/dev/net/tun`, no added capabilities, no route edits, no secret material on the handoff volume;
- the tunnel data path relays a TCP request through `clawpatrol0`, the gateway control path is pinned off the tunnel, and the control-plane pin carries the route-protocol tag;
- the enrolled `wg_peers` row exists with a peer IP;
- `rx_bytes` liveness holds a live peer past its window without being reaped;
- in-process fail-closed self-heal: scaling the gateway to 0 makes the sidecar's ICMP probe fail; it tears the tunnel down leaving **no** default route (with the tagged control-plane pins intact) and, with the container restart count **unchanged**, re-enrolls in process with a fresh key at the reused IP once the gateway returns, with the tunnel restored;
- cleanup: a force-deleted pod is reaped after its window, and graceful deletion deregisters immediately, with the WireGuard peer revoked either way.

### Behaviours covered end to end

- A pod enrolls and comes up with the tunnel already routing; the workload container stays unprivileged.
- The profile is taken from the Pod label and refused if not allowlisted; a mismatched namespace or service account is refused.
- A live pod stays enrolled; a dead pod is reaped; a graceful shutdown deregisters immediately.
- A peer that loses liveness self-heals in process (no container restart) without operator involvement and keeps its identity (reused IP, fresh key), while the workload's egress fails closed during the gap.
- Enrolled peers are visible and legible in the dashboard, with controls that don't apply to reaper-managed peers suppressed.

### Gates

`go test ./internal/config/... ./cmd/clawpatrol/`, `gofmt`, `go vet`, docgen-drift, and the dashboard `tsc` / `oxfmt` / `oxlint` checks are green; `examples/` and `e2e/` HCLs validate and `kustomize build` succeeds for the base and the overlay. Validated end-to-end on a kind cluster, including the in-process self-heal path.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+!Dockerfile
+!clawpatrol
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,13 +68,8 @@ jobs:
       - run: make fmt-check
       - run: cd dashboard && deno task lint:github
       - run: make test
-      - name: validate HCL examples
-        run: |
-          go build -o /tmp/clawpatrol ./cmd/clawpatrol
-          for f in examples/*.hcl; do
-            echo "validating $f"
-            /tmp/clawpatrol validate "$f"
-          done
+      - name: validate example + e2e configs
+        run: make validate-examples
 
   # Exercise the Landlock fallback backend: keep AppArmor's userns
   # restriction in place (the runner default) and force the backend so
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ macos/netstack/libwgnetstack.h
 macos/netstack/wgnetstack
 *.swp
 /plugin-example/plugin-example
+e2e/.e2e-runtime.*/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# Runtime image for a prebuilt clawpatrol binary.
+#
+# Build a Linux binary first, then build the image from the same directory:
+#   CGO_ENABLED=0 GOOS=linux GOARCH=amd64 make build
+#   docker build -t clawpatrol:dev .
+FROM debian:stable-slim
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    curl \
+    iproute2 \
+    netcat-openbsd \
+    sqlite3 \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY clawpatrol /usr/local/bin/clawpatrol
+
+ENTRYPOINT ["/usr/local/bin/clawpatrol"]
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,23 @@
-.PHONY: build release dashboard test fmt fmt-check lint go-lint dashboard-lint clean install
+.PHONY: build release dashboard test fmt fmt-check lint go-lint dashboard-lint validate-examples e2e clean install
 
 build: dashboard
 	go build -o clawpatrol ./cmd/clawpatrol
+
+# validate-examples compiles every example/e2e gateway config. These HCLs
+# are the source of truth for the Kubernetes ConfigMaps (generated via
+# kustomize configMapGenerator), so validating them keeps "what CI checks"
+# equal to "what deploys" — not just a standalone sample.
+validate-examples: dashboard
+	@set -e; bin="$$(mktemp)"; trap 'rm -f "$$bin"' EXIT; \
+	go build -o "$$bin" ./cmd/clawpatrol; \
+	git ls-files '*.hcl' | grep -E '^(examples|e2e)/' | while read -r f; do \
+		echo "validate $$f"; "$$bin" validate "$$f"; \
+	done
+
+# e2e drives the Kubernetes WireGuard enrollment flow against a local kind
+# cluster (see e2e/kubernetes-wireguard-e2e.sh for prerequisites + knobs).
+e2e:
+	./e2e/kubernetes-wireguard-e2e.sh
 
 # release mirrors .github/workflows/release.yml: strip the symbol
 # table and DWARF (-s -w) and rewrite source paths (-trimpath). On
--- a/cmd/clawpatrol/bridge.go
+++ b/cmd/clawpatrol/bridge.go
@@ -1,0 +1,164 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"log"
+	"net/netip"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// kubernetesDefaultTokenPath is the projected ServiceAccount token the
+// kubernetes_token_review provider presents by default.
+const kubernetesDefaultTokenPath = "/var/run/secrets/tokens/clawpatrol-token"
+
+// bridgeOptions configures `clawpatrol bridge`: a resident, foreground,
+// privileged data plane that self-enrolls through an authorizer, hosts a
+// userspace WireGuard tunnel, and routes the whole network namespace through
+// the gateway — bridging the netns to the gateway. It stays up for the netns
+// lifetime (the userspace device dies with the process) and deregisters
+// best-effort on SIGTERM. The execution workload runs in a sibling,
+// unprivileged container that shares this netns and reads the handoff files
+// below.
+type bridgeOptions struct {
+	GatewayURL string
+
+	// AuthorizerType/Name mirror the gateway's two-label enrollment block
+	// `enrollment "<type>" "<name>"`. Type selects the client-side claims
+	// provider (how identity is gathered); Name is what's sent on the wire
+	// to pick the server authorizer.
+	AuthorizerType string
+	AuthorizerName string
+
+	// KubeTokenPath is the kubernetes_token_review provider's credential.
+	// Provider-scoped on purpose — not every authorizer authenticates with
+	// a token file.
+	KubeTokenPath string
+
+	// Sibling-handoff outputs: files the unprivileged workload container
+	// reads off the shared volume once the tunnel is up.
+	EnvOut    string
+	CAOut     string
+	ReadyFile string
+
+	Iface string
+	MTU   int
+
+	// LocalResetMisses is how many missed keepalives trigger an in-place
+	// tunnel rebuild before the server-dictated reconnect horizon. A
+	// client-side knob — the reap horizon comes from the gateway, but how
+	// eagerly to attempt a cheap local recovery first is the sidecar's call.
+	// Capped below the reconnect horizon; 0 disables the local rebuild.
+	LocalResetMisses int
+
+	// RouteProto is the rt_proto tag stamped on the control-plane host routes
+	// the bridge pins to the underlay, and the filter used to recover them on
+	// an in-process reconnect. Escape hatch: override only if the default collides
+	// with a route protocol another component already uses in the netns.
+	RouteProto string
+}
+
+// defaultRouteProto is the rt_proto the bridge tags its underlay pins with by
+// default (overridable via --route-proto). Arbitrary and unregistered; it only
+// has to be stable and unlikely to collide with the CNI's own routes.
+const defaultRouteProto = "111"
+
+// runBridge parses the `clawpatrol bridge` flags and dispatches to the
+// platform implementation. Kept cross-platform (no wireguard imports) so
+// the flag surface and validation errors are identical everywhere; bridgeRun
+// is linux-only.
+func runBridge(args []string) {
+	fs := flag.NewFlagSet("bridge", flag.ExitOnError)
+	var (
+		authorizer string
+		opt        bridgeOptions
+	)
+	fs.StringVar(&opt.GatewayURL, "gateway-url", "", "gateway API URL, e.g. https://clawpatrol.clawpatrol.svc:8443")
+	fs.StringVar(&authorizer, "authorizer", "", "enrollment authorizer to register through, as <type>/<name> (e.g. kubernetes_token_review/agents)")
+	fs.StringVar(&opt.KubeTokenPath, "kubernetes-token-path", kubernetesDefaultTokenPath, "kubernetes_token_review: projected ServiceAccount token path")
+	fs.StringVar(&opt.EnvOut, "env-out", "/clawpatrol/env", "path to write shell exports for the workload container")
+	fs.StringVar(&opt.CAOut, "ca-out", "/clawpatrol/ca.crt", "path to write the gateway CA bundle")
+	fs.StringVar(&opt.ReadyFile, "ready-file", "/clawpatrol/ready", "path to touch after network and env setup succeed")
+	fs.StringVar(&opt.Iface, "iface", "clawpatrol0", "TUN interface name")
+	fs.IntVar(&opt.MTU, "mtu", enrollmentDefaultMTU, "TUN MTU")
+	fs.IntVar(&opt.LocalResetMisses, "local-reset-missed", bridgeWatchdogDefaultResetMisses,
+		"missed keepalives before an in-place tunnel rebuild; 0 disables it, and a value at or above the gateway's reconnect threshold skips directly to re-enrollment")
+	fs.StringVar(&opt.RouteProto, "route-proto", defaultRouteProto,
+		"rt_proto tag for the underlay control-plane pins (gateway + DNS); override only if it collides with another component's routes")
+	_ = fs.Parse(args)
+
+	if len(fs.Args()) > 0 {
+		fail("clawpatrol bridge does not take a command; the workload runs in a sibling container sharing this netns")
+	}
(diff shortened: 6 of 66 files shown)


piscisaureus (changes requested):
Thank you for carrying #709 through to a complete implementation, and for the kind e2e. I reviewed it against current main: it builds, the gateway and config suites pass, docgen is fresh, every example validates, and it rebases with one trivial test-file conflict. The layering (config plugin, compiled enrollment envelope, runtime authorizer, reaper) matches what we agreed in #709 on every point that matters, and the liveness design (observe WireGuard rx, no lease, client-only keepalive) is better reasoned than the RFC's heartbeat. Two small divergences worth a sentence in the PR: the runtime side dispatches on `enrollment.Type` in a switch rather than a registry (fine for v1), and `newWGTransportFromConf` is split out but nothing calls it (drop it).

Four things need to change before this lands:

1. **An authorized pod can knock a durable device off WireGuard.** `findEnrolledPeers` only scans `enrolled = 1`, so registering with an onboarded device's public key (visible on the dashboard, not secret) goes through `allocatePeer` → `addPeerLocked`, whose `ON CONFLICT(pubkey) DO UPDATE` re-points the durable row to a fresh IP, and `markEnrolledPeer` stamps it `enrolled = 1`. The real owner's packets then fail allowed-IPs and the reaper deletes the row. Before allocating, look the key up across all of `wg_peers`; any existing non-enrolled row is a 409. Please add a test for it.
2. **The unauthenticated register path is cheap to abuse.** Every POST with any bearer string builds a fresh `http.Transport` plus `x509.SystemCertPool()`, reads two files, and fires a TokenReview with the gateway's service account; transports are never closed. Flooding it throttles the gateway SA at the apiserver and starves real enrollments. Cache the transport and cert pool (re-read only the token), and put a small concurrency limit on the register handler. Subnet exhaustion is bounded by distinct pod UIDs with valid bound tokens, which is fine, but the docs should say what a `/24` gives you.
3. **A probe socket failure looks like liveness loss.** `pingGatewayTunnel` needs `net.ipv4.ping_group_range` to include the sidecar's GID; on containerd before 2.0 without `enable_unprivileged_icmp` (and some CRI-O setups) `ListenPacket` returns EPERM. The watchdog counts that as a failed probe, so a healthy tunnel is rekeyed, torn down and re-enrolled every window, forever, with egress failing closed each cycle. Open the socket once at bring-up and fail loudly, or treat socket-open errors as "probe unavailable" rather than escalating, and document the sysctl (or `NET_RAW`) in the Kubernetes page.
4. **Tests for the invariants the design rests on.** `TestReapStaleEnrolledPeers` only reaps the backdated peer; the description says it also checks a live one is held, but it doesn't. Add: a live peer survives, a non-enrolled row survives the reaper, and the case in point 1. The TokenReview client tests (audience, pod binding, mismatch) are good.

Worth doing in the same PR, not blocking: ship a gateway-ingress NetworkPolicy next to the agent-egress one and tell operators to run `--set-dashboard-password` before applying (the example exposes `0.0.0.0:8080` cluster-wide, and first-run has an open set-password form); prefer the free IP with the oldest `added_ns` so a recycled IP does not inherit the previous pod's request history in the UI; add a `MethodDelete` row to the route table for the DELETE the handler dispatches itself; log a warning when a same-subject registration evicts a live key; and a sentence on the one-way-stream liveness edge case (rx suppressing keepalives) since the design note itself predicts it.

Once 1 to 4 are in I will rebase, resolve the test-file conflict, and land it as one commit with you as author.

piscisaureus:
A second, independent review pass turned up five more items; adding them here so they land in the same revision.

- **Host-network pods must be excluded from bridge injection** (`examples/kubernetes/agent-bridge-mutatingadmissionpolicy.yaml`). Injected into a `hostNetwork: true` pod, the bridge shares the node's network namespace and its `NET_ADMIN` lets `replaceDefaultRoutes` rewrite the node's default route. Add a match condition that skips host-network pods.
- **Readiness is published before the data path works** (`bridge_linux.go`, the ready-file write after `dev.Up`). If the registration API is reachable but the WireGuard UDP endpoint is blocked or misadvertised, the ready marker is still written, the startup probe passes, and the workload starts with a dead default route. Require a handshake or a successful tunnel probe first.
- **IPv6 addresses collide beyond the first /24** (`wireguard.go` allocator). With a `subnet_cidr` wider than `/24`, allocation now reaches the second `/24`, but `wg6FromV4` encodes only the last octet and `canonicalPeerIP` rebuilds the first three from the gateway subnet, so `10.55.1.2` and `10.55.0.2` both map to `fd77::2`. Either cap allocation at the representable `/24` or extend both mappings to all host bits.
- **Validation accepts an enrollment config with no reachable registration listener** (`internal/config/k8s_enrollment.go`). With `dashboard_listen` unset or loopback-only, an unenrolled bridge has nothing to register against. Require a routable API listener when enrollment is enabled.
- **Gateway-provided CA variables can override the client-local ones** in the exported env file (`bridge_linux.go`, the env export). Apply the same `dropClawpatrolCAVars` filtering the other pushdown paths use so a provider-emitted `SSL_CERT_FILE` or `NODE_EXTRA_CA_CERTS` cannot displace the generated bundle.

The first three are must-fix alongside the four in the review above.

PR #798, opened 2026-08-09
fix(kubernetes): select credentials by bearer placeholder

rajiteh:
## Context

Kubernetes endpoints did not expose a placeholder detector, so profiles with multiple credentials bound to one cluster could not select a bearer credential. The placeholder reached the API server unchanged, and credential-pinned rules could not observe the intended credential.

This change adds exact Kubernetes bearer-placeholder detection. The selected credential is resolved before policy matching and reused for upstream injection.

## Risks

The detector accepts only an exact `Authorization: Bearer <placeholder>` value. Unknown or missing placeholders remain unmatched. Singular credentials, explicit fallback credentials, and the existing HTTPS detector keep their current behavior.

## Validation

- Focused endpoint detector tests
- Kubernetes MITM regression for reader and admin placeholders
- Credential-pinned approval regression
- Unknown, missing, fallback, and overlapping-placeholder coverage
- `go test ./...`
- `gofmt -l .`
- Dashboard format and lint checks
- golangci-lint
- HCL example validation

### DIFF
--- a/cmd/clawpatrol/kubernetes_credential_dispatch_test.go
+++ b/cmd/clawpatrol/kubernetes_credential_dispatch_test.go
@@ -1,0 +1,241 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"crypto/tls"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/denoland/clawpatrol/internal/config"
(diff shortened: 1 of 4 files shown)


piscisaureus:
Landed on main as 9c54be1 with you as author. Reviewed the dispatch path end to end: the detector is only consulted for multi-credential profiles, candidates are profile-scoped so cross-profile selection isn't possible, and the exact-match semantics are stricter than the https substring scan in a good way. Thanks for the fix and the thorough tests.

PR #800, opened 2026-08-17
fix: make HTTP body capture completion-aware

magurotuna:
## Summary

- synchronize body sampling and record complete, incomplete, or aborted state
- emit a full-body SHA only when the body is complete
- preserve denied and partial-read request captures
- reject fixtures built from incomplete, lossy, redacted, or transformed bodies
- show partial capture state in the dashboard

## Why

An early response could snapshot a body while it was still being written. Fixture export could also treat a partial or changed body as the original request.

## Tests

- `go test ./...`
- focused race tests
- dashboard tests, format, lint, and build

### DIFF
--- a/cmd/clawpatrol/credential_match_test.go
+++ b/cmd/clawpatrol/credential_match_test.go
@@ -85,9 +85,24 @@ profile "default" {
 	// `http.method != 'GET'` condition, so it still falls through to the
 	// default deny. This guards against a "fix" that ignores the condition.
 	t.Run("condition still gates the pinned rule", func(t *testing.T) {
+		events, cancelEvents := h.gateway.sink.Subscribe()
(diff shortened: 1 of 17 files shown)


piscisaureus:
Landed on main as c883c85 with you as author. Reviewed with two passes; folded four follow-ups into the same commit rather than bounce it back after a month: a truncated text body cut inside a multibyte rune keeps its text preview (the utf8.Valid check turned those into hex dumps under the default 4 KiB cap), rows with no recorded state show no badge, the fixture-export error renders under the button, and a request is marked transformed whenever injection edits its body or URL (Slack and Telegram rewrite without implementing HTTPRequestRewriter, so their captures were exportable as the original). Thanks for the careful sampler work; the deterministic lock-hold test was a nice touch.

PR #802, opened 2026-09-08
gateway: add unknown_host inspect MITM for unmatched SNI

sverdlov93:
## Summary

Extends #782. Rebased onto current `main` (includes v0.5.10 / #838).

- MITM unmatched HTTPS SNI as declared `endpoint "https" "unknown"` when `defaults.unknown_host = "inspect"`
- keep named-endpoint dispatch, `passthrough` splice, and `deny` hang-up unchanged
- compile-error if inspect is set without type `https` `"unknown"`; hang up at runtime if that endpoint is missing
- reject any credential bound to `https.unknown` (a leftover-host bearer would inject on every unmatched SNI)
- rules on `https.unknown` require `unknown_host = "inspect"`; an endpoint of another type named `"unknown"` is unaffected
- UDP/443 is refused for every destination on `main` (#838); inspect no longer has its own QUIC branch

```hcl
defaults { unknown_host = "inspect" }
endpoint "https" "unknown" { hosts = [] }
```

## Why

`passthrough` splices with no URL. `deny` hangs up at SNI. Neither can allow one path on an unknown host and deny another.

## Verification

- `go test ./cmd/clawpatrol -run '^TestSNIDispatch(UnknownHostPolicy|UnknownHostInspect)|TestInspect(UnknownHostRules|NamedHostWins)$' -count=1`
- `go test ./internal/config -run '^TestCompileUnknown' -count=1`
- `go test ./cmd/clawpatrol -run '^TestTsnetUDPDisposition$' -count=1`
- `go test -race ./cmd/clawpatrol -run '^TestSNIDispatchUnknownHost(Policy|Inspect)' -count=10`
- `gofmt` and `git diff --check`


piscisaureus (changes requested):
Thanks, this is a real gap: `deny` and `passthrough` give no way to allow one path on an unknown host and deny another, and the dispatch here is right. I traced it end to end: the minted leaf carries the SNI as SAN, the forward path re-resolves the SNI host with normal upstream cert verification, and `hosts = []` is harmless because the forward path never reads it. Tests drive the real `handle()` path. Two things need to change before this lands, and two follow-ups are worth knowing about.

**Must fix**

1. `internal/config/compile.go` `validateUnknownInspect`: reject credentials bound to `https.unknown`. Today this compiles:
   ```hcl
   credential "bearer_token" "tok" { endpoint = https.unknown }
   profile "default" { credentials = [bearer_token.tok] }
   ```
   and `ResolveCredential` returns the single entry unconditionally, so the secret is injected into every unmatched host the agent picks. The endpoint doesn't need profile membership at all (dispatch fetches it from `policy.Endpoints` directly), so `len(ce.Credentials) > 0` on it should be a compile error.

2. Same function, the `rules on https.unknown require inspect` branch: `ok` is a bare name lookup and endpoint names are unique per kind, not per type, so an existing `endpoint "openai_codex_https" "unknown"` with rules and `unknown_host = "passthrough"` now fails to compile with a misleading message. Gate that branch on `ce.Plugin.Type == "https"` like the inspect-set branch already does.

**Follow-ups (not blocking, happy to take them as separate PRs from you or do them myself)**

3. There is no `http.host` facet, so rules on `https.unknown` can't say "allow `*.npmjs.org`, deny the rest", which is the obvious use of inspect. `http.headers["Host"]` is empty too because `http.ReadRequest` moves it to `req.Host`. Plumbing the SNI host into the https facet as `http.host` would make this feature what it wants to be.

4. `CertCache.mint` caches one key+leaf per name with no eviction. With inspect the name space is agent-chosen. This is pre-existing (wildcard endpoints have the same shape) but inspect widens it; an LRU bound would close it.

Small nits: `main.go` `handle()` uses the literal `"unknown"` where `config.UnknownInspectEndpoint` exists; `TestInspectUnknownHostRules` reuses a harness whose dialer ignores the address, so it doesn't actually pin that the upstream dial goes to `<sni>:443` (recording the dialed addr and asserting it would make the test load-bearing); `examples/gateway.example.hcl` could list `inspect` next to the other values.

If you push 1 and 2 I'll land it.

sverdlov93:
@piscisaureus 1 and 2 are in `bad74e7` (plus the three nits).

- `https.unknown` with any credential bind is a compile error
- rules-require-inspect is gated on type `https`, so `openai_codex_https` `"unknown"` + passthrough still compiles
- `handle()` uses `UnknownInspectEndpoint`; inspect dial test asserts `<sni>:443`; example HCL lists `inspect`

CI is green on this head (lint, test, sandbox landlock + macos). 3 (`http.host`) and 4 (CertCache LRU) left as the follow-ups you offered.

sverdlov93:
@piscisaureus 1 and 2 are in `bad74e7` (plus the three nits).

- `https.unknown` with any credential bind is a compile error
- rules-require-inspect is gated on type `https`, so `openai_codex_https` `"unknown"` + passthrough still compiles
- `handle()` uses `UnknownInspectEndpoint`; inspect dial test asserts `<sni>:443`; example HCL lists `inspect`

CI is green on this head (lint, test, sandbox landlock + macos). 3 (`http.host`) and 4 (CertCache LRU) left as the follow-ups you offered.

piscisaureus:
Re-reviewed bad74e7: items 1 and 2 are exactly as described and the nits are in, thanks. I squashed the branch into one commit attributed to you and pushed it to your branch, with two small maintainer fixups folded in that the second pass turned up:

- a rule on `https.unknown` with `credential = X` compiled but could never fire (no credential can bind there, so `ResolveCredential` is nil and the rule is skipped); it is now a compile error like the binding itself
- `rules on https.unknown require inspect` counted `disabled = true` rules, so switching back to `passthrough` failed until the rule was deleted; only enabled rules count now

plus a paragraph in the security model and the config README noting that `inspect` decrypts and body-captures traffic that `passthrough` forwards untouched, and that `hosts = []` on `https.unknown` is required but never consulted.

The two follow-ups are now #849 (`http.host` facet; the dashboard Block button on an inspected request currently generates a rule that blocks that path on every unknown host, which makes it more pressing than I first thought) and #850 (cert cache bound and re-mint). Landing once CI and the bot review are green.

PR #819, opened 2026-09-10
config: keep <<file:...>> includes inside the config directory

### LINKED ISSUE #315: Constrain `<<file:...>>` includes to the config directory
## TL;DR
`<<file:...>>` expansion accepts absolute paths and relative paths that can traverse out of the config directory. A dashboard-authenticated config writer can make the gateway read files outside the intended config tree.

## Why this matters
The gateway often runs with more filesystem privilege than a dashboard operator should exercise. File include syntax is useful for certs/large literals, but it should not become a general privileged file-read primitive.

## Current code/data flow to inspect
- `config/file_include.go`: `resolveFileIncludes` extracts `name` from `<<file:NAME>>`.
- If `filepath.IsAbs(name)` is true, the path is read as-is.
- Otherwise it uses `filepath.Join(configDir, name)` without checking that the cleaned result remains under `configDir`.
- Current opt-in surface is narrow; for example `config/plugins/endpoints/kubernetes.go` opts `ca_cert` into file includes.

## Preconditions / impact
- Attacker has dashboard/config-write permission or can otherwise submit/modify `gateway.hcl`.
- Gateway process can read a sensitive file outside the config directory.
- Expanded file bytes may be surfaced via validation errors, dashboard config output, emitted HCL, or used as unintended trust material.

## Suggested fix shape
- Reject absolute paths in `resolveFileIncludes`.
- Clean relative paths and ensure the final resolved path is under `configDir`.
- Consider allowing an explicit configured include root later, but default should be config-local.
- Ensure diagnostics are helpful without leaking file contents.

## Acceptance criteria
- [ ] `<<file:/absolute/path>>` is rejected with a config diagnostic.
- [ ] `<<file:../outside>>` and symlink/clean-path escapes are rejected or safely constrained.
- [ ] Legitimate config-local includes still work.
- [ ] Error messages do not include file contents.

## Regression tests to add
- Unit tests for absolute path, `..` traversal, normal relative include, nested config-local include.
- If symlink policy is defined, include a symlink escape test.

---

Context: this came from a third-party Claw Patrol penetration-test triage and was re-checked against `origin/main` around `c65020e07e61bf65e0e2b1191c2cb5b86adf553b`. Please avoid putting real credentials/tokens in tests, logs, screenshots, or issue comments; use placeholder values such as `[REDACTED]`.

piscisaureus:
`resolveFileIncludes` accepted absolute paths and relative paths with
`..` traversal, so anyone who can write `gateway.hcl` (the dashboard
config editor included) could make the gateway read any file its
process can see and surface the bytes through diagnostics, the emitted
config, or trust material.

* Reject absolute names and any path that resolves outside the config
  directory, with a diagnostic naming the reason.
* Names inside the directory, including `sub/../ca.pem`, keep working.
* Test covers accepted and rejected shapes and checks no file content
  leaks into the output on rejection.

Fixes #315

### DIFF
--- a/internal/config/file_include.go
+++ b/internal/config/file_include.go
@@ -63,8 +63,11 @@ type FileIncludeField struct {
 
 // resolveFileIncludes scans s for `<<file:NAME>>` markers and
 // substitutes each with the contents of NAME read relative to
-// configDir. Multiple markers in one string are supported. Markers
-// with absolute paths are read as-is (no configDir join).
+// configDir. Multiple markers in one string are supported. NAME must
+// stay inside configDir: absolute paths and `..` traversal are
+// rejected, so a config writer (the dashboard editor included) cannot
+// turn the include syntax into a read of arbitrary files the gateway
+// process can see.
 func resolveFileIncludes(s, configDir, entityName string, blockRange hcl.Range) (string, hcl.Diagnostics) {
 	if !strings.Contains(s, "<<file:") {
 		return s, nil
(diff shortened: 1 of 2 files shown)


piscisaureus:
Codex flagged that a symlink inside the config directory can still point outside it. That is deliberate: the threat in #315 is a config writer (the dashboard editor), not someone who can place links in the config directory, who already has the gateway user's filesystem access, and operators do symlink certificates next to the config. Added a comment saying so.

PR #822, opened 2026-09-10
wireguard: derive the client MTU from the path, add a gateway mtu knob

### LINKED ISSUE #790: wg-over-tailscale: gateway sendmmsg EMSGSIZE black-holes transfers over ~16KiB, with no MTU lever
## Problem

With a gateway reached over Tailscale, every transfer through the tunnel runs at roughly 0.6–1.5 KB/s and any response larger than about 16KiB never completes. Small requests succeed, so the deployment looks healthy until something tries to move real data.

The gateway logs this continuously while traffic flows:

```
ERROR: [wg] peer(...) - Failed to send data packets: write udp 0.0.0.0:51820: sendmmsg: message too long
```

`EMSGSIZE` here is WireGuard forming datagrams larger than the Tailscale path can carry: the Tailscale interface is MTU 1280, while wg's tun defaults to 1420. Every full-size packet is dropped at the socket, and nothing adapts.

## Measurements

Same URL, same client, with and without `clawpatrol run`. Both hosts are **unbrokered** (no endpoint declared, no credential injected), so neither the MITM path nor policy evaluation is involved:

| request | bytes | direct | through gateway |
| --- | --- | --- | --- |
| a JSON API response | 6,868 | 0.26s (26.7 KB/s) | 6.60s (1.0 KB/s) |
| a raw file over HTTPS | 3,807 | 0.23s (16.7 KB/s) | 5.91s (0.6 KB/s) |

A size sweep stalls at a fixed ceiling rather than degrading gradually. Three different objects stopped at ~15.59KB, and one JSON response stopped at exactly 16,384 bytes:

| object | bytes received | outcome |
| --- | --- | --- |
| 7.7KB thumbnail | 7,731 | completed in 4.9s |
| 20KB thumbnail | 20,449 | stalled at 20s timeout |
| larger thumbnails (x2) | 15,587 / 15,594 | stalled |
| 1.3MB file | 15,593 | stalled |
| JSON listing | 16,384 | stalled |

## Why the error rate is misleading

The `sendmmsg` errors arrive at a flat ~3.5/second regardless of what the traffic is doing — 36 errors across five small API calls versus 68 across twenty seconds of bulk download. That flatness reads as harmless background noise, which is how this went unnoticed through several rounds of functional testing. It is flat because it is paced by TCP retransmission backoff, not by transfer volume.

This is the same size-dependent black hole class described in #787, but on the IPv4 full-tunnel path with a WireGuard-over-Tailscale transport, and with the drop occurring on the gateway's own send path rather than on an unhandled ICMP signal.

## What we tried

- `MTU = 1200` in the client's `wg.conf` `[Interface]` — no effect. Per-process mode uses a userspace netstack rather than bringing up the kernel interface via `wg-quick`, so the file's MTU line appears not to be consulted.
- Searching the config schema for an MTU attribute: `mtu` is rejected in `gateway`, `gateway.wireguard`, `defaults` and `endpoint` blocks, and does not appear among the binary's HCL tags.
- Searching for an environment override: no `CLAWPATROL_*` variable exposes an MTU.
- Capping stored response bodies (`limits.body_storage`), in case audit persistence was the bottleneck — no change, which is what ruled that out.

So there appears to be no way to configure around this from the deployment side.

## Suggested directions

Any one of these would unblock it:

1. A configurable WireGuard MTU in the `gateway.wireguard` block, pushed to clients through the config channel so both ends agree.
2. Automatically clamping the tun MTU when the transport is tsnet/Tailscale, where 1280 is the known path MTU.
3. Honoring `wg.conf`'s `MTU` in per-process mode, so operators at least have a client-side lever.

Happy to test a patch or collect more data — the reproduction is reliable and takes seconds.

## Environment

- Claw Patrol 0.5.8
- Gateway: Linux x86_64, kernel 6.19, systemd unit, userspace WireGuard on udp/51820, reached over Tailscale
- Client: Linux, per-process mode (`clawpatrol run`), joined without `--whole-machine`
- Tailscale interface MTU 1280 on both ends; direct (non-relayed) path between them

piscisaureus:
Both ends of the WireGuard tunnel used a fixed 1420-byte MTU. When
the gateway is reached over Tailscale (a 1280-byte interface) every
full-size packet is dropped on the sender's socket with `sendmmsg:
message too long`, nothing adapts, and transfers over about 16 KiB
never complete while small requests keep working (#790, measured 57x
slower by the reporter).

* Linux daemon: derive the MTU the way `wg-quick` does, from the MTU
  of the interface the kernel routes the gateway endpoint through
  (`ip -o route get`), minus the 80 bytes of framing, clamped to
  [1000, 1420]. The value and its source are logged.
  `CLAWPATROL_WG_MTU` overrides.
* Gateway: new `wireguard { mtu = N }` attribute (default 1420, range
  1000 to 1500) for the gateway's own side, since it cannot know the
  path to each peer. Documented in configure-gateway.md with the
  Tailscale value of 1200 and the symptom to look for; config
  reference regenerated; example config annotated.
* The macOS extension keeps 1420; deriving there is a follow-up.

Verified: unit tests for the derivation, route parsing and override;
`GOOS=linux go vet`; example config validates. End-to-end check on a
Linux client against a Tailscale-reached gateway still to do.

Fixes #790


piscisaureus:
Converting to draft. A second review pass found that gVisor's IPv6 stack refuses a link MTU below 1280 (`header.IPv6MinimumMTU`), so the 1200 this PR derives for a Tailscale path would silently disable in-tunnel IPv6 (the fd77 address, `::/0`, and the fd78 DNS VIPs) on both the client stack and the gateway. Kernel WireGuard behaves the same way when `wg-quick` sets 1200, which is why the reporter's manual test worked for IPv4, but the userspace stacks here carry IPv6 traffic we want to keep.

The alternative worth testing is to let the kernel fragment the outer UDP datagram (`IP_PMTUDISC_DONT` on the WireGuard socket) instead of shrinking the inner MTU: no config knob, IPv6 intact, works for any narrow path. wireguard-go does not expose a socket-control hook, so that needs either a small fd-level setsockopt after the device is up or a vendored bind. Either way it needs a Linux client against a Tailscale-reached gateway to verify, which I don't have at hand right now. The `mtu` knob and the emit/dump/validation gaps from the review are on hold until that is settled.

piscisaureus:
Withdrawn in favour of the real fix. Testing on a Linux host showed the MTU is not the cause: wireguard-go's UDP GSO batches are refused by the kernel when a segment exceeds the route MTU, and wireguard-go only backs off on EIO. Sending one datagram per syscall after the first EMSGSIZE fixes it with no MTU change and no config knob; see the follow-up PR on #790.

Issue #811, opened 2026-09-09
CLI for configuring a remote gateway (credentials, device approval, profiles)


piscisaureus:
Today everything that is not in `gateway.hcl` has to be done through the dashboard: setting credential secrets, connecting OAuth credentials, approving a joining device, assigning a profile. That does not fit scripted or fleet setups, and it does not fit an agent provisioning another agent. #482 prototyped `approve` / `oauth-start` / `oauth-exchange` subcommands and was closed to rethink the shape.

Proposal: an `admin` group of subcommands that talks to a gateway's existing dashboard API from anywhere except an agent context.

```
clawpatrol admin [--server URL] [--username NAME] [--password PW] <verb>

  approve <code> [--profile NAME]        approve a pending join code
  set-profile <device> <profile>         reassign a device's profile
  credential set <id> --slot K=V ...     set static credential slots
  credential connect <id>                interactive OAuth start + exchange
  credential clear <id>                  drop a credential's stored material
```

Constraints:

- Auth reuses what the dashboard already has: the root password (prompted on a TTY, or `CLAWPATROL_PASSWORD`), or the tailnet operator allowlist when reached over Tailscale. Nothing is stored on the client machine; in particular no long-lived admin key.
- Verbs are generic over credential types (`set --slot` covers static tokens; `connect` covers OAuth) rather than one verb per flow.
- Rules, endpoints and profile definitions stay in HCL. This is for the mutable state the dashboard already owns, not a second config format.
- Keep the verb set small; approvals of HITL operations and config push can come later if there is demand.

Open questions: how the `admin` group is presented next to the gateway-side (`gateway`, `validate`, `test`) and agent-side (`join`, `run`, `env`, `status`) commands in `--help`, and whether `--server` should default from a joined device's gateway URL.

piscisaureus:
Design note is up as #832 (`doc/admin-cli-design.md`). Short version: six verbs over the existing dashboard endpoints, per-invocation root-password auth with the session cookie kept in memory and logged out at exit, nothing stored on the client, `--server` defaulting to the URL `join` saved, `--json` for scripts. Three open questions at the end: device addressing (hostname/IP vs a stable id), whether `CLAWPATROL_PASSWORD` should exist alongside `--password-file`, and whether `approve` should accept a hostname as well as the code. Comments welcome on the PR.

PR #625, opened 2026-06-02
feat: clawpatrol apply — config version store, semantic diff, history

littledivy:
## What

`clawpatrol apply` — a Terraform-style config workflow. The **DB is the authoritative state backend**, every change takes a **lock**, and writes check a **serial**, so concurrent applies and stale plans can't clobber each other.

- **State backend** = `config_versions`; the latest row is the deployed config, its `id` is the **serial**.
- **The gateway runs from the backend**, not a watched file. `gateway <file>` only *seeds* the backend when empty; after that the file is the *desired* input to `apply` (like a `.tf`). `watchConfig` polls the serial and hot-swaps. Editing the file on disk no longer changes the running config — only `apply` does.
- **Lock** (`config_lock`, migration `0021`): `apply` holds it plan→write→release; a concurrent apply fails "locked by X". Stale locks stolen after 10m or via `config unlock`. The lock records `who@host` (Terraform's lock "Who", env-derived).
- **Serial CAS** (`recordConfigVersionCAS`): insert only if the latest serial is still the one planned against.
- **Saved-plan staleness**: `plan -out` binds a plan to a serial; `apply <planfile>` refuses if the deployed serial moved since.

State is **OSS-Terraform-shaped**: no who/why audit columns — just serial, content, revision (+ a timestamp for `history`). No `--by`/`--note`.

## Commands

```
clawpatrol plan   [-out file] <config.hcl>       diff desired vs deployed (read-only)
clawpatrol apply  [-y] [--expect N] <config.hcl | planfile>   locked re-plan + CAS
clawpatrol config history <config.hcl>           serial, time, revision, schema
clawpatrol config unlock  <config.hcl>           force-release a stuck lock
```

## Conflicts — actually handled

| Scenario | Behavior |
|---|---|
| Two `apply` at once | Second fails: "config is locked by X". |
| Stale saved plan / `--expect` mismatch | Rejected: "deployed serial is M but the plan was built against N". |
| `scp`/`vim` on the box | No effect — gateway reads the backend, not the file. |
| Crashed apply holding lock | Auto-stolen after 10m, or `config unlock`. |

## Real output

Plan + apply, seeding an empty backend:

```
$ clawpatrol plan gateway.hcl
revision: 9fb793ddb751  (1 endpoints, 1 profiles)
deployed: (none — backend is empty, this seeds serial 1)
changes: +6  -0  ~0
  + credential github
  + endpoint github
  + gateway
  + profile default
  + rule github-reads
  + rule github-writes

[...]

PR #832, opened 2026-09-10
doc: design note for the admin CLI (#811)

piscisaureus:
Design note for `clawpatrol admin` (#811). Nothing is built; this is
the shape to agree on before code.

* Six verbs over endpoints that exist today: `approve`, `devices`,
  `set-profile`, `delete-device`, `credentials`, `credential
  set|clear|connect|disconnect`.
* Auth per invocation with the dashboard root password (prompt,
  `CLAWPATROL_PASSWORD`, or `--password-file`), `cp_session` held in
  memory and logged out at exit; nothing stored on the client; TLS
  follows WebPKI with no `--insecure`.
* `--server` defaults to the URL `join` saved; `--json` for scripts;
  `--slot NAME=@file` and `NAME=-` keep secrets out of argv.
* Refuses to run inside a `clawpatrol run` child (a guard, not a
  boundary).
* `--help` regrouped into gateway / agent / admin without renaming
  any command.
* Later: hitl verbs, config apply (with #625), tailnet-operator auth
  for admin routes, a local socket on the gateway host.
* Alternatives considered and three open questions at the end.

### DIFF
--- a/doc/admin-cli-design.md
+++ b/doc/admin-cli-design.md
@@ -1,0 +1,200 @@
+# Design: `clawpatrol admin`, configuring a gateway from the CLI
+
+Status: proposal for #811. Nothing here is built.
+
+## Problem
+
+Everything that is not in `gateway.hcl` is done through the
+dashboard: pasting credential secrets, connecting OAuth
+credentials, approving a joining device, assigning a profile. That
+does not fit scripted or fleet setups, and it does not fit an agent
+provisioning another agent. #482 prototyped `approve`,
+`oauth-start` and `oauth-exchange` subcommands and was closed to
+rethink the shape.
+
+## Principles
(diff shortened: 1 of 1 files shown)


piscisaureus:
Added the command-grouping section. Recommended: flat gateway/join/run kept canonical; new config (validate, test, plugins), agent (env, status, uninstall, plus documented agent join / agent run aliases) and admin groups; internal entry points under a hidden internal group; permanent hidden aliases for the old flat spellings. Full grouping (gateway serve, agent join, agent run, two-release alias window) is written up as an option under consideration with the strongest argument each way. Admin verbs are now resource-oriented: join approve, device list|set-profile|delete, credential list|set|clear|connect|disconnect.

Issue #843, opened 2026-09-11
Dashboard-approved sync HITL request is rendered in Slack as "waiting for matching client retry"


piscisaureus:
Observed on the internal gateway (prerelease d17afc32, but the code
involved is unchanged since v0.5.9).

A Postgres `NOTIFY` matched `pg-small-mutations` (human_approver on
Slack). The client stayed connected (`operation_state = sync_waiting`,
`approval_effect = execute_upstream`). The operator approved from the
dashboard (`POST /api/hitl/decide`). The gateway then ran the
statement upstream and the client got its result, and the action row
went `pending -> approved`.

The Slack message, however, was edited to:

```
Approved.
Waiting for the client to retry the original request.
Upstream has not been called yet.
:white_check_mark: Approved — waiting for matching client retry
```

with the header switched from the endpoint name to the upstream host
and the `*Query*` label replaced by `*Path*`. Upstream was called;
nothing is waiting for a retry.

Cause: `updateRecordedMessageRefs` (cmd/clawpatrol/hitl_async_runtime.go,
the `switch result.State` around line 283) maps every
`HITLStateApproved` to `HITLOperationStateApprovedWaitingForRetry`,
which is the async retry-grant wording. It should distinguish the
sync path (approval effect `execute_upstream`): render "approved by
<who>" and, once the upstream call completes, the succeeded/failed
outcome, the way the Slack-button path does (`:white_check_mark:
approved by @user`).

Same message also shows the verb twice (`NOTIFY NOTIFY ...`):
`pgSummary` prefixes `strings.ToUpper(verb)` to a statement that
already starts with the verb. Pre-existing and cosmetic, but it reads
as a bug to reviewers; drop the prefix when the statement is present.

PR #845, opened 2026-09-11
hitl: say approved-by on sync approvals, not the retry-grant wording

piscisaureus:
Refs #843 (outcome reporting after the upstream call stays open).

After a dashboard decision on a synchronous HITL request (client
still connected, approval effect execute_upstream) the Slack message
was edited with the async retry-grant wording ("Waiting for the
client to retry... Upstream has not been called yet") and its header
switched to the upstream host and `*Path*`. The pending-message
updater mapped every approved decision to
`approved_waiting_for_retry`; it is only reachable on the sync path,
so the mapping was wrong for essentially every call. Slack-button
approvals looked right only because their response_url replace
landed after the chat.update.

* new message-only state `approved` (never stored, never a pending
  entry's operation_state); the updater picks it unless the approval
  effect is create_retry_grant, which keeps the async wording
* `HITLMessageUpdate.DecidedBy` carries who decided; Slack renders
  "Approved by X" / "Denied by X" (mrkdwn-escaped); the Signal
  notifier's delete_on_decision fires on the new state too
* the Slack message ref records the endpoint name and query label the
  original post used, so the edit keeps "Approve NOTIFY · pg-prod"
  and `*Query*`; old refs fall back to host/Path
* a late RecordMessageRef after a decision now still gets the update
  (terminal entries keep a body-stripped copy of the pending)
* pgSummary no longer prefixes the verb when the statement already
  starts with it ("NOTIFY NOTIFY ..." -> "NOTIFY ...")

Async operation updates are unchanged. HITLMessageUpdate does not
cross the plugin boundary.

### DIFF
--- a/cmd/clawpatrol/hitl_async_runtime.go
+++ b/cmd/clawpatrol/hitl_async_runtime.go
@@ -243,7 +243,11 @@ func (g *Gateway) updateHITLOperationMessage(ctx context.Context, op HITLOperati
 	}
 }
 
-func (g *Gateway) updatePendingHITLMessage(ctx context.Context, pending runtime.HITLPending, ref string, result runtime.HITLResolveResult) {
+// updatePendingHITLMessage edits the channel message behind a pending
+// sync HITL prompt once the entry reaches a terminal state. decidedBy
+// is the operator who approved or denied (empty for expiry / client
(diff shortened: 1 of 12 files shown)


piscisaureus:
Landed on main as 6b55f33f (rebased; content identical to the approved head 39cc4f1a).

Issue #839, opened 2026-09-11
UDP/443 drop follow-ups: ICMP unreachable for instant fallback, NODATA for HTTPS/SVCB queries, shared dispatch helper


piscisaureus:
Follow-ups from the review of the unconditional UDP/443 drop (#835, the fix PR). None are security issues; the drop is the enforcement.

1. **Signal the drop instead of black-holing.** Today the WireGuard forwarder's UDP handler returns `true` and closes the endpoint, and tsnet's hook runs after `CreateEndpoint` with ICMP suppressed, so a QUIC client sees nothing and waits on its own timer before falling back to TCP. Returning `false` from the transport handler before `CreateEndpoint` makes gVisor emit ICMP port-unreachable sourced from the original destination (the Linux daemon already uses this pattern in `daemon_session_linux.go`); Chromium, Firefox and curl treat the resulting `ECONNREFUSED` on their connected UDP socket as an immediate failure. Two lines on the WG path; a small patch to the vendored `wrapUDPProtocolHandler` / `acceptUDPNoICMP` on tsnet. It does not reach Linux `clawpatrol run` children (the relay cannot translate ICMP back), so the daemon should reject UDP/443 locally the same way.

2. **Answer HTTPS/SVCB (type 65/64) queries with NODATA in `forwardUpstream`.** VIP'd names already return no such record; plain `https` hosts get the upstream record relayed verbatim, including `alpn=h3` and `ech=`. With the drop in place this only costs h3-capable browsers a first-request delay per origin, but suppressing it removes the probe entirely. Separately worth a look: an ECH-enabled client presents the outer public name as SNI to the MITM, so SNI dispatch and `unknown_host` would be evaluated against e.g. `cloudflare-ech.com` rather than the real host. Not verified as exploitable; check what the MITM does in that case.

3. **One helper for the port decision.** The decision is two literal `443`s in `main.go` (`udpDispatch`, a closure inside `runGateway`, untestable) and `tailscale.go`. A shared `udpPortDisposition(port)` consumed by both would let `TestTsnetUDPDisposition` cover both transports.

4. **tsnet hook install is fail-open.** Every skip path in the catch-all installer logs and returns, after which tsnet's default `forwardUDP` relays everything. Unreachable today (vendored type assertion), but the invariant deserves `log.Fatalf`.

[...]

PR #846, opened 2026-09-11
gateway: refuse UDP/443 with ICMP unreachable; NODATA for HTTPS/SVCB

piscisaureus:
Follow-ups to #838 from #839. Two commits; the second can be dropped
on its own.

**gateway: refuse UDP/443 with ICMP unreachable and share the port
decision**

* WireGuard: the promiscuous forwarder consults `refuseUDPPort`
  before `CreateEndpoint` and returns false for 443, so gVisor emits
  ICMP port-unreachable from the original destination. A QUIC client
  gets ECONNREFUSED at once instead of waiting out its timer
  (measured 0.001 s vs 5 s on a whole-machine WireGuard client).
* tsnet: `RejectUDPFlow` hook in the vendored netstack
  (`third_party/tailscale`, documented as patch 2 in PATCH.md),
  consulted in `acceptUDPNoICMP` before `CreateEndpoint`; ICMP
  suppression stays for every other flow.
* Linux `clawpatrol run` daemon refuses UDP/443 locally the same way,
  since the relay cannot carry ICMP back to the child.
* one `udpPortDisposition` helper (53 DNS, 443 refuse, else relay)
  used by both transports and the daemon; `TestUDPPortDisposition`
  pins the table and asserts the transports agree port by port.
* `installTsnetUDPCatchAll` skip paths are now `log.Fatalf`: leaving
  tsnet's default relay in place would forward UDP/443 unchecked.
* docs: security-model notes https authorities on non-443 ports;
  tailscale and architecture docs say "refused with ICMP".

**dnsvip: answer HTTPS and SVCB queries NODATA for every name**

Plain `https` hosts had their upstream HTTPS record relayed verbatim,
including `alpn=h3` and `ech=`. With ECH the MITM only sees the
provider's public name as SNI, so under `unknown_host = passthrough`
the real host's rules never run. Answering type 64/65 with NODATA
removes both the h3 probe and the ECH config for clients that resolve
through the gateway. Clients resolving via DoH elsewhere are not
covered; `unknown_host = "deny"` remains the backstop (documented).

Not exercised end to end on the tsnet path; the WireGuard path was
measured on two GCP VMs with tcpdump confirming no UDP/443 leaves the
gateway. macOS `clawpatrol run` still swallows the refusal in the
extension's UDP pump (follow-up).

### DIFF
--- a/cmd/clawpatrol/dnsvip/dnsvip.go
+++ b/cmd/clawpatrol/dnsvip/dnsvip.go
@@ -763,6 +763,23 @@ func (a *Allocator) lazyAllocateForPattern(hostname string) bool {
 // PostUp) produces NXDOMAIN for any internal name and is the bug
 // this layer exists to fix.
 //
+// HTTPS (type 65) and SVCB (type 64) are answered NODATA — NOERROR with
+// an empty answer section, the same shape a VIP'd name returns for a
+// type it doesn't serve — and never relayed. The upstream record is
+// what tells a client to try HTTP/3 (`alpn=h3`) and hands it the
+// Encrypted Client Hello config (`ech=`). The gateway refuses UDP/443
+// regardless, so `alpn=h3` would only cost an h3-capable client a
+// refused probe per origin; `ech=` is the one that matters: a client
+// holding an ECH config sends the provider's public name as the outer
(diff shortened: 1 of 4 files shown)


piscisaureus:
Landed on main as 90640790 + 68060695 (rebased; content identical to the approved head 3296b2a6).

